PR #2, opened 2015-06-18
Use docker for travis builds.

benpatterson:


benpatterson:
@chrisndodge I recommend using docker here. There are no sudo installs needed for the tests at this point, and using docker containers is faster because Travis doesn't (necessarily) need to spin up a new image from scratch. I've noticed it is a lot faster to getting builds started when they run in this way.

Anyway...this should also validate that they are running on travis so hooray!

chrisndodge:
thanks!

PR #3, opened 2015-06-22
PHX-35 Expose REST API to populate Timer Bar

muhhshoaib:
@chrisndodge Kindly review the changes for PHX-35.


chrisndodge:
Thanks. Overall looks good. Here at edX, there's a new directory structure and filename standard since we did edx-notifications. I'll move and rename some files here, but basically this is right.

PR #6, opened 2015-06-25
(WIP) PHX-38 added the api level functionality

muhhshoaib:


edx_proctoring/views.py
```diff
@@ -3,23 +3,177 @@
...
+
+    HTTP GET
+        ** Scenarios **
+        ?exam_id=533
+        returns an existing exam  object matching the exam_id
+
+        ?course_id=edX/DemoX/Demo_Course&content_id=123
+        returns an existing exam object matching the course_id and the content_id
+    """
+    def post(self, request):
```
chrisndodge: Just note that we have to tie in a permissions system here. Only course authors should be able to create new exams. For now, we should write simply a has_permissions(user, action) helper function and any function that would require staff permissions (e.g. create new exam, add/update/delete allowance, etc.) should check user.is_staff to see if it is on, if not raise a PermissionDenied exception.

Later on (this coming sprint) we'll have to get the real permissions checks (CourseAccessRole) but that subsystem will have to be dependency injected in from the LMS (since we don't have access to those models from this repo).

Can you apply these patterns to all POST and PUT operations? Maybe - as bonus points - you can write a method decorator?

edx_proctoring/api.py
```diff
@@ -34,6 +71,11 @@ def get_exam_by_id(exam_id):
 
     Returns dictionary version of the Django ORM object
     """
+    proctored_exam = ProctoredExam.get_exam_by_id(exam_id)
+    if proctored_exam is None:
+        raise ProctoredExamNotFoundException
+
+    return proctored_exam.__dict__
```
chrisndodge: I see you are using this pattern often in this file. I think I noticed from @afzaledx's comment that this might be changing to returning output from the Serializer?

I think I would prefer output from the Serializer as well, but - either way - I think we should document what the schema looks like in the documentation comment string so that callers (in case they are calling into api.py as an in-proc library) know what to expect.

edx_proctoring/views.py
```diff
@@ -3,23 +3,176 @@
...
+        HTTP GET handler.
+            Scenarios:
+                by exam_id: calls get_exam_by_id()
+                by course_id, content_id: get_exam_by_content_id()
+
+        """
+
+        course_id = request.QUERY_PARAMS.get('course_id', None)
+        content_id = request.QUERY_PARAMS.get('content_id', None)
+        exam_id = request.QUERY_PARAMS.get('exam_id', None)
```
chrisndodge: In general, we never use query string parameters, but rather put them in the URL paths so that they are passed into the handler.  I'll see if I can find some examples for you.

muhhshoaib:
@chrisndodge We have made all the changes related to PHX-39 and PHX-40.

PR #15, opened 2015-07-03
improved coverage.

afzaledx:


edx_proctoring/tests/test_api.py
```diff
@@ -14,8 +14,8 @@
     stop_exam_attempt,
     get_active_exams_for_user,
     get_exam_attempt,
-    create_exam_attempt
-)
+    create_exam_attempt,
+    get_student_view)
```
nedbat: Please use a trailing comma and put the close-paren on its own line, so that future diffs will be cleaner.

edx_proctoring/tests/test_api.py
```diff
@@ -295,3 +295,99 @@ def test_get_active_exams_for_user(self):
         self.assertEqual(len(student_active_exams), 2)
         self.assertEqual(len(student_active_exams[0]['allowances']), 2)
         self.assertEqual(len(student_active_exams[1]['allowances']), 0)
+
+    def test_get_student_view(self):
+        """
+        Test for get_student_view
+        """
```
nedbat: There's no need for docstrings for test_* methods.

edx_proctoring/tests/test_api.py
```diff
@@ -295,3 +295,99 @@ def test_get_active_exams_for_user(self):
...
+            course_id=self.course_id,
+            content_id=self.content_id,
+            context={
+                'is_proctored': True,
+                'display_name': self.exam_name,
+                'default_time_limit_mins': 90
+            }
+        )
+        self.assertIn('data-exam-id="%d"' % self.proctored_exam_id, rendered_response)
+        self.assertIn('Would you Like to take %s as Proctored Exam?' % self.exam_name, rendered_response)
```
nedbat: I "Like" really capitalized?  Seems like we should fix that message.

PR #18, opened 2015-07-07
(WIP) PHX-49 initial work

chrisndodge:


edx_proctoring/static/proctoring/js/proctored_exam_allowance_view.js
```diff
@@ -0,0 +1,49 @@
...
+    'use strict';
+
+    edx.instructor_dashboard = edx.instructor_dashboard || {};
+    edx.instructor_dashboard.proctoring = {};
+
+    edx.instructor_dashboard.proctoring.ProctoredExamAllowanceView = Backbone.View.extend({
+        initialize: function (options) {
+            this.$el = options.el;
+            this.model = options.model;
+            this.temPlateUrl = options.allowance_template_url;
```
chrisndodge: typo

edx_proctoring/urls.py
```diff
@@ -30,6 +30,11 @@
         name='edx_proctoring.proctored_exam.attempt'
     ),
     url(
+        r'edx_proctoring/v1/proctored_exam/allowance/{}$'.format(settings.COURSE_ID_PATTERN),
+        views.ExamAllowanceView.as_view(),
+        name='edx_proctoring.proctored_exam.allowance'
+    ),
```
chrisndodge: I think the course_id should come earlier in the URL pattern:

edx_proctoring/v1/proctored_exam/{}/allowance

Also, we'll have to check the COURSE_ID_PATTERN regex, we do have a "new style" format. There is a regex defined in edx-platform. I'll have to dig it up.

edx_proctoring/static/proctoring/templates/add-new-allowance.underscore
```diff
@@ -0,0 +1,52 @@
+<div class='modal-header'>Add a new Allowance</div>
```
chrisndodge: We need to I18N this file

chrisndodge:
Fixed circular reference problem that was causing Studio to not start up. Needed to move an import into the init() method.

+1

PR #20, opened 2015-07-09
Switch to use User object rather than integer. Also add new field for…

chrisndodge:
… allowed_time_limit_mins


chrisndodge:
@afzaledx @muhhshoaib FYI here are the changes you asked for re: changing to User being a foreign key.

Also, I've added a new column 'allowed_time_limit_mins' which gets populated when a UserAttempt is created, basically it'll take the time limit from the Exam, but this is where we will add on any additional allowances.

PR #23, opened 2015-07-13
(WIP) PHX-11

muhhshoaib:


chrisndodge:
Hi @afzaledx I just noticed your last commit. I don't think that approach is going to work very well, as that field is expected to be an integer.

If it helps to simplify the backbone JS, I don't think we need to "humanize" the allowed minutes, we can simply keep it an integer. That'd be a nice-to-have, and something we can omit, if need be.

Thanks

chrisndodge:
Downloaded and tried locally. Works great.

PR #25, opened 2015-07-15
integrate with new SoftwareSecure v2.x build

chrisndodge:


edx_proctoring/backends/tests/test_backend.py
```diff
@@ -33,6 +33,13 @@ def stop_exam_attempt(self, exam, attempt):
...
+    def get_software_download_url(self):
+        """
+        Returns the URL that the user needs to go to in order to download
+        the corresponding desktop software
+        """
+        return None
+
 
 class PassthroughBackendProvider(ProctoringBackendProvider):
     """
```
nedbat: Maybe I'm not understanding something here, but what is the point of all these methods?  Isn't it true that if you just omit all the methods, and make the body of the class be `pass`, that it will work the same?  If a child class is missing a method, the parent's method is called.

PR #30, opened 2015-07-17
Cdodge/touchups3

chrisndodge:


chrisndodge:
Note to self, we need to regenerate this migration

PR #31, opened 2015-07-21
Build out review callback support for SoftwareSecure

chrisndodge:


chrisndodge:
@afzaledx or @muhhshoaib can you do a code review. Unfortunately, you can't actually manually test the callback as SoftwareSecure needs to post back the results. Can look at the code quality/style?

edx_proctoring/backends/software_secure.py
```diff
@@ -99,6 +110,108 @@ def get_software_download_url(self):
         """
         return self.software_download_url
 
+    def on_review_callback(self, payload):
+        """
+        Called when the reviewing 3rd party service posts back the results
+
+        Documentat on the data format can be found from SoftwareSecure's
```
afzaledx: Documentation

edx_proctoring/backends/software_secure.py
```diff
@@ -99,6 +110,108 @@ def get_software_download_url(self):
...
+        log_msg = (
+            'Received callback from SoftwareSecure with review data: {payload}'.format(
+                payload=payload
+            )
+        )
+        log.info(log_msg)
+
+        # payload from SoftwareSecure is a JSON payload
+        # which has been converted to a dict by our caller
+        data = payload['payload']
```
afzaledx: Is it ok to have no validation on the payload coming from the SoftwareSecure? Worst case: an exception, but we're logging the payload at the very beginning to help debug. so it might be ok to have no validation. Some validation, however, would make it easier to figure out the exact problem with the payload received (in case there arises one later).

edx_proctoring/models.py
```diff
@@ -473,3 +485,71 @@ def _make_archive_copy(item):
...
+class ProctoredExamSoftwareSecureComment(TimeStampedModel):
+    """
+    This is where we store the proctored exam review comments
+    from the exam reviewers
+    """
+
+    # which student attempt is this feedback for?
+    # This is not a FK because it might be a review for
+    # an attempt which had been moved to the archive table
+    review = models.ForeignKey(ProctoredExamSoftwareSecureReview)
```
afzaledx: The comment says not a foreign key but it's still an FK.

afzaledx:
All good other than a couple of typos. :+1:

PR #32, opened 2015-07-22
According pane content display Section

muhhshoaib:

1) expand/collapse regions of the proctoring section of Instructor Dashboard.
2) The HTTP delete works fine for me and @afzaledx . I have tried it on my local and also in the sandbox and it seems working fine. Although if we delete a single attempt listed on the page, then only the header displays on the page. I have hide the header and the search, when there is no exam attempt on the page.

Thanks


chrisndodge:
Thanks. Seems like coveralls is not picking up the task. Not sure if that service is down or not.

chrisndodge:
OK, ran this branch locally. LGTM, merging. Thanks.

PR #35, opened 2015-07-23
Add matching header bar to heartbeat page

chrisndodge:


muhhshoaib:
@chrisndodge I have implemented the " Proctoring Set Up is Completion Page"

chrisndodge:
@muhhshoaib thanks. @catong can you suggest some student documentation to put on this screen? SOL-1073 is the JIRA ticket with corresponding mockup. I'd like to merge this in today, if possible.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+          <li class="roomscan completed">Record Room Scan</li>
+          <li class="photo completed">Take User Photo</li>
+          <li class="end active" >Enter Exam</li>
+        </ul>
+      </div>
+    </div>
+
+    <div class="steps-container">
+      <div class="body-container">
+        <h3>{% blocktrans %} Your Proctoring Set Up is Complete {% endblocktrans %}</h3>
```
catong: "Set up" -> "Setup". Uppercase "i" for "Is":
Your Proctoring Setup Is Complete

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+            Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
+            Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim.
+          {% endblocktrans %}
+        </p>
+      </div>
+    <div class="footer">
+      <button>
+        {% blocktrans %} Enter My Exam {% endblocktrans %}
+      </button>
+    </div>
```
chrisndodge: Note to @catong, I'm not sure if we are going to be able to keep this button. It's still TBD. It's a bit hard to explain, but I can speak to you in person.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+          <li class="photo completed">Take User Photo</li>
+          <li class="end active" >Enter Exam</li>
+        </ul>
+      </div>
+    </div>
+
+    <div class="steps-container">
+      <div class="body-container">
+        <h3>{% blocktrans %} Your Proctoring Set Up is Complete {% endblocktrans %}</h3>
+        <p>{% blocktrans %} You may proceed and enter exam now {% endblocktrans %}</p>
```
catong: You can now return to the courseware window and start your exam.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+
+        <div class="alert">
+          {% blocktrans %} Please leave the window open in the background {% endblocktrans %}
```
catong: Do not close this window. Leave it open in the background as you take your exam.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
-    {% blocktrans %}
-      Your proctoring session has started. Do not close this window.
-      Return to the browser window where your course is open to take your exam.
-    {% endblocktrans %}
-  </p>
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>Sample Progress Steps edX</title>
```
chrisndodge: We should change this title.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
catong: Thanks, @chrisndodge. With no button, we would be starting the timer in edx platform at the same time as SS's own "timer"? I can't remember if there is a separate "Start exam" button on the edX end. If not, it seems like a less than ideal student experience to have to race back to the edX page to avoid losing time. And it probably means we wouldn't want to put too much text on this page that would take time to read. Or is there a buffer amount of time to allow the student to switch windows and get set?

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
catong: Where does this title display? I don't see it in the mockup.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
chrisndodge: Yea, I think we should all huddle on this issue.

Basically - at this point - the proctoring recording has started (and I'm not 100% sure on how they handle their own sense of elapsed time).

I had coded it up to start our own timer when this page is rendered. If we want to change it so that the student says "yes, start now", it's not a big change on our end.

That said, I'm not sure what the consequences might be if the student stays at this point (while proctoring is recording) and doesn't hit the 'start' button for a long time. That's probably a question we should ask SoftwareSecure (aka does the desktop software keep recording indefinately? does it record for the registered time limit + N spare minutes?)

catong: So it sounds as in the text on this page it would be good to make clear that proctoring recording has started and that from this point on, students need to be behave in a way that won't cause a Fail result in proctoring, even though they haven't started answering the exam questions yet. I'll put some text like that in, and then @chrisndodge and @pbaruah can we meet tomorrow to discuss the workflow?

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
catong: @chrisndodge @pbaruah 
Maybe "Your Proctoring Session Is Ready" instead

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+
+    <div class="steps-container">
+      <div class="body-container">
+        <h3>{% blocktrans %} Your Proctoring Set Up is Complete {% endblocktrans %}</h3>
+        <p>{% blocktrans %} You may proceed and enter exam now {% endblocktrans %}</p>
+
+        <div class="alert">
+          {% blocktrans %} Please leave the window open in the background {% endblocktrans %}
+        </div>
+        <h5> {% blocktrans %} Remember to abide by the guidelines and rules of your proctored exam, including: {% endblocktrans %}</h5>
```
catong: @chrisndodge Given my latest suggestions above, perhaps we don't want or need too much other text on this page. 
Perhaps just a link to the online proctoring policy and a link to contact edX support for help? And only if there is some way to "stop the clock" we can provide guidance for that.
@pbaruah

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
```diff
@@ -1,11 +1,205 @@
...
+          {% blocktrans %}
+            Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor.
+            Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
+            Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim.
+          {% endblocktrans %}
+        </p>
+      </div>
+    <div class="footer">
+      <button>
+        {% blocktrans %} Enter My Exam {% endblocktrans %}
```
catong: If we keep the button, my suggested text for the label is: "Go to my exam"

catong:
@chrisndodge I've finished my pass with the text suggestions, pending discussion with @pbaruah.

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
chrisndodge: We don't currently have a means to "stop the clock". We can offer "restart from scratch"....

edx_proctoring/templates/proctoring/proctoring_launch_callback.html
chrisndodge: The problem here is that we do not have a consistent technological means to bring the LMS browser window to the foreground. Many browsers forbid this (this of the havoc that could be wrought if webpages could always bring themselves to the foreground). For example, we can do so in Chrome, but not other browser types, and I think there are user settings that people can set to prevent browsers from bringing themselves to the foreground.

This is why I'm hesitant to having any sort of a button here, because expectations might be mis-set.

chrisndodge:
@catong, @muhhshoaib, @afzaledx Merging this for now as I need to deploy to a sandbox. We can continue to revisit the text and make updates along the way.

PR #38, opened 2015-07-24
SOL-1085

muhhshoaib:

- add new columns in the ProctoredStudentAttempt Table
- record the last polling contact info
- fixed the bug in the search bar of proctored exam attempts.


edx_proctoring/callbacks.py
```diff
@@ -100,12 +102,16 @@ def get(self, request, attempt_code):  # pylint: disable=unused-argument
         """
 
         attempt = get_exam_attempt_by_code(attempt_code)
+        ip_address = get_ip(request)
```
muhhshoaib: NOTE: We are using the Django-ipware library method that is already in the edx-platform requirements.

PR #58, opened 2015-08-04
PHX-78 used the new service for resetting the student attempts

muhhshoaib:
@chrisndodge Here are the changes for the PHX-78. Used the new service for the resetting the student attempts.


chrisndodge:
This approach looks good but in the PR for edx-platform, I'm suggesting we call the class InstructorService and naming the service 'instructor'. You'll also need to change that here as well.

chrisndodge:
Tried this locally. Seems to work. Thanks.

PR #60, opened 2015-08-05
PHX-79

afzaledx:
Moved the instructor dashboard view from being initialized in the courseware.


chrisndodge:
OK, tried it out locally and it appears to work. Thanks.

PR #65, opened 2015-08-06
PHX-86 added the css class names for the color bars to appear on the …

muhhshoaib:
@chrisndodge added the color bar css in the templates.


chrisndodge:
tried it locally. Looks good. Thx

PR #97, opened 2015-08-18
change proctoring eligibility to be true unless we explicitly failed …

chrisndodge:
…a icrv

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -498,7 +498,7 @@ def mark_exam_attempt_as_ready(exam_id, user_id):
     return update_attempt_status(exam_id, user_id, ProctoredExamStudentAttemptStatus.ready_to_start)
 
 
-def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True):
+def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, cascade_effects=True):
     """
     Internal helper to handle state transitions of attempt status
     """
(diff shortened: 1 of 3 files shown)


chrisndodge:
@afzaledx @muhhshoaib can you take a look at this when you get a chance. Basically I think we need to change two bits of logic regarding when we show the proctoring screen:

- We always show proctoring as long as there is no failure in the ICRV processing. Current implementation is that all ICRV needs to be passed before a proctoring can occur. But this is incorrect because an ICRV can happen after the first proctoring: e.g. ICRV1, Proctoring1, ICRV2, Proctoring2

- Also, when we fail/decline on proctoring exam, the business logic is supposed to be that all other proctored exams not yet started should be also marked as declined, so that the user does not get the option to choose proctoring

I'd like to get this change into next week's release so that we can have the correct business logic in production.

afzaledx:
@chrisndodge, Since you made a ProctoredExamStudentAttemptStatus class, it made sense to use it everywhere we were checking the status to be in a list or not. Made the changes and have created a PR https://github.com/edx/edx-proctoring/pull/97. Otherwise :+1:

afzaledx:
@chrisndodge, Since you made a ProctoredExamStudentAttemptStatus class, it made sense to use it everywhere we were checking the status to be in a list or not. Made the changes and have created a PR https://github.com/edx/edx-proctoring/pull/98. Otherwise :+1:

chrisndodge:
I want to do a manual check on this PR before I merge it.

chrisndodge:
OK, I did a manual check on localdev with the following gating order:

ICRV1, Proctoring1, ICRV2, Proctoring2

Now we only remove the proctoring option when the user explicitly fails ICRV1/2, however if it is in a pending state, the proctoring option will still be available.

PR #98, opened 2015-08-18
refactored all checks of status to the ProctoredExamStudentAttemptSta…

afzaledx:
…tus class


chrisndodge:
Thanks, I think this looks right

PR #99, opened 2015-08-19
PHX-118 Javascript for the timer bar should be visible when scrolling the page

muhhshoaib:

### DIFF
--- a/edx_proctoring/static/proctoring/js/proctored_app.js
+++ b/edx_proctoring/static/proctoring/js/proctored_app.js
@@ -5,4 +5,16 @@ $(function() {
         model: new ProctoredExamModel()
     });
     proctored_exam_view.render();
+
+    // show the timer bar when the window scrolls.
+    $(window).scroll(function() {
+        if ($(this).scrollTop() > 60){
+            $(".proctored_exam_status").addClass('fixed');
+            $(".wrapper-course-material").css('margin-top', '58px');
+        }
+        else {
+            $(".proctored_exam_status").removeClass('fixed');
+            $(".wrapper-course-material").css('margin-top', '0');
+        }
+    });
 });
--- a/edx_proctoring/static/proctoring/js/proctored_app.js
+++ b/edx_proctoring/static/proctoring/js/proctored_app.js
@@ -5,16 +5,4 @@ $(function() {
         model: new ProctoredExamModel()
     });
     proctored_exam_view.render();
-
-    // show the timer bar when the window scrolls.
-    $(window).scroll(function() {
-        if ($(this).scrollTop() > 60){
-            $(".proctored_exam_status").addClass('fixed');
-            $(".wrapper-course-material").css('margin-top', '58px');
-        }
-        else {
-            $(".proctored_exam_status").removeClass('fixed');
-            $(".wrapper-course-material").css('margin-top', '0');
-        }
-    });
 });
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
@@ -11,6 +11,7 @@ var edx = edx || {};
             _.bindAll(this, "detectScroll");
             this.$el = options.el;
             this.timerBarTopPosition = this.$el.position().top;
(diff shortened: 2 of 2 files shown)


edx_proctoring/static/proctoring/js/proctored_app.js
```diff
@@ -5,4 +5,16 @@ $(function() {
         model: new ProctoredExamModel()
     });
     proctored_exam_view.render();
+
+    // show the timer bar when the window scrolls.
+    $(window).scroll(function() {
+        if ($(this).scrollTop() > 60){
+            $(".proctored_exam_status").addClass('fixed');
+            $(".wrapper-course-material").css('margin-top', '58px');
```
chrisndodge: I'm wondering if this might be fragile to have an inline CSS rule here which affects the rendering of courseware styling that is defined in edx-platform. If we don't have this, what happens visually? I'm going to get this running locally to see for myself....

muhhshoaib: There is a disrupt change in the course navigation when the window scroll position matches the timer bar position.

edx_proctoring/static/proctoring/js/proctored_app.js
```diff
@@ -5,4 +5,16 @@ $(function() {
         model: new ProctoredExamModel()
     });
     proctored_exam_view.render();
+
+    // show the timer bar when the window scrolls.
+    $(window).scroll(function() {
+        if ($(this).scrollTop() > 60){
```
chrisndodge: Is there a way to compute this rather than hard coding "60" here. My sense is that this won't be correct if the page header region changes in height.

Sorry, also just noticed that this scroll event handler is in the proctored_app.js file. Seems like it should be in the proctored_exam_view only if the timer bar is visible.

I think we only have to compute this once per page load, so if we can compute it, we should store it.

edx_proctoring/static/proctoring/js/proctored_app.js
muhhshoaib: by adding that line `$(".wrapper-course-material").css('margin-top', '58px');`
screen shot 2015-08-19 at 7 28 12 pm

muhhshoaib: by removing that line `$(".wrapper-course-material").css('margin-top', '58px');`
screen shot 2015-08-19 at 7 28 37 pm

muhhshoaib: the course navigation bar will not appear when window scroll position reaches the start of the timer bar.

chrisndodge:
@muhhshoaib yea, this PR regresses scroll behavior in coursewhere when there is not a timed exam in progress. The margin-top inline definition is the culprit. To reproduce, don't start a timed exam and scroll on any page in the courseware. You'll see what I mean.

I think to fix this you need to put the scroll handler in the view.js file and quickly exit the handler if the timer bar is not visible.

muhhshoaib:
@chrisndodge Can you review now. I have moved the scroll event in the view.js file

edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -43,6 +45,17 @@ var edx = edx || {};
             /* will call into the rendering */
             this.model.fetch();
         },
+        detectScroll: function(event) {
+            if ($(event.currentTarget).scrollTop() > this.timerBarTopPosition) {
+                $(".proctored_exam_status").addClass('fixed');
+                $(".wrapper-course-material").css('margin-top', '58px');
```
chrisndodge: can '58px' also be computed here rather than hard coded? It seems like it'd be very similar (maybe with an offset) to this.timerBarTopPosition...

chrisndodge:
@muhhshoaib OK behavior looks better now. Just one remaining question about if we can compute that '58px' rather than hard code it, as I'm guessing it is dependent on the vertical size of the page header....

edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
muhhshoaib: done

muhhshoaib:
@chrisndodge added the computed value rather than the hardcoded value.

PR #101, opened 2015-08-19
bump django version

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib FYI, edx-platform is upgrading to 1.4.22.

I took this opportunity to refactor how we are managing dependencies to avoid version conflicts with edx-platform. I made a local_requirements.txt that should be pip installed when running on a localdev. This way the setup.py process will not try to install conflicting/downgraded versions of any dependent library.

edx_proctoring/tests/utils.py
```diff
@@ -30,10 +30,8 @@ def login_user(self, user):
 
         # Create a fake request to store login details.
         request = HttpRequest()
-        if self.session:
-            request.session = self.session
-        else:
-            request.session = engine.SessionStore()
+
+        request.session = engine.SessionStore()
```
chrisndodge: Note to reviewers, the change to 1.4.22 broke this helper session manager. 1.4.22 changed the way sessions are managed.

PR #102, opened 2015-08-20
PHX-121

muhhshoaib:
@chrisndodge Kindly review the changes for PHX-121. 
- Fixed the bug when clicking outside of the active exam page, don't redirect to the exam page.

### DIFF
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
@@ -85,12 +85,12 @@ var edx = edx || {};
                               action: 'stop'
                             },
                             success: function() {
-                              // Reloading page will reflect the new state of the attempt
-                              location.reload();
+                              // change the location of the page to the active exam page
+                              // which will reflect the new state of the attempt
+                              location.href = self.model.get('exam_url_path');
                             }
                         });
                     });
-                    //$('.proctored-exam-action-stop').css('cursor', 'pointer');
                 }
             }
             return this;


edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -85,12 +85,12 @@ var edx = edx || {};
                               action: 'stop'
                             },
                             success: function() {
-                              // Reloading page will reflect the new state of the attempt
-                              location.reload();
+                              // change the location of the page to the active exam page
+                              // which will reflect the new state of the attempt
+                              location.href = self.model.get('exam_url_path');
```
chrisndodge: Hi @muhhshoaib just confirming will this force reload the browser (across all browser types?) if the user is already on the same page as 'exam_url_path'?

muhhshoaib: I have checked on the firefox, chrome and safari browsers and it is working fine(reloading the page to the exam url).

for browser support I have also looked at the following link http://www.w3schools.com/jsref/prop_loc_href.asp

muhhshoaib: the `exam_url_page` is something like this
`/courses/course-v1:edX+DemoX+Demo_Course/jump_to/block-v1:edX+DemoX+Demo_Course+type@sequential+block@edx_introduction`.

When we hit the URL above, the server redirects us to something like:
`/courses/course-v1:edX+DemoX+Demo_Course/courseware/d8a6192ade314473a78242dfeedfbf5b/edx_introduction/`

which will always be a different URL, so no worries about hitting the same URL again.

chrisndodge: Ah, yes, you are correct. Thanks for the diligence on that one.

PR #104, opened 2015-08-20
send email on proctoring attempt status change

hasnain-naveed:


edx_proctoring/templates/emails/proctoring_attempt_status_email.html
```diff
@@ -0,0 +1,9 @@
+{% load i18n %}
+
+Dear {{ student_name }},
+
+    This email is to notify that status of {{ exam_name }} for {{ course_name }} has changed to {{ status }} on {{ date }}.
+
+Thank you.
+
+
```
chrisndodge: Thanks for including i18n at the top, but these strings are not wrapped as translatable. See some of the Django templates in edx_proctoring/templates/... to see the pattern that is used. 

Also, @catong @griffresch we are going to need some real text here. Are you available to provide that? There are three states: pending review, satisfactory review, failed review, which I presume need different text.

@hasnain-naveed depending on the complexity of the differences, we might actually need three different email templates to make it more legible (from a code perspective) rather than having lots of 'if-conditionals' in the template.

edx_proctoring/tests/test_api.py
```diff
@@ -1539,3 +1540,46 @@ def test_update_exam_attempt(self):
...
+        Assert that email is sent on the following statuses of proctoring attempt.
+        """
+
+        exam_attempt = self._create_started_exam_attempt()
+        update_attempt_status(
+            exam_attempt.proctored_exam_id,
+            self.user.id,
+            status
+        )
+        self.assertEquals(len(mail.outbox), 1)
```
chrisndodge: Is there a way we can get the content of the email in the outbox, so we can verify, via self.assertIn(...)?

edx_proctoring/api.py
```diff
@@ -640,9 +641,43 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
...
+            'student_name': exam_attempt_obj.student_name,
+            'course_name': 'edx demo course',
+            'exam_name': exam_attempt_obj.proctored_exam.exam_name,
+            'status': exam_attempt_obj.status,
+            'date': exam_attempt_obj.modified
+        })
+    )
+
+    subject = (
+        'Status update of {exam_name} for {course_name}'.format(
```
chrisndodge: This needs to be internationalized. Please wrap a _() around this.

edx_proctoring/api.py
```diff
@@ -640,9 +641,43 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
...
+            course_name='edx demo course',
+            exam_name=exam_attempt_obj.proctored_exam.exam_name
+        )
+    )
+
+    EmailMessage(
+        body=body,
+        from_email=settings.DEFAULT_FROM_EMAIL,
```
chrisndodge: I'm wondering if we should allow for an override in PROCTORING_SETTINGS, maybe this should be:

from_email = (
     settings.PROCTORING_SETTINGS['STATUS_EMAIL_FROM_ADDRESS'] if
     'STATUS_EMAIL_FROM_ADDRESS' in settings.PROCTORING_SETTINGS
     else settings.DEFAULT_FROM_EMAIL
)

and then pass in from_email into this constructor...

chrisndodge:
@hasnain-naveed thanks for the PR, nice work!

Just some high level feedback. We have to i18n'ify some of the code. You should be able to find some established patterns in the edx-proctoring and edx-platform repos.

edx_proctoring/templates/emails/proctoring_attempt_status_email.html
catong: @chrisndodge @griffresch What do you mean by "real text"? If the intention of these emails is simply to notify students that the status of their proctoring review has changed, it could be something like this: 

"This email is to let you know that the status of your proctoring review for the {NAME} in {COURSE} has changed to {STATUS}."

I'd suggest that if the course name/number appears in the subject, it doesn't need to be repeated in the body text. I did not include the date; I'm assuming that the actual change triggers these emails, so the change date = the email date. I'm not sure it's that valuable to students to explicitly be told a date, but I could be wrong.
Thoughts?

edx_proctoring/api.py
```diff
@@ -640,9 +641,43 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
...
+            'course_name': 'edx demo course',
+            'exam_name': exam_attempt_obj.proctored_exam.exam_name,
+            'status': exam_attempt_obj.status,
+            'date': exam_attempt_obj.modified
+        })
+    )
+
+    subject = (
+        'Status update of {exam_name} for {course_name}'.format(
+            course_name='edx demo course',
```
chrisndodge: whoa! We can't hard code the course name here.... :-)

chrisndodge: That said, we need to come up with a means to get this information as there's no access to course names at this point in time. @afzaledx is this something you can design? We probably can extend the 'credit' service to call out to get the course name.

edx_proctoring/templates/emails/proctoring_attempt_status_email.html
chrisndodge: @catong sorry to be vague, we're just using placeholder content for now. I pinged @griffresch on this and he'll try to propose some initial content for review.

edx_proctoring/api.py
chrisndodge: Also, @griffresch @catong do you want to propose a subject line (for the 3 states: Submitted, Successful, Unsuccessful)?

catong: Aren't we going with what we already show on the exam page after the student has submitted? Results are: Pending, Satisfactory, Unsatisfactory?

catong: @chrisndodge @griffresch A single subject line for all states, or one for each?
Maybe something like Proctoring Session Results Update for {course_name} {exam_name}?

edx_proctoring/api.py
hasnain-naveed: @chrisndodge 
I have hard coded course name intentionally. I will replace it whenever we will decide about this.

edx_proctoring/tests/test_api.py
```diff
@@ -1539,3 +1540,47 @@ def test_update_exam_attempt(self):
...
+        self.assertIn('Proctoring Session Results Update for', mail.outbox[0].subject)
+
```
hasnain-naveed: @chrisndodge 
I have added the subject check, we can add this check on the email body later. Because for now, I have not the exact template for email body.

hasnain-naveed:
@griffresch @catong, we need the exact content for email body and subject. Thanks.

catong:
@hasnain-naveed I made a suggestion earlier in a comment that got covered by a commit: https://github.com/edx/edx-proctoring/pull/104#discussion_r37541786

"This email is to let you know that the status of your proctoring session review for the {NAME} in {COURSE} has changed to {PENDING/SATISFACTORY/FAILED}."

I'll follow up with @griffresch to see if he has additional or other thoughts.

catong:
@hasnain-naveed 
Here is text that @griffresch confirms:
"This email is to let you know that the status of your proctoring session review for {EXAMNAME} in {COURSE} is {PENDING/SATISFACTORY/UNSATISFACTORY}. If you have any questions about proctoring, contact edX support at info@edx.org."

muhhshoaib:
Small Nit Picks. Looks good to me :+1:

edx_proctoring/templates/emails/proctoring_attempt_status_email.html
```diff
@@ -0,0 +1,9 @@
+{% load i18n %}
+
+{% blocktrans %}
+
+This email is to let you know that the status of your proctoring session review for {{ exam_name }} in
+<a href="{{ course_url }}">{{ course_name }} </a> is {{ status }}. If you have any questions about proctoring,
+contact edX support at info@edx.org.
```
chrisndodge: Please don't hard code 'edX' and the info@edx.org here since other Open edX users will be using this software.

Instead of 'edX', pass in as a context field settings.PLATFORM_NAME. I'm not sure where to get info@edx.org from configuration, but I'm sure there's an existing configuration setting. I can look if you have trouble finding it.

edx_proctoring/tests/test_api.py
chrisndodge: Just a nudge on this question

edx_proctoring/api.py
```diff
@@ -640,9 +642,78 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
         exam_attempt_obj.started_at = datetime.now(pytz.UTC)
         exam_attempt_obj.save()
 
+    if ProctoredExamStudentAttemptStatus.needs_status_change_email(exam_attempt_obj.status):
```
chrisndodge: We should not send a status email for "practice exams" as well as exams which were not taken as proctored. Both of these have corresponding flags/fields in the Exam and Attempt models. Can you also add a unit test to make sure practice exams and tests taken non-proctored don't send email. Thanks.

chrisndodge:
@hasnain-naveed thanks for the PR. Can you add logic to not send status emails for practice or non-proctored exams. Also, we shouldn't hard code edX specific words or information.

edx_proctoring/templates/emails/proctoring_attempt_status_email.html
hasnain-naveed: @chrisndodge I couldn't info@edx.org in any config file. Can you please look into it? Thanks

chrisndodge: I'm pretty sure that should be settings.CONTACT_EMAIL

edx_proctoring/api.py
```diff
@@ -640,9 +643,72 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
         exam_attempt_obj.started_at = datetime.now(pytz.UTC)
         exam_attempt_obj.save()
 
+    # email will be send when the exam is proctored and not practice exam
+    # and the status is verified, submitted or rejected
+    if exam_attempt_obj.taking_as_proctored and not exam_attempt_obj.is_sample_attempt and\
+            ProctoredExamStudentAttemptStatus.needs_status_change_email(exam_attempt_obj.status):
```
chrisndodge: Typically, as a pattern, for long conditional statements, in order to make a bit more readable and to avoid the \ character between line wraps, I do the conditional evaluation into a local variable - with parenthesis - right above the if statement like this:

should_send_status_email = (
   exam_attempt_obj.taking_as_proctored and
  not exam_attempt_obj.is_sample_attempt and
 ProctoredExamStudentAttemptStatus.needs_status_change_email(exam_attempt_obj.status)
)
if should_send_status_email:
   :

chrisndodge:
LGTM. One small style improvement suggestion. I'll do it myself and squash commits and merge.

Thanks! Good work!

PR #105, opened 2015-08-20
Add some more robustness in case a user's fullname has unicode charac…

chrisndodge:
…ters in it


chrisndodge:
@afzaledx @muhhshoaib I need a quick PR on this as I see some exceptions happening in production and we need to get this into the RC that is being cut tomorrow

afzaledx:
:+1:

PR #106, opened 2015-08-20
when the client app polls, be sure to timeout when time expires

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib here is another one for team review. Basically we want to make sure that if someone closes their browser and leave the desktop client running, the polling mechanism in the client will ultimate cause the attempt to state transition to timed_out (which then transitions to submitted). This is to get the client app to shut itself down.

edx_proctoring/api.py
```diff
@@ -279,6 +279,16 @@ def get_exam_attempt_by_id(attempt_id):
...
 
 
+def get_exam_attempt_by_code(attempt_code):
+    """
+    Signals the beginning of an exam attempt when we only have
+    an attempt code
+    """
+
+    exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_exam_attempt_by_code(attempt_code)
+    return _get_exam_attempt(exam_attempt_obj)
```
afzaledx: Not related to this code, but the line in _get_exam_attempt()

we should probably check the exam_attempt_obj for None in the start rather than in the second line:

attempt = serialized_attempt_obj.data if exam_attempt_obj else None

otherwise
attempt = serialized_attempt_obj.data if serialized_attempt_obj else None might be better.

afzaledx:
otherwise :+1:

PR #107, opened 2015-08-21
Do the pagination before doing serialization, also refactor out the A…

chrisndodge:
…PI endpoint which is for instructor only access


chrisndodge:
@afzaledx @muhhshoaib yet another. I noticed in the monitoring of our environment, that querying for all attempts in a course is suboptimal. Looking at the code, I see we were doing the serialization before the pagination (over all objects). This should just serialize just the current page.

chrisndodge:
I should also note that I refactored out the endpoint which services the Instructor Dashboard from the endpoint which supports a single user's calling. They are different permissions (Instructor Dashboard is only for global staff) and I think this cleans up the code a bit more.

muhhshoaib:
:+1:  Looks Good 
I have also added couple of test cases.

PR #112, opened 2015-08-21
make sure that we also time out when status is not just started

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib noticed that I missed a use case regarding timeout detection. This isn't urgent - I already missed the deadline. It's OK to go out in the following week.

chrisndodge:
@afzaledx @muhhshoaib can one of you take a look at that. I'd like to get this merged in.

afzaledx:
:+1:  from my side.

PR #114, opened 2015-08-21
Upgrade django-model-utils to 2.3.1

wedaly:


requirements.txt
```diff
@@ -1,4 +1,10 @@
-
+# Django/Framework Packages
+django>=1.4.12,<=1.4.22
+django-model-utils==2.3.1
+South>=0.7.6
+djangorestframework>=2.3.5,<=2.3.14
+pytz>=2012h
+pycrypto>=2.6
```
wedaly: @chrisndodge I saw that you recently moved these requirements to a separate file to resolve conflicts with edx-platform.  What errors were you running into?  When I move these back and bump the `setup.py` version number, everything seems to install correctly.

chrisndodge: It was more pre-emptive, I wasn't hitting a conflict, but I just don't always want to have to keep updating my requirements when edx-platform does.

It seems like ripe for human error.

chrisndodge:
What's the context here for this PR?

requirements.txt
chrisndodge: Actually I mentioned this challenge to a few people and it doesn't seem like we have a best practice around library dependency sychronization.

wedaly:
This is pre-work for the Django 1.8 upgrade.

setup.py
```diff
@@ -34,7 +34,7 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-proctoring',
-    version='0.6.2',
+    version='0.7.2',
```
chrisndodge: Hi Will, this is going to run in conflict on a commit that I have on rc/2015-08-25 which bumps the version for this upcoming release. I haven't completely ironed out my release process, but I'm setting the setup.py version on the rc branches and then merging those back to master.

If you need to bump it up, can you make it 0.6.3 for now and I'll deal with the merge conflict. The version on the rc is 0.7.1

PR #119, opened 2015-08-25
PHX-120

muhhshoaib:
updated the dialog box for the Open exam options

FYI @chrisndodge

PR #126, opened 2015-08-26
PHX-133/avoid multiple click

hasnain-naveed:


edx_proctoring/templates/proctoring/seq_proctored_exam_entrance.html
```diff
@@ -64,6 +72,7 @@
       if (typeof action_url === "undefined" ) {
         return false;
```
muhhshoaib: bind the click event back here.

edx_proctoring/templates/proctoring/seq_proctored_exam_entrance.html
```diff
@@ -38,19 +38,27 @@
...
       var start_immediately = $(this).data('start-immediately');
 
+      // unbind the click event to prevent the multiple calls
+      $(this).unbind('click');
+
+      // adding class for moving cursor and to change the color of button
+      $(this).parent().addClass('disable-proctored-exam-btn');
+
       if (!attempt_proctored) {
         var msg = gettext(
```
muhhshoaib: bind the click event back here.

chrisndodge:
@hasnain-naveed @muhhshoaib I have a feeling that these changes will conflict with: https://github.com/edx/edx-proctoring/pull/119

We'll have to resolve merge conflicts after 119 gets merged in (or vice versa)

muhhshoaib:
@chrisndodge yes, we do have in mind. Just waiting for the first one to merge then we can resolve the conflicts.

chrisndodge:
@hasnain-naveed @muhhshoaib @afzaledx to unblock us from having to get Sass changes in edx-platform reviewed, is there a way we could do this all in JavaScript (and thus all self-contained in edx-proctoring)?

hasnain-naveed:
@chrisndodge @muhhshoaib @afzaledx I have done the css in JavaScript.

chrisndodge:
@hasnain-naveed @muhhshoaib this basically looked good, but I added some stuff (mainly applied the same javascript to the practice and timed-only exam .html templates). Also, I removed the setting of the link colors, as I think just using a spinner cursor should suffice for now. Also, I noticed that the spinner cursor disappeared if the end user moved the pointer to other areas on the webpage, so I defined a style definition to the whole body.

I'll merge this in when tests pass and close the corresponding PR in edx-platform.

chrisndodge:
Sorry, forgot to mention, I added a fail() callback handler on the ajax to display an error message to the user (as opposed to silently failing and the user not knowing what is happening). While it wasn't part of the story, we had a report of a failure for a user who hadn't activated his/her account (which will fail the API authentication) and contacted edX support as things "weren't working" and they didn't know why and how to fix it.

PR #130, opened 2015-08-31
PHX-122

muhhshoaib:
@afzaledx  @chrisndodge Kindly review the changes for PHX-122.

Fixed the issue when deleting a proctored exam attempt does not remove the record in the CreditRequirementStatus table

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -666,12 +666,27 @@ def remove_exam_attempt(attempt_id):
         raise StudentExamAttemptDoesNotExistsException(err_msg)
 
     username = existing_attempt.user.username
+    user_id = existing_attempt.user.id
     course_id = existing_attempt.proctored_exam.course_id
     content_id = existing_attempt.proctored_exam.content_id
+    to_status = existing_attempt.status
+
     existing_attempt.delete_exam_attempt()
     instructor_service = get_runtime_service('instructor')
+
     if instructor_service:
         instructor_service.delete_student_attempt(username, course_id, content_id)
+
+    # see if the status transition this changes credit requirement status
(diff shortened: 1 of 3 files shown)


edx_proctoring/api.py
```diff
@@ -666,13 +666,36 @@ def remove_exam_attempt(attempt_id):
...
+        else:
+            verification = u'failed'
+
+        credit_service.remove_credit_requirement_status(
+            user_id=user_id,
+            course_key_or_id=course_id,
+            req_namespace=u'proctored_exam',
+            req_name=content_id,
+            status=verification
+        )
```
chrisndodge: See note in edx-platform PR about making status parameter optional. I don't think we should compute (up above) what we thing the 'verification' parameter should be. We should simply naively delete any row associated with the user.

edx_proctoring/tests/test_api.py
```diff
@@ -482,6 +482,25 @@ def test_remove_exam_attempt(self):
...
+        """
+
+        proctored_exam_student_attempt = self._create_unstarted_exam_attempt()
+        # now update the status of the exam.
+
+        proctored_exam_student_attempt.status = to_status
+        proctored_exam_student_attempt.save()
+        # now remove exam attempt which calls the credit service method 'remove_credit_requirement_status'
+        remove_exam_attempt(proctored_exam_student_attempt.id)
+
```
chrisndodge: You're not asserting in the test that the requirement_status data was actually removed here :-)

I'll add it so we can merge it in (provided the corresponding change in edx-platform has been +1'd)

chrisndodge:
@muhhshoaib this looks good but your unit tests didn't really assert that the requirement was removed. I updated the test as well as the Mocked credit service to help better simulate it.

PR #136, opened 2015-09-07
SOL-1171

muhhshoaib:
Update timer bar so that it doesn't always refer to a proctored exam


chrisndodge:
+1.

Can you squash commits and merge. Then re-kick off the bokchoy build in edx-platform and see if that test passes.

muhhshoaib:
Squashed and merged.

PR #154, opened 2015-09-16
Cdodge/rdf 3.1 rebase2

chrisndodge:


chrisndodge:
@wedaly I think this is the right PR, I'll have to tag this as 0.9.8 once it merged to the RC branch

PR #159, opened 2015-09-17
DRF Upgrade support

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib @hasnain-naveed FYI when this lands into master you will have to re-run all requirements to pick up the new version of Django Rest Framework into your working virtualenv.

muhhshoaib:
@chrisndodge sure we will keep in mind.

PR #163, opened 2015-09-18
add some protection about not transition from a completed state to an…

chrisndodge:
… error state


chrisndodge:
@afzaledx @muhhshoaib can one of you two review. I'm thinking about hotpatching edx-proctoring next week. I saw a potential vulnerability where the status could potentially transition from a 'submitted' state to an 'error' state. Not sure if it's happened in production, but we shouldn't allow for that.

Note this is a PR against release and I've bumped the version up to 0.9.7

PR #165, opened 2015-09-20
Rc/2015 09 22

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib can one of you do a quick review. We need to get this into this upcoming week's release candidate. It addresses a few concerns we are having with proctoring that have come up over the weekend. Thanks.

edx_proctoring/management/commands/set_attempt_status.py
```diff
@@ -0,0 +1,64 @@
...
+                to_status=to_status
+            )
+        )
+        print msg
+
+        if not ProctoredExamStudentAttemptStatus.is_valid_status(to_status):
+            raise Exception('{to_status} is not a valid attempt status!'.format(to_status=to_status))
+
+        # get exam, this will throw exception if does not exist, so let it bomb out
+        get_exam_by_id(exam_id)
```
afzaledx: line 531 of api.py is raising the exception so is this call necessary to raise just the exception?

afzaledx:
:+1:  except for the additional line to fetch the exams which you might want to remove. I didn't actually get the logic behind the line.

edx_proctoring/management/commands/set_attempt_status.py
chrisndodge: yea, basically. I guess there's a better way to do that. I'll change it after the release, because this hash is already tagged 0.9.6d and edx-platform is now pointing to it

edx_proctoring/management/commands/set_attempt_status.py
```diff
@@ -0,0 +1,64 @@
...
+from optparse import make_option
+
+from django.core.management.base import BaseCommand
+
+from edx_proctoring.models import ProctoredExamStudentAttemptStatus
+
+
+class Command(BaseCommand):
+    """
+    Django Management command to force a background check of all possible notifications
```
muhhshoaib: update the docstring.

PR #168, opened 2015-09-22
Return only active exams when asked for a list of proctored exams.

afzaledx:


afzaledx:
@chrisndodge, @muhhshoaib

chrisndodge:
@afzaledx we now have a merge conflict. I'll resolve and force push.

chrisndodge:
@muhhshoaib can you do an initial review?

edx_proctoring/api.py
```diff
@@ -811,7 +811,7 @@ def get_all_exams_for_course(course_id):
     ..
     ]
     """
-    exams = ProctoredExam.get_all_exams_for_course(course_id)
+    exams = ProctoredExam.get_all_exams_for_course(course_id, active_only)
```
muhhshoaib: I think we have to pass `active_only` as a keyword argument like `active_only= active_only`

edx_proctoring/api.py
```diff
@@ -787,7 +787,7 @@ def remove_exam_attempt(attempt_id):
         )
 
 
-def get_all_exams_for_course(course_id):
+def get_all_exams_for_course(course_id, active_only=False):
```
muhhshoaib: We are already passing keyword argument in the models method. I think we can just pass simple parameter here. We are following the same structure elsewhere.

muhhshoaib:
added some comments. Otherwise looks good to me :+1:

edx_proctoring/api.py
afzaledx: active_only seems more readable than is_active.

edx_proctoring/api.py
afzaledx: Made the change.

chrisndodge:
@afzaledx I looked and there's no corresponding PR in edx-platform right? If so, please squash commits and merge.

chrisndodge:
@afzaledx can you rebase and resolve conflicts. Looking at it, I'm not 100% how to resolve it myself. It's not too bad of a merge conflict. Then squash and merge. Thanks.

PR #171, opened 2015-09-23
pull examination review policy from configuration and pass to softwar…

chrisndodge:
…e secure


chrisndodge:
@afzaledx @muhhshoaib can you review?

chrisndodge:
Hold off on review. DevOps is requesting I take a different approach...

PR #172, opened 2015-09-23
allow for optional (default False) ability to allow SoftwareSecure to…

chrisndodge:
… retransmit proctored results


chrisndodge:
@muhhshoaib @afzaledx @hasnain-naveed one more for someone to review.

We're seeing some trouble with software secure failing tests that we'd like to see passed. Our current implementation here is that we don't allow for (for integrity reasons) to allow for updates. However - in hindsight - that might have been overly restrictive.

So this PR allows for an optional configuration to allow for accepting proctored reviewed results even if we've received a previous one.

afzaledx:
:+1:  LGTM. Is the functionality to delete a review attempt going to be added to PHX-158? I saw the delete signal for archiving but I am not sure if we're using this flow other than the tests.

PR #173, opened 2015-09-23
fixed the bug with invalid footer link in the verified html template file

muhhshoaib:
@chrisndodge can you take a quick review on this one.

### DIFF
--- a/edx_proctoring/templates/proctored_exam/verified.html
+++ b/edx_proctoring/templates/proctored_exam/verified.html
@@ -26,4 +26,4 @@
     {% endblocktrans %}
   </p>
 </div>
-{% include 'proctored_exam/seq_proctored_exam_footer.html' %}
+{% include 'proctored_exam/footer.html' %}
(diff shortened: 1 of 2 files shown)


chrisndodge:
Thanks! Is there a unit test that tries to render this verified.html template? It seems like this should have caused a test failure. Can you add a unit test that would have caught this error?

chrisndodge:
Hrm. seems like there is a unit test for rendering a verified status, so I'm wondering why it didn't catch it. Let me look into it.

chrisndodge:
OK, it seems like a failure to find an included template will not throw an exception - which is why the unit tests did not fail.

I'm going to add some content checking tests to make sure the footer content is visible in the HTML.

chrisndodge:
Added tests to assert that the footer content is in the various templates. Will merge when build passes. Thanks for noticing and submitting the PR. Surprised that a missing template include would not cause an unhandled exception.

PR #174, opened 2015-09-23
(WIP) As an edX staff member, I would like to be able to review all submitted proctoring sessions and change the disposition that SoftwareSecure sets

muhhshoaib:


chrisndodge:
I'm going to close this for now (but keep the branch around) as it doesn't seem like we'll be moving forward with this feature for now.

PR #175, opened 2015-09-23
remove unused templates

chrisndodge:


chrisndodge:
@muhhshoaib @afzaledx @hasnain-naveed I don't believe we are using these templates any more (I did a code search). However, seeing how there are no exceptions thrown when sub-templates are %included, it'd be great if one of you could review this

muhhshoaib:
:+1:

PR #185, opened 2015-10-01
(WIP) Added a new state - Exam Not Submitted Yet.

afzaledx:


edx_proctoring/api.py
```diff
@@ -652,7 +654,7 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
                 cascade_effects=False
             )
 
-    if to_status == ProctoredExamStudentAttemptStatus.submitted:
+    if to_status == ProctoredExamStudentAttemptStatus.waiting_for_app_timeout:
```
chrisndodge: sorry to suggest this while you are so far through, but the term "timeout" is a bit negative. Can we use waiting_for_app_shutdown instead? Hopefully, it's just a search & replace operation.

edx_proctoring/api.py
```diff
@@ -933,6 +935,11 @@ def _check_eligibility_of_prerequisites(credit_state):
         'suggested_icon': 'fa-unlock',
         'in_completed_state': False
     },
+    ProctoredExamStudentAttemptStatus.waiting_for_app_timeout: {
+        'short_description': _('Exam Not Submitted'),
```
chrisndodge: This is pretty negative sounding. How about "Exam pending submission"?

edx_proctoring/api.py
```diff
@@ -962,6 +969,11 @@ def _check_eligibility_of_prerequisites(credit_state):
         'suggested_icon': 'fa-lock',
         'in_completed_state': False
     },
+    ProctoredExamStudentAttemptStatus.waiting_for_app_timeout: {
+        'short_description': _('Exam Not Submitted'),
```
chrisndodge: This is pretty negative sounding. How about "Exam pending submission"?

edx_proctoring/api.py
```diff
@@ -1186,6 +1198,19 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
...
+                SOFTWARE_SECURE_CLIENT_TIMEOUT:
+            update_attempt_status(
+                exam_id,
+                user_id,
+                ProctoredExamStudentAttemptStatus.submitted
+            )
+            attempt['status'] = ProctoredExamStudentAttemptStatus.submitted
+            student_view_template = 'practice_exam/submitted.html'
+        else:
+            student_view_template = 'proctored_exam/waiting_for_app_timeout.html'
```
chrisndodge: This seems like duplicate code compared to what is in _get_proctored_exam_view(). Can we make s shared helper method?

edx_proctoring/callbacks.py
```diff
@@ -159,3 +160,24 @@ def get(self, request, attempt_code):  # pylint: disable=unused-argument
...
+        about the exam
+        """
+        return Response(
+            data={
+                # IMPORTANT: Don't add more information to this as it is an
+                # unauthenticated endpoint
+                'status': 'submitted',
+            },
+            status=200
+        )
```
chrisndodge: Hrm, can you explain this test endpoint? Is it for unit tests, or doing some manual tests? If it is for unit tests, I'd prefer some flag to expose this API as this will be a public API endpoint. I see that it doesn't really "do anything" (i.e. modify the database), but I'd rather not have this exposed.

edx_proctoring/models.py
```diff
@@ -114,6 +114,9 @@ class ProctoredExamStudentAttemptStatus(object):
     # and we will not allow transitions to status above this mark
     #
 
+    # waiting for the student to close the software secure client window.
+    waiting_for_app_timeout = 'waiting_for_app_timeout'
```
chrisndodge: suggested name change at the top of the PR

edx_proctoring/static/proctoring/js/views/proctored_exam_attempt_view.js
```diff
@@ -11,6 +11,7 @@ var edx = edx || {};
         ready_to_start: gettext('Ready to start'),
         started: gettext('Started'),
         ready_to_submit: gettext('Ready to submit'),
+        waiting_for_app_timeout: gettext('Waiting on App'),
```
chrisndodge: I think we should display 'Pending Submission'

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -198,13 +198,16 @@
...
         $.ajax(url).success(function(data){
           // has the end user completed and submitted the exam in the LMS?!?
-          if (data.status === 'submitted' || data.status === 'error' || data.status === 'verified' || data.status === 'rejected') {
+          if (data.status === 'waiting_for_app_timeout' || data.status === 'error' || data.status === 'verified' || data.status === 'rejected') {
             // Signal that the desktop software should terminate
             // NOTE: This is per the API documentation from SoftwareSecure
             window.external.quitApplication();
+            $.ajax(test_url).success(function(data){
+              // action
+            });
```
chrisndodge: Are you still working on this area of the code?

edx_proctoring/urls.py
```diff
@@ -85,5 +85,10 @@
         callbacks.AttemptStatus.as_view(),
         name='edx_proctoring.anonymous.proctoring_poll_status'
     ),
+    url(
+        r'edx_proctoring/test_api/$',
+        callbacks.StudentProctoredExamAttempt.as_view(),
+        name='edx_proctoring.anonymous.test_api'
+    ),
```
chrisndodge: back to my question on what this is for.

chrisndodge:
@afzaledx can you give an update on this PR? Have we validated if the Javascript blocks when the quitApplication() is called and prompts the user to confirm? Thanks.

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
chrisndodge: Have you confirmed that this area of the code will be blocking waiting for the user to acknowledge the modal dialog?

afzaledx:
@chrisndodge. It's kind of tricky about the callback from the proctoring application. the quitApplication() is blocking - that's for sure. One more thing we know about the response is that if the user presses "Cancel" on the popup confirming closing the application, we get an undefined in the response, and some truthy value if the user chooses to quit.

There are some tricky aspects to it as well. i.e. we get an unknown amount time before the app terminates i.e. it doesn't necessarily let us finish the call. For most of our testing, we've been able to receive the callback but we'll need to test it out a bit more. The endpoint as well as the ajax call section is incomplete. was left there just for various tests we've been doing. Will sit with Hasnain to finish it today.

edx_proctoring/callbacks.py
afzaledx: The idea here is that the page opened in the proctoring application will call back into the backend to tell the backend that it has shut down. We'll modify the status of the attempt to 'submitted' if it's in the 'waiting for app shutdown' state. This one is here for manual testing, but we'll probably merge it with the AttemptStatus endpoint, passing a query parameter
 ?action=shutdown.

edx_proctoring/api.py
```diff
@@ -43,6 +43,8 @@
 from edx_proctoring.backends import get_backend_provider
 from edx_proctoring.runtime import get_runtime_service
 
+SOFTWARE_SECURE_CLIENT_TIMEOUT = settings.PROCTORING_SETTINGS.get('SOFTWARE_SECURE_CLIENT_TIMEOUT', 30)
```
chrisndodge: can we put this in constants.py?

hasnain-naveed:
@chrisndodge, kindly take a look.

afzaledx:
@chrisndodge, kindly just look at the changes in the `proctoring_launch_callback.html` and the `AttemptStatus(APIView)`'s `get()` in callbacks.py
I think you've already gone through the rest of the changes. Hasnain is going to try this sequence 10 times with both shutdown and cancel buttons to see the behavior if we receive the hit for shutdown.

edx_proctoring/api.py
```diff
@@ -305,14 +305,18 @@ def update_exam_attempt(attempt_id, **kwargs):
     exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_exam_attempt_by_id(attempt_id)
     for key, value in kwargs.items():
         # only allow a limit set of fields to update
-        # namely because status transitions can trigger workflow
-        if key not in ['last_poll_timestamp', 'last_poll_ipaddr']:
+        # namely because status t ransitions can trigger workflow
```
muhhshoaib: remove the space in `t ransitions`

edx_proctoring/api.py
```diff
@@ -1058,6 +1062,11 @@ def _resolve_prerequisite_links(exam, prerequisites):
         'suggested_icon': 'fa-pencil-square-o',
         'in_completed_state': False
     },
+    ProctoredExamStudentAttemptStatus.waiting_for_app_shutdown: {
```
muhhshoaib: ProctoredExamStudentAttemptStatus.waiting_for_app_shutdown is defined multiple times. Remove the other one

afzaledx:
@chrisndodge, after doing a lot of testing, we have concluded that the implementation where we wait for the response from quitApplication(), and then send another ajax call notifying shutdown - doesnt work.

On the other hand, if we don't have that functionality, it takes 30 seconds after the user has quit the app to take us to the submitted page.

One other way is to to reduce the timeout duration - not globally - but just for the status waiting_for_shutdown to 10 seconds from the otherwise 30 seconds. Trying with that now.

afzaledx:
There's another possibility that we ask the Software secure developers to add such a functionality the same way they validate/create the exam attempt - i.e. to notify us of shutdown - if it is feasible.

chrisndodge:
@afzaledx thanks for the update, it will be hard to assume that we can get SoftwareSecure to commit to doing any changes, so I think we should come up with a solution that does not require any changes to the client.

Another tact might be to keep just the 'submitted' state. When rendering the initial 'submitted' content , the server will look at 'last_poll_timestamp' and see if it was > 10 seconds. If < 10 then it will render HTML content which instructs the student to make sure to close the client app. Then - on the client side - that screen will poll the server every 5 seconds to get the last time the client "called home". When that timestamp transitions > 10 seconds then we reload the page which will then show the existing 'submitted' content (aka talking about submitted exam and results will take 24-48 hours).

This approach will probably work, the only thing we loose is the database persistence status of that intermediate step. Also, it will undo a lot of your work here, so that might be a bit of a bummer...

afzaledx:
@chrisndodge, closing this PR now. Although you were right about discarding the effort made for this branch, but your suggestion really made sense. The new branch is `hasnain-naveed/PHX-149-without-new-state`. Hasnain is just finishing with test coverage and will be pushing the code in a few minutes.

PR #189, opened 2015-10-05
fixed the bug

muhhshoaib:
@chrisndodge found a bug related to the timed_exam_only parameter in the api.py methods that's get called from the platform repository. 

- added the default value to the timed_exam_only parameter in the api methods


chrisndodge:
Thanks @muhhshoaib but can you write a unit test the reproduces the failure and then - after applying this fix - shows it working.

muhhshoaib:
Merged into master but i'll create a new PR for the test cases.

PR #192, opened 2015-10-06
only declines should cascade downwards. If a previous ICRV failed, th…

chrisndodge:
…en we should not auto-decline, but rather display the failed prerequisites to the user

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -643,11 +643,11 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, 
 
         # we just want other exams which are proctored and are not practice
         exams = [
-            exam
-            for exam in _exams
+            _exam
+            for _exam in _exams
             if (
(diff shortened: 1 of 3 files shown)


chrisndodge:
@afzaledx @muhhshoaib can one of you do an internal review. Basically now that we have the notion of pre-requisites, if we have two Proctored Exams and we fail the first one, we shouldn't simply "auto-decline" the 2nd one, but rather the user will see the screen saying that the 1st one failed and the user has no option but to explicitly decline (as a mouse click) the 2nd one.

We're doing this because when we auto-decline, it's causing some confusion because we can't distinguish a decline as being user-initiated (via mouse click) and the system auto-declining. I think declines should be an explicit action that a user takes.

afzaledx:
:+1: . Just 1 clarification needed : if the user explicitly declines, this would cascade-decline all other proctored exams. Right?

chrisndodge:
correct

PR #196, opened 2015-10-09
make sure default ordering is by reverse update field

chrisndodge:
@afzaledx @muhhshoaib small change regarding admin.py screens


muhhshoaib:
:+1:

PR #198, opened 2015-10-12
PHX-171/Added due date

hasnain-naveed:


edx_proctoring/api.py
```diff
@@ -126,6 +127,8 @@ def update_exam(exam_id, exam_name=None, time_limit_mins=None,
         proctored_exam.exam_name = exam_name
     if time_limit_mins is not None:
         proctored_exam.time_limit_mins = time_limit_mins
+    if due_date is not None:
+        proctored_exam.due_date = due_date
```
chrisndodge: Hi, ironically we don't want to use None here to mean "do not update", because the caller might want to actually set due_date to None intentionally. Can you come up with with another default value in the call signature to mean "do not update", sometimes in this situation, I use a "minumum date" constant. I don't know if Python has something like that, can you research? Let me know if you don't know what I mean...

chrisndodge:
@afzaledx @muhhshoaib can one of you two review? Also be sure to squash commits before merging. Thanks.

chrisndodge:
squashed commits

chrisndodge:
@afzaledx @hasnain-naveed I'm going to merge this for now with the assumption that PHX-172 will be worked on next

PR #200, opened 2015-10-13
store declines in the CreditRequirementStatus table

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib can you do a review. This change is for - when a user declines a proctored exam - that we mark the CreditRequirementStatus table as 'declined' (we used to store 'failed'). Also if a user declines any of the prerequisites, then we auto-decline the user when he/she comes upon the next proctored exam.

edx_proctoring/api.py
```diff
@@ -1386,7 +1394,23 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
...
+                # so we can mark it as declined
+                create_exam_attempt(exam_id, user_id)
+
+                update_attempt_status(
+                    exam_id,
+                    user_id,
+                    ProctoredExamStudentAttemptStatus.declined,
+                    raise_if_not_found=False
+                )
+                return None
```
afzaledx: Since we are retruning None from here, the user will be taken directly to the exam. He will not be shown a screen telling him that a prereq was declined, that's why he's not being shown the proctoring option. Just wanted to make sure that this flow is OK.

afzaledx:
:+1:  other than the concern mentioned above.

edx_proctoring/api.py
chrisndodge: Yes, that is intentional. Hopefully we won't change our mind on that :-)

PR #201, opened 2015-10-15
PHX-172/template added if due date for exam has passed

hasnain-naveed:


afzaledx:
:+1:

edx_proctoring/tests/test_api.py
```diff
@@ -1144,6 +1147,91 @@ def test_get_studentview_started_timed_exam(self):
         )
         self.assertIsNone(rendered_response)
 
+    def test_practice_exam_attempt_with_past_due_datetime(self):
+        """
+        Test for get_student_view for proctored practice exam with past due datetime
```
muhhshoaib: update the docstrings. should be practice proctored exam

edx_proctoring/tests/test_api.py
```diff
@@ -1144,6 +1147,91 @@ def test_get_studentview_started_timed_exam(self):
...
+                    'is_proctored': True,
+                    'display_name': self.exam_name,
+                    'default_time_limit_mins': self.default_time_limit
+                }
+            )
+            self.assertIn(self.exam_past_due_date_msg, rendered_response)
+
+    def test_timed_exam_attempt_with_past_due_datetime(self):
+        """
+        Test for get_student_view for proctored exam with past due datetime
```
muhhshoaib: change proctored exam to timed exam

edx_proctoring/templates/proctored_exam/expired.html
```diff
@@ -0,0 +1,20 @@
...
+    {% endblocktrans %}
+  </p>
+  <hr>
+  <p>
+    {% blocktrans %}
+      View your credit eligibility status on your <a href="{{progress_page_url}}">Progress</a> page.
+    {% endblocktrans %}
+  </p>
+</div>
+{% include 'proctored_exam/footer.html' %}
```
chrisndodge: @mhoeber can you do a review on this little bit of content? Use case here is when a user clicks on a proctored exam when it is past the exam due date.

chrisndodge:
@hasnain-naveed @afzaledx @muhhshoaib FYI, I made some edits here. I've added a new 'expired' Status so that we record in the attempt that the user visited the exam past the due date.

edx_proctoring/templates/proctored_exam/expired.html
mhoeber: @chrisndodge   I'd just change: "Since the due date for this exam has already passed, you are no longer able to take this examination."

to:

"Because the due date has passed, you are no longer able to take this exam."

(I try to avoid "since" as it can have two meanings)

chrisndodge: thanks. updated.

PR #211, opened 2015-10-28
PHX-195 Adding a non-integer value for the time extension allowance will cause an exception to be thrown

muhhshoaib:
@chrisndodge Kindly review the changes for PHX-195.
- fixed the bug when adding a non-integer value for the time extension allowance will not raise any exception

### DIFF
--- a/edx_proctoring/backends/tests/test_software_secure.py
+++ b/edx_proctoring/backends/tests/test_software_secure.py
@@ -144,7 +144,7 @@ class SoftwareSecureTests(TestCase):
             self.assertEqual(attempt['external_id'], 'foobar')
             self.assertIsNone(attempt['started_at'])
 
-    @ddt.data(None, 'additional person allowed in room')
+    @ddt.data(None, '10')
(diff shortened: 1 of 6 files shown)


chrisndodge:
Hi @muhhshoaib thanks for taking this on. I must not have been very clear in the JIRA, but we do need to be able to allow for strings for some allowances. For example, there's an "exam policy exception" allowance where a course team member can add in additional information that gets passed to the Proctored Exam review.

This is the problem, some allowances should be integers and some can be allowed to be strings.

We could take this approach that you wrote, but we'll need to add in Python type information mapping information, e.g. "additional time must be an Integer" and then use that information whether to do validation or not. I guess you can add a classmethod to the Allowance class which does the validation...

muhhshoaib:
@chrisndodge I have updated the code. Kindly review this and let me know if this is the desired functionality.

chrisndodge:
@muhhshoaib thanks. Looks good to me. Merging

PR #217, opened 2015-11-03
SOL-1360 Timed Exams should allow learners into the exam after submission

muhhshoaib:
@afzaledx can you do the initial review.

@chrisndodge 
- alter the logic on the timed exam view to allow learners into the exam after submission
- update the message when the due date is not None and it has not been passed

### DIFF
--- a/edx_proctoring/templates/timed_exam/submitted.html
+++ b/edx_proctoring/templates/timed_exam/submitted.html
@@ -10,5 +10,10 @@
     {% blocktrans %}
       Your grade for this timed exam will be immediately available on the <a href="{{progress_page_url}}">Progress</a> page.
     {% endblocktrans %}
+    {% if has_due_date %}
+      {% blocktrans %}
+        After the due date has passed, you can review the exam, but you cannot change your answers.
+      {% endblocktrans %}
+    {% endif %}
   </p>
 </div>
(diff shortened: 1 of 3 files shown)


chrisndodge:
@griffresch @catong @mhoeber there's some message changes here in the HTML template. Can you review as we just came up with this content ourselves?

edx_proctoring/templates/timed_exam/submitted.html
```diff
@@ -10,5 +10,10 @@
     {% blocktrans %}
       Your grade for this timed exam will be immediately available on the <a href="{{progress_page_url}}">Progress</a> page.
     {% endblocktrans %}
+    {% if has_due_date %}
+      {% blocktrans %}
+        Once the exam due date has passed, you will be able to once again see your submitted exam answers but not change them.
```
mhoeber: "After time has run out for the exam, you can review the exam, but you cannot change your answers."

edx_proctoring/tests/test_api.py
```diff
@@ -103,6 +103,7 @@ def setUp(self):
         self.start_an_exam_msg = 'This exam is proctored'
         self.exam_expired_msg = 'The due date for this exam has passed'
         self.timed_exam_msg = '{exam_name} is a Timed Exam'
+        self.submitted_timed_exam_msg_with_due_date = 'Once the exam due date has passed,'
```
mhoeber: we should avoid "once" - has multiple meanings.  
"After the exam due date has passed,"

edx_proctoring/templates/timed_exam/submitted.html
chrisndodge: Thanks, but "time has run out" is a bit ambiguous because there are two time gates here: the allowed minutes a user has to complete the exam as well as a 'due date'. This message is specific to the due date having to pass before the end-user can see the exam.

Would this be an OK compromise?

"After the due date has passed, you can review the exam, but you cannot change your answers."

mhoeber: @chrisndodge my edits and your comments crossed

PR #219, opened 2015-11-03
Cdodge/style fixes

chrisndodge:


chrisndodge:
@frrrances I think this is all that we need on the edx-platform side. There's another set of diffs at: https://github.com/edx/edx-proctoring/pull/219

chrisndodge:
I've refreshed the sandbox with this PR (but I haven't smoke tested it yet)

frrrances:
:+1:

PR #220, opened 2015-11-04
SOL-1359 Incorrect label - Adding an allowance

muhhshoaib:
@afzaledx can you do the initial review. It's a simple fix.

@chrisndodge 
- fixed the label when adding an allowance


chrisndodge:
@muhhshoaib, thanks for the tiny change :-)

PR #230, opened 2015-11-13
add better filter definitions for the SoftwareSecure Proctored Review…

chrisndodge:
… admin page

### DIFF
--- a/edx_proctoring/admin.py
+++ b/edx_proctoring/admin.py
@@ -3,10 +3,15 @@ Django Admin pages
 """
 # pylint: disable=no-self-argument, no-member
 
+import pytz
+from datetime import datetime, timedelta
+
+from django.db.models import Q
 from django.contrib import admin
 from django.utils.translation import ugettext_lazy as _
 from django import forms
 from edx_proctoring.models import (
+    ProctoredExam,
     ProctoredExamReviewPolicy,
     ProctoredExamSoftwareSecureReview,
     ProctoredExamSoftwareSecureReviewHistory,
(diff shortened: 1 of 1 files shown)


chrisndodge:
@afzaledx @muhhshoaib can someone do a review on this? The purpose is to apply better filtering so that the right hand "quick filters" don't get cluttered up with inactive, timed/practice exams, or old exams. Otherwise, the list of filters are going to get too large over time.

chrisndodge:
@afzaledx @muhhshoaib just a nudge on the PR :-)

edx_proctoring/admin.py
```diff
@@ -104,13 +108,121 @@ def queryset(self, request, queryset):
...
+            is_active=True,
+            is_practice_exam=False,
+            due_date__gte=thirty_days_ago,
+            due_date__lte=one_week_from_now
+        )
+
+        lookups = (())
+
+        for exam in exams:
+            lookups += ((exam.id, exam.exam_name),)
```
afzaledx: Would it be better showing just the exam name? We could end up having three "Midterm"s in the list, not knowing which course it belongs to.

afzaledx:
:+1:  other than the small thing I noticed since I am not aware how the exams are named in the actual courses.

If we can somehow identify the course just by looking at the exam name (e.g. it's named as 'Introduction to Solar Systems Astronomy - Midterm Exam', then this wont matter at all. Otherwise if the exam name looks like 'Midterm Exam", then a lot of active courses with proctored exams running at the same time would render this filter totally useless.

P.s. I have worked with admin before but this was my first time with custom filters. This is a very nice way to add filters. Good learning for me.

chrisndodge:
Thanks @afzaledx I think the best we could do here is modify api.y "create/update" exam call endpoints to have an (optional) display course name passed in which would be stored in the exam tables.

Either that, or we'd have to add an additional endpoint to our 'services' so that edx-proctoring can call into edx-platform to resolve the Display Names for a list of course_id's.

Either way, there would have to be a corresponding change in edx-platform.

Let me think about it for a little bit.

- Chris

afzaledx:
Yea, the entire course ID string wasn't looking very nice in the list of courses either. Can we parse the display name out of the course id?

chrisndodge:
rebased against Django 1.8 merge

chrisndodge:
@afzaledx added some parsing of the course_id to give a slightly better exam name context

chrisndodge:
@afzaledx @muhhshoaib just a nudge on this one. Thanks.

afzaledx:
:+1: . We could also use the same parsing for the other filter (list of courses), but you can make that decision.

PR #234, opened 2015-11-18
Update the call to the utility method to expect a tuple to be returned

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib this is a bug found in production

muhhshoaib:
:+1:

PR #235, opened 2015-11-23
don't allow a blank exam name to be registered with SoftwareSecure ot…

chrisndodge:
…herwise the API call will fail


chrisndodge:
@afzaledx @muhhshoaib @ibrahimahmed443 I noticed this when trying to register an exam in Chinese. The examName ended up being an empty string, so in that case we need to use some default name.

muhhshoaib:
:+1:

PR #237, opened 2015-11-25
address feedback from analytics team

chrisndodge:


chrisndodge:
@mulby this PR addresses your feedback on the PR to edx-platform where we directly import/depend on event-tracker repo. This PR will take the place of the PR to edx-platform, so if we're OK with this, I'll close the other one.

FYI @afzaledx @muhhshoaib @smagoun

chrisndodge:
@afzaledx can one of you do an internal review. In short, it's intent is to get rid of the PR in edx-platform and call directly into the event-tracking repository.

mulby:
Looks reasonable to me.

:+1:

PR #238, opened 2015-11-26
do the timestamp changes at the same time we set the status

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib I noticed this inefficiency where we are calling .save() many times. This just consolidates it.

muhhshoaib:
:+1:  LGTM

PR #239, opened 2015-11-26
change over any http: links over to https:

chrisndodge:


chrisndodge:
@afzaledx @muhhshoaib I just noticed that SoftwareSecure will post back video_url_links which are http: and not https:

We should always use https: when linking over to SoftwareSecure

muhhshoaib:
:+1: LGTM

PR #241, opened 2015-12-07
PHX-216 / user can visit the exam once due has passed.

hasnain-naveed:


chrisndodge:
@afzaledx can someone do initial review? Thanks

FYI @douglashall

edx_proctoring/views.py
```diff
@@ -779,3 +780,30 @@ def get(self, request):
...
+        Update the is_status_acknowledge flag for the specific attempt
+        """
+        try:
+            update_exam_attempt(attempt_id, is_status_acknowledged=True)
+
+            return Response(
+                status=status.HTTP_200_OK
+            )
+
+        except ProctoredBaseException, ex:
```
muhhshoaib: `update_exam_attempt` sends `StudentExamAttemptDoesNotExistsException`. change the exception to `StudentExamAttemptDoesNotExistsException `

edx_proctoring/views.py
```diff
@@ -779,3 +780,30 @@ def get(self, request):
             user_id=request.data.get('user_id', None),
             course_id=request.data.get('course_id', None)
         ))
+
+
+class ProctoredExamAttemptReviewStatus(AuthenticatedAPIView):
+    """
+    Endpoint for the Active Exams for a user.
```
muhhshoaib: update the docstring. may be something `Endpoint for updating exam attempt review status to acknowledged`

edx_proctoring/tests/test_views.py
```diff
@@ -1877,6 +1877,50 @@ def test_review_callback_get(self):
...
+
+        response = self.client.put(
+            reverse(
+                'edx_proctoring.proctored_exam.attempt.review_status',
+                args='5'
+            ),
+            {},
+            content_type='application/json'
+        )
+        self.assertEqual(response.status_code, 400)
```
muhhshoaib: I think it's better to move the 400 case into a separate test case where you can actually validates the exception that has been raised when hitting the url with the invalid attempt_id
`with self.assertRaises(StudentExamAttemptDoesNotExistsException)`

edx_proctoring/templates/proctored_exam/visit_exam_content.html
```diff
@@ -0,0 +1,31 @@
+{% load i18n %}
+{% if has_due_date_passed %}
+
+  <button type="button" style="margin-bottom: 15px;" name="visit-exam-content" class="visit-exam-button btn btn-pl-primary btn-base" data-action-url="{{update_is_status_acknowledge_url}}">
```
muhhshoaib: don't use inline css.

hasnain-naveed: it is just for time being, because we don't have the exact requirement

muhhshoaib:
Once comments addressed :+1:

edx_proctoring/views.py
```diff
@@ -779,3 +780,30 @@ def get(self, request):
...
+
+    Supports:
+        HTTP PUT: Update the is_status_acknowledge flag
+    """
+    def put(self, request, attempt_id):     # pylint: disable=unused-argument
+        """
+        Update the is_status_acknowledge flag for the specific attempt
+        """
+        try:
+            update_exam_attempt(attempt_id, is_status_acknowledged=True)
```
chrisndodge: We should assert here that the logged in user actually is the rightful owner of the exam attempt, otherwise someone could write a script (using a valid login session_id) which just goes through and acknowledges all attempts that do not belong to them. There should be other API endpoints in this file which do something similar.

chrisndodge:
sorry, seems like we have some merge conflicts. I took a quick look and it doesn't look like something I can resolve on my own.

chrisndodge:
Thanks @hasnain-naveed looks good to me. Merging.

PR #246, opened 2015-12-16
set the left nav state when the attempt is in second_review_required

chrisndodge:
@afzaledx @muhhshoaib just one small change that I neglected to make on the original "second_review_required" work. The left nav bar needs to still act like it is in "submitted" state.


afzaledx:
:+1:

chrisndodge:
Wow, that was quick! Thanks @afzaledx

PR #247, opened 2015-12-17
PHX-225 / Admin Panel for Proctored Exam Attempt

hasnain-naveed:


edx_proctoring/admin.py
```diff
@@ -334,6 +343,255 @@ def save_model(self, request, review, form, change):
...
+    """
+
+    title = _('Course Id')
+    parameter_name = 'proctored_exam__course_id'
+
+    def lookups(self, request, model_admin):
+        """
+        List of values to allow admin to select
+        """
+        attempts = ProctoredExamStudentAttempt.objects.all()
```
hasnain-naveed: @chrisndodge 
Do we have to select all the attempts or should we filter it more?

edx_proctoring/admin.py
```diff
@@ -334,6 +343,255 @@ def save_model(self, request, review, form, change):
...
+
+        lookups = (())
+
+        unique_course_ids = []
+        for attempt in attempts:
+            course_id = attempt.proctored_exam.course_id.replace('+', ' ').replace('/', ' ').replace('course-v1:', '')
```
chrisndodge: I think since we have two places in this file where we do this 'prettifying' of the course_id, can you make a single utility method?

edx_proctoring/admin.py
```diff
@@ -334,6 +343,255 @@ def save_model(self, request, review, form, change):
...
+            if course_id not in unique_course_ids:
+                unique_course_ids.append(course_id)
+                lookups += ((
+                    attempt.proctored_exam.course_id,
+                    u'{course_id}: {exam_name}'.format(
+                        course_id=course_id,
+                        exam_name=attempt.proctored_exam.exam_name
```
chrisndodge: there can be multiple exams per course, so I think you just need to format course_id into the display string

edx_proctoring/admin.py
chrisndodge: Yea, this will likely be pretty slow approach. I think all you need to do here is 

ProctoredExam.objects.values('course_id').distinct()

Then that should give you a list of all course_ids that have exams

edx_proctoring/admin.py
```diff
@@ -334,6 +342,254 @@ def save_model(self, request, review, form, change):
...
+        Return the filtered queryset
+        """
+
+        if self.value() in [ProctoredExamStudentAttemptStatus.created, ProctoredExamStudentAttemptStatus.submitted]:
+            return queryset.filter(status=self.value())
+        else:
+            return queryset
+
+
+class ExamAttemptFilterIsProctored(admin.SimpleListFilter):
```
afzaledx: Can the BooleanFilter (http://django-filter.readthedocs.org/en/latest/ref/filters.html#booleanfilter) work for this one?

edx_proctoring/admin.py
```diff
@@ -334,6 +342,254 @@ def save_model(self, request, review, form, change):
...
+    ]
+
+    search_fields = [
+        'user__username',
+        'attempt_code'
+    ]
+
+    list_filter = [
+        ExamAttemptFilterByStatus,
+        ExamAttemptFilterIsProctored,
```
afzaledx: We can simply use the file name, and Django admin will detect that it's a boolean field and render the default boolean filter . Just add 'taking_as_proctored' and 'is_sample_attempt'. We can then remove these filter classes altogether.

afzaledx: For the "title" field, you can provide the verbose_name value to the model field and that will show up here. Make sure to i18n the verbose name.

edx_proctoring/admin.py
```diff
@@ -334,6 +342,183 @@ def save_model(self, request, review, form, change):
...
+        """ Return exam_name of attempt"""
+        return obj.proctored_exam.exam_name
+
+    def course_id(self, obj):
+        """ Return course_id of attempt"""
+        return obj.proctored_exam.course_id
+
+    def modified_timestamp(self, obj):
+        """ Return modified timestamp of attempt"""
+        return time.mktime(obj.modified.timetuple())
```
afzaledx: I guess the linux timestamp would not be useful here. You should probably use the obj.modified by itself to show a human readable date/time.

afzaledx:
:+1:

chrisndodge:
Thanks @hasnain-naveed, I'll pull this down and give it a try

chrisndodge:
Thanks. Looks good but I just added a few more filter values and columns in the list view.

PR #248, opened 2015-12-21
(WIP) PHX-231 / correction of quick filter by removing the whole class

hasnain-naveed:


afzaledx:
@chrisndodge, this change fixes a bug (PHX-231) which hasnain found while testing on the sandbox. Now, instead of all statuses showing up in the filters list, only those values of status will appear which exists in the database. Seems fine to me. The alternative would be to make all the possible values of status how up in the filter, which would require fixing the queryset method of the filter class (which has now been removed altogether).
:+1:

chrisndodge:
OK, this is fine.

PR #251, opened 2015-12-22
redact video url in the review postback

chrisndodge:
This corresponds to PHX-229. After this code is deployed, we can decide if we want to drop the column from the model in a subsequent release (and also remove any code references to the field).

@afzaledx @douglashall


chrisndodge:
@afzaledx @douglashall can we decide if we want to merge this in for the next release in January? If so, then I'd like to get it up on the sandbox for the manual regression test that will be done on Monday. Thanks.

douglashall:
@chrisndodge I think we need to run makemigrations on this.

chrisndodge:
Just spoke with Doug. This PR does not actually remove the video_url from the Django model

douglashall:
:+1:

PR #252, opened 2015-12-28
Rc/2016 01 06

chrisndodge:


chrisndodge:
@douglashall @afzaledx here is next week's RC branch for edx-proctoring. There appears to be a merge conflict (most certainly due to the version # changes in setup.py). The github tag (0.12.4) has already been created on the tip of this RC and the PR in edx-platform points to it.

In this case, due to the merge conflict, I'd recommend doing the deploy and then force pushing this RC branch to be 'release'. In that way 'release' will properly be pointing to what had been deployed.

PR #258, opened 2016-01-11
Hotfix PHX-242

douglashall:
Exam attempt status is not updated from "download_software_clicked" to "ready_to_start" when stepping through SoftwareSecure onboarding workflow.

### DIFF
--- a/edx_proctoring/callbacks.py
+++ b/edx_proctoring/callbacks.py
@@ -18,11 +18,11 @@ from rest_framework.negotiation import BaseContentNegotiation
 from edx_proctoring.api import (
     get_exam_attempt_by_code,
     mark_exam_attempt_as_ready,
-    update_exam_attempt)
-
-from edx_proctoring.exceptions import ProctoredBaseException
-
+    update_exam_attempt
+)
 from edx_proctoring.backends import get_backend_provider
+from edx_proctoring.exceptions import ProctoredBaseException
+from edx_proctoring.models import ProctoredExamStudentAttemptStatus
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


douglashall:
@mattdrayer @mjfrey Could I get a review of this?

douglashall:
Also, @afzaledx @ibrahimahmed443

edx_proctoring/callbacks.py
```diff
@@ -15,10 +15,12 @@
 from rest_framework.response import Response
 from rest_framework.negotiation import BaseContentNegotiation
 
+from edx_proctoring.models import ProctoredExamStudentAttemptStatus as ExamAttemptStatus
```
mattdrayer: Nit: relocate this import to below edx_proctoring.api (alphabetical).  

Also any particular reason you are importing the model with an alternate name?  I usually only do this when a conflict exists, even if the class/method name is long.

edx_proctoring/tests/test_views.py
```diff
@@ -556,6 +556,52 @@ def test_start_exam_callback(self):
         self.assertEqual(attempt['status'], "started")
         self.assertFalse(attempt['status'] == "ready_to_start")
 
+    def test_start_exam_callback_when_download_software_clicked(self):
+        """
+        Test that hitting software secure callback URL twice when the attempt state begins at
+        'download_software_clicked' does not change the state from 'started' back to 'ready to start'
+        """
+        # Create an exam.
+        proctored_exam = ProctoredExam.objects.create(
```
mattdrayer: How about pulling the exam creation logic out into a private helper that can be shared by this test and the one above?

edx_proctoring/callbacks.py
douglashall: Updated import ordering and aliasing.

mattdrayer:
@douglashall -- just a couple small structural suggestions -- overall LGTM -- :+1:

mjfrey:
Once pylint errors fixed.
:+1:

PR #259, opened 2016-01-12
PHX-242 Handled the status "download_software_clicked" for an attempt in the …

afzaledx:
…Practice Exams flow. Earlier it was not handled and was showing the exam content when the page was refreshed.


douglashall:
:+1:

PR #268, opened 2016-02-03
PHX-244 Increment package version

douglashall:
@afzaledx


douglashall:
Sorry, the Jira issue referenced here should have been SOL-1480

PR #270, opened 2016-02-09
fix the bug where the learner gets full time even after due date is p…

ibrahimahmed443:
…assed


ibrahimahmed443:
@douglashall @afzaledx

edx_proctoring/views.py
```diff
@@ -595,14 +595,19 @@ def post(self, request):
         exam_id = request.data.get('exam_id', None)
         attempt_proctored = request.data.get('attempt_proctored', 'false').lower() == 'true'
         try:
+            exam = get_exam_by_id(exam_id)
+            if exam['due_date'] and exam['due_date'] <= datetime.now(pytz.UTC):
+                err_msg = (
+                    'Attempted to access expired exam'
+                )
```
douglashall: Is this message displayed to the user? Do we need i18n?

douglashall:
:+1:

PR #274, opened 2016-02-22
Correct Django Requirement

feanil:
Django 1.9a is less than 1.9 and should not be installed as a requirement since edx-platform is running 1.8 and 1.9a has a security vulnerability in it.

@douglashall @mattdrayer @mjfrey


douglashall:
:+1:

PR #280, opened 2016-03-02
fix the bug where timed/proctored exams in Self-Paced Courses obey pr…

ibrahimahmed443:
…evious instructor-paced assignment deadlines

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1424,7 +1424,7 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
     attempt_status = attempt['status'] if attempt else None
     has_due_date = True if exam['due_date'] is not None else False
     if not attempt_status:
-        if has_due_date_passed(exam['due_date']):
+        if has_due_date_passed(exam['due_date']) and not _is_self_paced_course(user_id, course_id):
             student_view_template = 'timed_exam/expired.html'
         else:
             student_view_template = 'timed_exam/entrance.html'
@@ -1524,6 +1524,23 @@ def _calculate_allowed_mins(due_datetime, allowed_mins):
 
             actual_allowed_mins = int((due_datetime - current_datetime).seconds / 60)
     return actual_allowed_mins, is_exam_past_due_date
+
+
+def _is_self_paced_course(user_id, course_id):
+    """
+    Returns whether the course if self-paced or not
+    Args:
+        user_id:
+        course_id:
+    """
+    credit_service = get_runtime_service('credit')
+
+    credit_state = credit_service.get_credit_state(
+        user_id,
+        course_id,
+        return_is_self_paced_course=True
+    )
+    return credit_state.get('is_self_paced_course', False)
 
 
 def _get_proctored_exam_context(exam, attempt, course_id, is_practice_exam=False):
@@ -1672,7 +1689,7 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
         })
 
         # if exam due date has passed, then we can't take the exam
-        if has_due_date_passed(exam['due_date']):
+        if has_due_date_passed(exam['due_date']) and not _is_self_paced_course(user_id, course_id):
             student_view_template = 'proctored_exam/expired.html'
         elif not prerequisite_status['are_prerequisites_satisifed']:
             # do we have any declined prerequisites, if so, then we
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -893,7 +893,7 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, 
         credit_state = credit_service.get_credit_state(
             exam_attempt_obj.user_id,
             exam_attempt_obj.proctored_exam.course_id,
-            return_course_name=True
+            return_course_info=True
         )
 
         send_proctoring_attempt_status_email(
@@ -1538,7 +1538,7 @@ def _is_self_paced_course(user_id, course_id):
     credit_state = credit_service.get_credit_state(
         user_id,
         course_id,
-        return_is_self_paced_course=True
+        return_course_info=True
     )
     return credit_state.get('is_self_paced_course', False)
 
--- a/edx_proctoring/tests/test_services.py
+++ b/edx_proctoring/tests/test_services.py
@@ -31,7 +31,7 @@ class MockCreditService(object):
             'credit_requirement_status': []
         }
 
-    def get_credit_state(self, user_id, course_key, return_course_name=False,
+    def get_credit_state(self, user_id, course_key, return_course_name=False,  # pylint: disable=unused-argument
                          return_is_self_paced_course=False):  # pylint: disable=unused-argument
         """
         Mock implementation
--- a/edx_proctoring/tests/test_services.py
+++ b/edx_proctoring/tests/test_services.py
@@ -31,7 +31,8 @@ class MockCreditService(object):
             'credit_requirement_status': []
         }
 
-    def get_credit_state(self, user_id, course_key, return_course_name=False):  # pylint: disable=unused-argument
+    def get_credit_state(self, user_id, course_key, return_course_name=False,
+                         return_is_self_paced_course=False):  # pylint: disable=unused-argument
         """
         Mock implementation
         """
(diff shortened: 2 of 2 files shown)


ibrahimahmed443:
@douglashall @afzaledx. requires https://github.com/edx/edx-platform/pull/11727

schenedx:
@bderusha @cpennington Could you guys review this PR?
This is related to proctored/timed exam does not obey the self-paced course property
I don't know if I completely agree with this solution, so I hope you guys can have some input.

douglashall:
@schenedx Would you mind adding comments to this PR indicating what you don't agree with? Thanks.

edx_proctoring/api.py
```diff
@@ -1526,6 +1526,23 @@ def _calculate_allowed_mins(due_datetime, allowed_mins):
     return actual_allowed_mins, is_exam_past_due_date
 
 
+def _is_self_paced_course(user_id, course_id):
```
schenedx: This introduces the concept of self-paced course into proctoring module. The fix here is quite intrusive to me.
When we call the function `_get_timed_exam_view`, we did pass in a `context` object where contains the `due_date` from LMS. My question is why can't we leverage this context object instead?

schenedx:
@douglashall I explained my position. However, I don't know this area so I maybe short-sighted.

afzaledx:
We are listening to the course_published signal here:
https://github.com/edx/edx-platform/blob/afzaledx/WL-318_separating_course_details/cms/djangoapps/contentstore/signals.py#L26

We create the exam as well as update the exam every time the course is published, and after that we use our local object. The create_exam in the get_student_view() which calls the _get_timed_exam_view was just left there for an erroneous condition, and is not the actual place where the exams are created.

This thing should not arise in the first place if there is a course_publish event after setting the course to self-paced.

Anyway, we can call the update_exam() every time the get_student_view() is called, but that doesn't seem to make sense to me.

afzaledx:
If a course published signal is being emitted, then there might be an error in our flow which fails to update the timed exams and we can look into that...

edx_proctoring/api.py
douglashall: @schenedx But what if the due date sent in the `context` object has passed? We already were checking this due date with `has_due_date_passed(exam['due_date'])` at line 1427 of this file. Are you working under the assumption that setting the self-paced flag on the course will overwrite the `due` field on the `CourseModule`? I'm guessing that does not happen, otherwise this would already work as expected, no?

schenedx:
@afzaledx Thanks for point out the course publish listener. My comments:

1. I don't know how much you know about self-paced courses, but a course can switch back and forth from self-paced to instructor-paced(with due dates and run duration) anytime before the course start time, including when course is published and not yet started. Do we know if the publish listener you referred to would catch the due date change even when course is switched to self-paced in studio? I argue this needs to be confirmed.
2. My comment about the `context` object is not limited to the `create_exam` function. Like I mentioned about the quick switching of the self-paced course, only use the `context` object when creating the `exam` object in proctoring data store will not capture the changing due dates in time.

edx_proctoring/api.py
douglashall: @schenedx Ah, nevermind. Now I see that we were using the `due_date` from the exam object and not from the context. So based on your knowledge of self-paced courses, will the `due_date` field passed in the context object always be None for a self-paced course, regardless of what someone has set on the subsection in Studio?

schenedx: @douglashall, that's what I have discovered. The `due_date` field passed within `context` is a field overriden by self-paced field override at https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/self_paced_overrides.py#L10 
The override gets applied at the module_render time at https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/module_render.py#L573
As you can see in the `self_paced_overrides`, the `due` and `start` date of all course elements are overriden. The reason for this override is precisely because we want to allow user to switch back and forth from self-paced to instructor-paced quickly.

douglashall: @schenedx Thanks for the info.

@afzaledx @ibrahimahmed443 Can you figure out if there is something going wrong with processing the course publish event when switching between self-paced/instructor-paced? If so, let's fix it. If we can't rely on the due_date on the exam object we should be able to use the due_date from the context instead.

afzaledx:
After some debugging, here's what I found. When we switch the course from instructor-paced to self-paced, and vice versa, a published event does get triggered each time, but the "due" always contains the actual due date, and never gets a value of None. This gets saved to the exam object and we use that object then on. This leaves us with the option to update the exam object every time the values passed in the context are different from the values stored in the exam object.

douglashall:
@afzaledx Can we dig a little deeper and see why the data you receive with the publish signal does not contain a None due value? Perhaps there is something we can fix there.

schenedx:
@afzaledx This is valuable information. I would like to point you to my other piece of research/discovery.
See ECOM-3733 comments section. Basically, we in the e-commerce team realized recently that self-pace `due date` override does not get applied in all situations. Can you please read into what I found and then figure out where the `due date` passed into the course `publish` event comes from? I have a suspicion that it could be related to what I found in ECOM-3733

afzaledx:
That's probably because the self-paced override has been applied on the LMS side, and we're getting this published signal from the CMS side. just a hunch. only guessing.

douglashall:
Ah, so we are not actually making updates to the dates when a course is made self-paced. It is for display purposes only. @schenedx Do I have that right?

afzaledx:
Yes. this is the same issue as ECOM-3733. We're retrieving the due-date directly from the module_store as well.

schenedx:
@douglashall Not quite.
The issue is that, self-paced overrides are only applied at a certain point during the data flow from data store to the user view. The certain point is when user context (who the learner is, what the learner settings are) is available. The override is applied when the course data is bound with the user context.
Unfortunately, not every piece of data shown to the user needs or will go through the user context binding. The research I did was on discussion module, but looks like @afzaledx also believes this applies to timed exams as well.

schenedx:
@afzaledx Can you show me the line of code where exam due date is retrieved directly from the module_store during `publish`?

afzaledx:
https://github.com/edx/edx-platform/blob/master/cms/djangoapps/contentstore/proctoring.py#L42-L83

afzaledx:
Pointing out again that this above ^^ is in the CMS while the get_student_view() and the _get_timed_exam_view() are called from the LMS, and thus the context contains the correct due date.

edx_proctoring/api.py
```diff
@@ -1424,7 +1424,7 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
     attempt_status = attempt['status'] if attempt else None
     has_due_date = True if exam['due_date'] is not None else False
     if not attempt_status:
-        if has_due_date_passed(exam['due_date']):
+        if has_due_date_passed(exam['due_date']) and not _is_self_paced_course(user_id, course_id):
```
douglashall: @afzaledx I think the approach we should take here is to use the due_date from the context object instead.
```
if has_due_date_passed(context['due_date']):
```
We will also want to replace all references to `exam['due_date']` with `context['due_date']` in this function.

douglashall: I suppose we could also update the exam due_date here as well if it has changed as you suggested this morning. What do you think?

afzaledx: I was thinking of a leaner approach, where we just override 
`exam['due_date'] = context['due_date']` and it would fix all our woes.

douglashall: Sounds good. Any concerns about not updating the exam record in the database?

schenedx: My question to this approach is: in the proctoring service, do we want to always persist the `due date` information passed in from the `context` object? Basically, the true source of `due date` information is always outside of the proctoring service? 
I think I am definitely in support of the above idea because I do believe the exam information, since it's part of the course anyways, should be passed into the proctoring service as the trusted data.

afzaledx: Yup. The due date is used a lot, so we definitely need to updated it in our object.

ibrahimahmed443:
closing this PR in favor of https://github.com/edx/edx-proctoring/pull/282

PR #281, opened 2016-03-02
Pass requesting user on delete_student_attempt

efischer19:
The LMS implementation that will be performing this work is being changed
to require requesting user information as part of a seperate bugfix.

See TNL-4152, ARCH-18 for details. An updated version of edx-proctoring
containing this change will need to be merged into edx-platform's
github.txt

### DIFF
--- a/edx_proctoring/backends/tests/test_software_secure.py
+++ b/edx_proctoring/backends/tests/test_software_secure.py
@@ -649,7 +649,7 @@ class SoftwareSecureTests(TestCase):
         )
 
         # now delete the attempt, which puts it into the archive table
-        remove_exam_attempt(attempt_id)
+        remove_exam_attempt(attempt_id, requesting_user=self.user)
 
         # now process the report
         provider.on_review_callback(json.loads(test_payload))
@@ -873,7 +873,7 @@ class SoftwareSecureTests(TestCase):
         self.assertEqual(attempt['status'], attempt['status'])
 
         # now delete the attempt, which puts it into the archive table
-        remove_exam_attempt(attempt_id)
+        remove_exam_attempt(attempt_id, requesting_user=self.user)
 
         review = ProctoredExamSoftwareSecureReview.objects.get(attempt_code=attempt['attempt_code'])
 
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -962,9 +962,9 @@ def send_proctoring_attempt_status_email(exam_attempt_obj, course_name):
     email.send()
 
 
-def remove_exam_attempt(attempt_id):
+def remove_exam_attempt(attempt_id, requesting_user):
     """
-    Removes an exam attempt given the attempt id.
+    Removes an exam attempt given the attempt id. requesting_user is passed through to the instructor_service.
     """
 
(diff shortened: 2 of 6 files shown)


efischer19:
FYI @douglashall - this isn't ready for a full review yet, but I want to make sure you're aware of it since it's being made in a solutions repo.

I need to make some changes to the `reset_student_attempts` method in the LMS on edx-platform. These changes will require `requesting_user` information to be passed in to this method: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/instructor/enrollment.py#L205.

Since proctoring uses the InstructorService (which calls that method), I've attempted to track that codepath back to the point where that information is available so that it may be passed through. See https://openedx.atlassian.net/browse/ARCH-18 for further details/background, and let me know if there are any questions.

Also FYI @robrap - I think these are the only changes we'll need to make outside of the edx-platform repo. We'll need to coordinate with solutions to make sure a new version of edx-proctoring is installed into edx-platform at the same time our other changes are merged.

douglashall:
@efischer19 Thanks for the heads up. Feel free to ping me when you need a review.

efischer19:
@douglashall I have all my changes up and installed on https://ora2clearstate.sandbox.edx.org/, would you mind giving this a look when you have a chance? https://openedx.atlassian.net/browse/ARCH-18 has all the details of what we're trying to do here, and https://github.com/edx/edx-platform/pull/11666 is the corresponding edx-platform PR.

@robrap to be second reviewer.

douglashall:
:+1:  Can you update the setup.py version and tag with the new version?

efischer19:
Thanks @douglashall, will do. Is there anything else that needs to be done for an edx-proctoring release? If not, I'll just update the version to `0.12.13` and merge that with my edx-platform PR for next-next week'e release (the RC to be cut next Friday, deployed the week after)

douglashall:
No, that is all that is required. Sounds like a plan. Thanks for making this update.

edx_proctoring/backends/tests/test_software_secure.py
```diff
@@ -649,7 +649,7 @@ def test_review_on_archived_attempt(self):
         )
 
         # now delete the attempt, which puts it into the archive table
-        remove_exam_attempt(attempt_id)
+        remove_exam_attempt(attempt_id, "null_user")
```
robrap: As discussed, you can add an assertion that what is passed is actually a user to show how this should be used.  If you could name the variable ``requested_user``, I think that would make this more clear.

robrap:
:+1: Thanks.

efischer19:
@douglashall, it looks like there have been some commits to master recently (https://github.com/edx/edx-proctoring/pull/282 and https://github.com/edx/edx-proctoring/pull/283) - is it alright if they're included in a new tagged version along with this code?

efischer19:
Apologies @douglashall - just noticed that you put out a 0.12.13 release as well. I'll bump to 0.12.14 with mine and include that in my platform PR later today.

douglashall:
Thanks @efischer19. Sorry I had forgotten about this PR when I created the new version to roll out some recent bug fixes.

PR #282, opened 2016-03-07
Afzaledx/sol 1653 course due date for self paced

ibrahimahmed443:

PR #283, opened 2016-03-08
let the student take exam without proctoring if the  exam has no due …

ibrahimahmed443:
…date and course end date is passed

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -893,7 +893,7 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, 
         credit_state = credit_service.get_credit_state(
             exam_attempt_obj.user_id,
             exam_attempt_obj.proctored_exam.course_id,
-            return_course_name=True
+            return_course_info=True
         )
 
         send_proctoring_attempt_status_email(
@@ -1770,6 +1770,12 @@ def get_student_view(user_id, course_id, content_id,
     if user_role != 'student':
         return None
 
+    credit_service = get_runtime_service('credit')
+
+    # call service to get course end date.
+    credit_state = credit_service.get_credit_state(user_id, course_id, return_course_info=True)
+    course_end_date = credit_state.get('course_end_date', None)
+
     exam_id = None
     try:
         exam = get_exam_by_content_id(course_id, content_id)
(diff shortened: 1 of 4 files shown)


ibrahimahmed443:
@douglashall @afzaledx. This uses: https://github.com/edx/edx-platform/pull/11767

edx_proctoring/api.py
```diff
@@ -1770,6 +1770,20 @@ def get_student_view(user_id, course_id, content_id,
...
         return None
 
+    credit_service = get_runtime_service('credit')
+
+    # call service to get course end date.
+    credit_state = credit_service.get_credit_state(
+        user_id,
+        course_id,
+        return_course_info=True
+    )
```
douglashall: Nit: Put this on a single line.

edx_proctoring/api.py
```diff
@@ -1798,11 +1812,15 @@ def get_student_view(user_id, course_id, content_id,
...
+        sub_view_func = _get_practice_exam_view
     elif is_proctored_exam:
-        return _get_proctored_exam_view(exam, context, exam_id, user_id, course_id)
+        sub_view_func = _get_proctored_exam_view
 
-    return None
+    if sub_view_func:
+        return sub_view_func(exam, context, exam_id, user_id, course_id)
+    else:
+        return None
```
douglashall: Nice cleanup!

edx_proctoring/api.py
```diff
@@ -1770,6 +1770,20 @@ def get_student_view(user_id, course_id, content_id,
...
+    # call service to get course end date.
+    credit_state = credit_service.get_credit_state(
+        user_id,
+        course_id,
+        return_course_info=True
+    )
+
+    # if exam has no due date and course end date is passed, we'd still let the student take exam but without proctoring
+    end_date = credit_state.get('course_end_date', None)
+    if context.get('due_date') is None and has_due_date_passed(end_date):
```
douglashall: I think we need to add a test for this condition.

edx_proctoring/api.py
```diff
@@ -1770,6 +1770,16 @@ def get_student_view(user_id, course_id, content_id,
...
 
+    credit_service = get_runtime_service('credit')
+
+    # call service to get course end date.
+    credit_state = credit_service.get_credit_state(user_id, course_id, return_course_info=True)
+
+    # if exam has no due date and course end date is passed, we'd still let the student take exam but without proctoring
+    end_date = credit_state.get('course_end_date', None)
+    if context.get('due_date') is None and has_due_date_passed(end_date):
+        return None
```
douglashall: Thinking about this more. I think we want to still allow the exam to be timed if the course end date has passed. We only want to skip proctoring if the course end data has passed. Can we move this check down below the exam lookup and only return None if the course end date has passed and it is a proctored or practice exam? Also, I don't think we need to check the exam due date here. If the course end date has passed that is enough to skip proctoring.

douglashall:
It seems like we lost the previous fix for https://github.com/edx/edx-proctoring/pull/282/files#diff-c80ba566532963b4396830d546129f80R1782. I'm not seeing it removed in the diff and there are no conflicts. Can you rebase this branch?

douglashall:
:+1:

PR #286, opened 2016-03-15
Fix timer time difference

mushtaqak:
On Timed exams, there is issue with total time to attempt if due date is 24+ hours from now and we have set 24+ hours time_limit_mins ( 27 hours ) i.e it is like 1 day and 3 hours total time left to attempt the exam but it shows only 3 hours remaining time.

**Issue**
Actually `.seconds` is used instead of `.toatal_seconds()` to calculate total time between the dates/time interval.

[TNL-4216](https://openedx.atlassian.net/browse/TNL-4216)

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -29,6 +29,7 @@ from edx_proctoring.api import (
     create_exam_attempt,
     get_exam_attempt_by_id,
     update_attempt_status,
+    _calculate_allowed_mins
 )
 
 from .utils import (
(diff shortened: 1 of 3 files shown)


mushtaqak:
@douglashall @afzaledx @hasnain-naveed kindly review.

afzaledx:
:+1: The .seconds is not valid for durations >24 hours. timedelta.total_seconds() seems to be correct.

From Documentation: timedelta.total_seconds()
Return the total number of seconds contained in the duration. Equivalent to (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / 10**6 computed with true division enabled.

mushtaqak:
@afzaledx Yeah, precisely.

@douglashall Can I have your review as well. Thanks.

edx_proctoring/tests/test_views.py
```diff
@@ -674,11 +675,33 @@ def test_timer_remaining_time(self):
...
+        expected_total_minutes = 27 * 60
+        proctored_exam = ProctoredExam.objects.create(
+            course_id='a/b/c',
+            content_id='test_content',
+            exam_name='Test Exam',
+            external_id='123aXqe3',
+            time_limit_mins=1800,
+            due_date=datetime.now(pytz.UTC) + timedelta(minutes=expected_total_minutes),
+        )
+        total_minutes, _ = _calculate_allowed_mins(proctored_exam.due_date, proctored_exam.time_limit_mins)
```
douglashall: Prefer `__` instead of `_` for unused variables.

douglashall:
Minor comment, otherwise :+1: .

edx_proctoring/tests/test_views.py
mushtaqak: Great, didn't know about this before. I will address this comment and will merge when tests pass :)

PR #287, opened 2016-03-16
fix the bug where 'Proctored Option available' appears on the left pa…

ibrahimahmed443:
…ne for a proctored exam in an expired instructor paced course

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1309,6 +1309,11 @@ STATUS_SUMMARY_MAP = {
         'short_description': _('Failed Proctoring'),
         'suggested_icon': 'fa-exclamation-triangle',
         'in_completed_state': True
+    },
+    ProctoredExamStudentAttemptStatus.expired: {
+        'short_description': _('Proctored Option No Longer Available'),
+        'suggested_icon': 'fa-times-circle',
+        'in_completed_state': False
     }
 }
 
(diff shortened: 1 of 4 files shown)


ibrahimahmed443:
@douglashall @afzaledx

edx_proctoring/api.py
```diff
@@ -1309,6 +1309,11 @@ def _resolve_prerequisite_links(exam, prerequisites):
         'short_description': _('Failed Proctoring'),
         'suggested_icon': 'fa-exclamation-triangle',
         'in_completed_state': True
+    },
+    ProctoredExamStudentAttemptStatus.expired: {
+        'short_description': _('Proctored Option Not Available'),
```
douglashall: Could we change this to 'Proctored Option No Longer Available'?

edx_proctoring/api.py
```diff
@@ -1375,12 +1380,17 @@ def get_attempt_status_summary(user_id, course_id, content_id):
     # practice exams always has an attempt status regardless of
     # eligibility
     if credit_service and not exam['is_practice_exam']:
-        credit_state = credit_service.get_credit_state(user_id, unicode(course_id))
+        credit_state = credit_service.get_credit_state(user_id, unicode(course_id), return_course_info=True)
         if not _check_eligibility_of_enrollment_mode(credit_state):
```
douglashall: Can we add the `has_due_date_passed` check on the course_end_date here and return None if it has passed?

edx_proctoring/api.py
douglashall: Actually, I think it would be better to display no message in the summary at all. See comment below.

edx_proctoring/api.py
```diff
@@ -1375,12 +1380,17 @@ def get_attempt_status_summary(user_id, course_id, content_id):
...
             return None
 
     attempt = get_exam_attempt(exam['id'], user_id)
-    status = attempt['status'] if attempt else ProctoredExamStudentAttemptStatus.eligible
+    if attempt:
+        status = attempt['status']
+    elif not exam['is_practice_exam'] and has_due_date_passed(credit_state.get('course_end_date', None)):
```
douglashall: I think we need a unit test to cover this case.

douglashall: This case is not covered by the ddt addition below. I think we need to add a test for when there is no attempt, it is not a practice exam, and the course_end_date has passed.

douglashall:
:+1: Please squash commits.

PR #289, opened 2016-05-03
Increased in-exam polling interval from 5 seconds to 30 seconds.

afzaledx:


afzaledx:
@douglashall, kindly take a look. I have created a new tag 0.12.17 as well. The corresponding branch in edx-platform is : https://github.com/edx/edx-platform/pull/12343

douglashall:
👍

douglashall:
@ormsbee Can you give this a second review? We are looking to include this in a patch release going out to production tomorrow along with some other changes we need to push for a WL site relaunch.

ormsbee:
👍 on this, since it improves the situation and lets us gather more data. But I don't think this fully resolves SOL-1780 as we're still making potentially millions of requests for hundreds of students.

ormsbee:
Wait, doesn't setup.py need to get its version bumped in this commit as well?

douglashall:
Good catch, thanks. I will take care of it.

PR #294, opened 2016-06-20
Allow masquerading instructors to see content

efischer19:
## [TNL-4808](https://openedx.atlassian.net/browse/TNL-4808)

First take on this is:
- calculate what would be returned to the student
- if something would be returned (hides the content) and a staff user is masquerading, return the string `masquerade_hidden`

There's probably a cleaner way to return the value, it's a proof-of-concept for now.

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1843,6 +1843,9 @@ def get_student_view(user_id, course_id, content_id,
         sub_view_func = _get_proctored_exam_view
 
     if sub_view_func:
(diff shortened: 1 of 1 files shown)


efischer19:
As of the latest commit on https://github.com/edx/edx-platform/pull/12806, I've made it possible to override while masquerading without touching `edx-proctoring` code at all. Closing this PR unless that changes, as I'd really like to avoid coupling the idea of masquerading to this API.

efischer19:
@douglashall FYI, regarding our earlier discussion ^

PR #295, opened 2016-06-21
SOL-1844 Allow for distinguishing between proctored exam and timed exam requests

douglashall:
Adding a request parameter to exam attempt status polling that takes place during an exam which will allow us to distinguish between timed and proctored exams in logs.

Related PR: https://github.com/edx/edx-platform/pull/12815

@ormsbee @efischer19

### DIFF
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
@@ -137,8 +137,9 @@ var edx = edx || {};
             self.timerTick ++;
             self.secondsLeft --;
             if (self.timerTick % 30 === 0){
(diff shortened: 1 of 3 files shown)


ormsbee:
👍 

FWIW (and this is totally optional), if you're interested in querying from NR Insights, you can add the querystring params as custom parameters. Bright side is that querying Insights is really, really fast: https://openedx.atlassian.net/wiki/display/PERF/Custom+Metrics+in+New+Relic#CustomMetricsinNewRelic-AddYourOwn

efischer19:
Seems simple enough to me, although it looks like the test failure might be related to this work. Provisional 👍 assuming that gets fixed.

douglashall:
@ormsbee In the interest of time, I'm going to hold off on adding the NR tooling. I'm looking to get this change out to prod in a patch release tomorrow.

PR #298, opened 2016-07-29
Add a proctoring test course to use for sandboxes, automated tests etc

andy-armstrong:
This PR just adds my test proctoring course that I've used on my sandbox as well as for local testing. Eventually this could be the course that is loaded automatically onto a proctoring sandbox.

@dianakhuang @benpatterson please review.

@benpatterson @jibsheet Let me know if this would be better stored in edx-platform for the sake of automation.


benpatterson:
@andy-armstrong Although I realize it results in some sprawl, the easier path would be to create your own repo with this course as the only contents.

When sandboxes are set up...under the covers, they use the [git_add_course](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/dashboard/management/commands/git_add_course.py) management command to import the course. That command will clone/pull a git repo into the system and then import it into the modulestore. 

If you want to be able to check in a tar to a repo, like how we do in ora2, we'd ultimately need to update that command to handle sub directories of a given repo. (Which is a good idea, but will take an investment of yours/someone-else's time.)

Unless that idea is palatable to you, I suggest you follow the patterns from the other automatic-test-courses, like the [accessibility demo course](https://github.com/edx/demo-test-course) or the [performance test course](https://github.com/edx/demo-performance-course).

Hope this helps.

andy-armstrong:
Thanks @benpatterson. This is really great advice. I'm now wondering whether it would make more sense to add a proctored exam to the accessibility demo course, rather than having a new demo course repo per feature. What do you think?

andy-armstrong:
Closing this PR since adding it to this repo isn't going to be the right solution.

PR #299, opened 2016-08-03
User navigated to the progress page.

attiyaIshaque:
### Description
 
[TNL-5160](https://openedx.atlassian.net/browse/TNL-5160)

Progress page link on You have submitted your timed exam, would navigate User to the progress page.

Also having the Fix for:  [TNL-5180](https://openedx.atlassian.net/browse/TNL-5180)

### Sandbox
[http://progress-ecom-5132.sandbox.edx.org/](progress.sandbox.edx.org)

### Testing

- Navigate to [https://progress-ecom-5132.sandbox.edx.org/courses/course-v1:Edx+101+2016/courseware/a9e66cf978ff4ce39cdfaee927eff8a1/376f0040a5ef4314ac288f9259ee23de/](https://progress-ecom-5132.sandbox.edx.org/courses/course-v1:Edx+101+2016/courseware/a9e66cf978ff4ce39cdfaee927eff8a1/376f0040a5ef4314ac288f9259ee23de/)            Make sure you are logged in as a student 
- Complete the exam and End the timed exam 
- Click the Progress page link 
- User should be navigated to Progress page

### Reviewers
- [x] @adampalay 
- [x] @mushtaqak

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -923,12 +923,12 @@ def send_proctoring_attempt_status_email(exam_attempt_obj, course_name):
     course_info_url = ''
     email_template = loader.get_template('emails/proctoring_attempt_status_email.html')
     try:
-        course_info_url = reverse('courseware.views.course_info', args=[exam_attempt_obj.proctored_exam.course_id])
+        course_info_url = reverse(
+            'courseware.views.views.course_info',
+            args=[exam_attempt_obj.proctored_exam.course_id]
+        )
     except NoReverseMatch:
-        # we are allowing a failure here since we can't guarantee
-        # that we are running in-proc with the edx-platform LMS
-        # (for example unit tests)
-        pass
+        log.exception("Can't find Course Info url for course %s", exam_attempt_obj.proctored_exam.course_id)
 
     scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
     course_url = '{scheme}://{site_name}{course_info_url}'.format(
@@ -1258,14 +1258,11 @@ def _resolve_prerequisite_links(exam, prerequisites):
         if prerequisite['namespace'] in JUMPTO_SUPPORTED_NAMESPACES and prerequisite['name']:
             try:
                 jumpto_url = reverse(
-                    'courseware.views.jump_to',
+                    'courseware.views.views.jump_to',
                     args=[exam['course_id'], prerequisite['name']]
                 )
             except NoReverseMatch:
-                # we are allowing a failure here since we can't guarantee
-                # that we are running in-proc with the edx-platform LMS
-                # (for example unit tests)
-                pass
+                log.exception("Can't find jumpto url for course %s", exam['course_id'])
 
         prerequisite['jumpto_url'] = jumpto_url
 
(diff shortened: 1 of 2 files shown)


mushtaqak:
@attiyaIshaque we would probably need a simple unit test to check the progress page url works in this case, checking the response status to 200 would be fine.

edx_proctoring/api.py
```diff
@@ -1497,14 +1497,11 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
         progress_page_url = ''
         try:
             progress_page_url = reverse(
-                'courseware.views.progress',
+                'courseware.views.views.progress',
                 args=[course_id]
             )
         except NoReverseMatch:
```
mushtaqak: Do we need to have except here as well. 

I think, this is called by [LMS progress view](https://github.com/edx/edx-platform/blob/5346ee25b8eed7d13eec9dc451f22e442909c9b3/lms/djangoapps/courseware/module_render.py#L272) which is passing valid course id, so if the view is passed correctly in the reverse method, I don't think, we need exception block here.

@hasnain-naveed  was there any specific reason this was added ?

attiyaIshaque: @mushtaqak Yeah, you are right. Thanks for pointing this. :)

attiyaIshaque:
@mushtaqak, can't think of a test on the proctoring app's side that would catch this. So might be it would be on LMS side.

mushtaqak:
OK - Looks good to me otherwise 👍 

Thanks for creating [TNL-5180](https://openedx.atlassian.net/browse/TNL-5180) to change the outdated code.

edx_proctoring/api.py
adampalay: The exception block is for running this in an environment where that view isn't defined (like unit tests)

adampalay:
@attiyaIshaque , can just make the fix for TNL-5180 on this PR? There are only like two other reverses that need to be changed.

We should also give the TNL team a heads up that that change affected apps not in edx-platform: @efischer19 and @jcdyer

attiyaIshaque:
@adampalay I think we should make the fix for [TNL-5180](https://openedx.atlassian.net/browse/TNL-5180) in this PR and I will.

attiyaIshaque:
@adampalay @mushtaqak PR Updated.

edx_proctoring/api.py
```diff
@@ -1555,14 +1547,11 @@ def _get_proctored_exam_context(exam, attempt, course_id, is_practice_exam=False
...
-            'courseware.views.progress',
+            'courseware.views.views.progress',
             args=[course_id]
         )
     except NoReverseMatch:
-        # we are allowing a failure here since we can't guarantee
-        # that we are running in-proc with the edx-platform LMS
-        # (for example unit tests)
-        pass
+        log.exception('cant find progress url for course %s', course_id)
```
adampalay: `cant` should be `Can't` (in all these examples)

edx_proctoring/api.py
```diff
@@ -923,12 +923,10 @@ def send_proctoring_attempt_status_email(exam_attempt_obj, course_name):
     course_info_url = ''
     email_template = loader.get_template('emails/proctoring_attempt_status_email.html')
     try:
-        course_info_url = reverse('courseware.views.course_info', args=[exam_attempt_obj.proctored_exam.course_id])
+        course_info_url = reverse('courseware.views.views.course_info',
```
adampalay: nit: style for breaking long lines: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#breaking-long-lines

adampalay:
@attiyaIshaque couple nits, otherwise 👍  :)

mushtaqak:
👍 from me

attiyaIshaque:
@andy-armstrong  and @dianakhuang Please make it into release as soon as possible. Thanks

FYI: @adampalay

dianakhuang:
Thanks @attiyaIshaque ! I don't foresee any problems getting it into the release as soon as our other proctoring work is done.

andy-armstrong:
@attiyaIshaque This change will be going out in next week's release, so be sure to verify it on stage. Thanks!

attiyaIshaque:
@andy-armstrong Thanks :)

PR #300, opened 2016-08-08
Modify Proctoring Test Instructions Behavior

dianakhuang:
### Description
 
[TNL-5087](https://openedx.atlassian.net/browse/TNL-5087)

I have modified the behavior of this page such that it no longer polls to see whether or not the learner has completed the proctoring setup.

The color updates to the instructions will need to be a different edx-platform change: https://github.com/edx/edx-platform/pull/13187/commits/d7a2aa611fc1edc2f09bb319f0d03f9e9984bfaf

Here's some screenshots of the prototype:
New instructions
Error modal

### Sandbox
1. Visit https://dianakhuang.sandbox.edx.org/courses/course-v1:Proctoring+101+Now/courseware/e38af50b24a74f079af0c202bd364baa/8ce10919375843529a68d88e56860e9a/
2. Log in with the verified@example.com user information.
3. At the bottom of the instructions, click the 'Start Proctored Exam' button to view an error modal telling you need to set up proctoring.

### Reviewers
- [x] Code review: @andy-armstrong 
- [x] Code review: @bjacobel 
- [x] Doc Review: @catong 
- [ ] UX review: @chris-mike 
- [x] Product Review: @marcotuts 

### Post-review
- [ ] Squash commits

### DIFF
--- a/edx_proctoring/templates/proctored_exam/instructions.html
+++ b/edx_proctoring/templates/proctored_exam/instructions.html
@@ -38,8 +38,44 @@
         You will be asked to verify your identity before you begin the exam. Make sure you have valid photo identification, such as a driver's license or passport, before you continue.
       {% endblocktrans %}
     </p>
+    <p>
+        {% blocktrans %}
+          3. When you have finished setting up proctoring, start the exam.
+        {% endblocktrans %}
+    </p>
+    <div>
+        <button type="button" class="exam-action-button btn-pl-primary start-proctored-exam">
+            {% trans "Start Proctored Exam" %}
+        </button>
+    </div>
   </div>
 </div>
+<div id="accessible-error-modal" class="modal" aria-hidden="true">
+  <div class="inner-wrapper" role="dialog" aria-labelledby="accessible-error-title">
+    <button class="close-modal">
+      <span class="icon fa fa-remove" aria-hidden="true"></span>
+      <span class="sr">
+          {% trans "Close" %}
+      </span>
+    </button>
+
+    <header>
+        <h2 id="acessible-error-title">{% trans "Cannot Start Proctored Exam" %}</h2>
+        <span class="sr">,
+            {% trans "modal open"}
+        </span>
+    <hr aria-hidden="true" />
+    </header>
+    <div role="alertdialog" class="status message" tabindex="-1">
+        <p class="message-title" style="text-align: center; font-size: 16px;"></p>
+    </div>
+    <div class="actions" style="text-align: center;">
+        <button class="dismiss ok-button"></button>
+    </div>
+  </div>
+  <a href="#accessible-error-modal" rel="leanModal" id="confirm_open_button" style="display:none">{% trans "open"}</a>
+</div>
+
 <div class="footer-sequence border-b-0 padding-b-0">
   {% if not is_sample_attempt %}
   <p class="proctored-exam-instruction">
@@ -53,7 +89,14 @@
 
 <script type="text/javascript">
 
-  var _waiting_for_proctored_interval = null;
+  var accessible_error = function(message) {
+
+      accessible_modal("#accessible-error-modal #confirm_open_button",
+              "#accessible-error-modal .close-modal", "#accessible-error-modal", ".content-wrapper");
+      $("#accessible-error-modal #confirm_open_button").click();
+      $("#accessible-error-modal .message-title").html(message);
+      $("#accessible-error-modal .ok-button").html(gettext("OK"));
+  };
 
   $('.proctored-decline-exam').click(
     function(e) {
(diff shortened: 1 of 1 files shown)


chris-mike:
I have a few straightforward suggestions for changes.
1. In first screenshot, contrast between font and background is not high enough.  Suggest making dark grey text (in whole box, not just red-region) this color:
Sass: palette(grayscale, base) RGB: rgb(65, 65, 65) HEX: 414141

Optional: You could also make the background a pattern library color. The closest is Sass: palette(primary, x-back) RGB: rgb(242, 248, 251) HEX: f2f8fb.

2.  Since this is student facing, you could also change the button design. However, I'm not sure if it can be "default" or "brand" styles (http://ux.edx.org/components/buttons/).  At this point, it looks like documentation (iow, the button looks inactive).  In addition, I'd look to @catong  to see if this line needs to be numbered as "step 3."  

3. The font in the red-highlighted box looks larger than the rest of it.  Can we make it consistent with all other text on the page?


For the modal:
1. I think buttons could also be updated - default style is probably OK here.
2. You could center-align the text here ("You have not completed the set up")
3. There is also a weird line (white / light grey) under the "We're Sorry".  
4. Also, we're trying to get away from use of all caps (driven by accessibility). This is not a critical case, but it could be less of a scream.

catong:
@dianakhuang I second all @chris-mike 's recommendations, though any color and contrast changes should be made to all of the various proctored exam pages, not just this one. 
I do have suggestions for text, which I'll add as comments inline.

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -87,27 +136,26 @@
...
-        }
         // we've state transitioned, so refresh the page
         // to reflect the new state (which will expose the test)
         location.reload();
       }
+      else {
+          debugger;
+          // The proctoring setup is not yet complete.
+          // Show a modal indicating that the user is not done yet.
+          accessible_error("You have not yet completed the proctoring setup.");
```
catong: Can we change the message to: "You must complete the proctoring setup before you can start the exam."

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -38,8 +38,44 @@
         You will be asked to verify your identity before you begin the exam. Make sure you have valid photo identification, such as a driver's license or passport, before you continue.
       {% endblocktrans %}
     </p>
+    <p>
+        {% blocktrans %}
+          When you have finished setting up the proctoring application, you may start the exam.
```
catong: Can we make this step 3 as Chris suggests, and change the text to:
"3. When you have finished setting up proctoring, start the exam."

catong: @dianakhuang There's a lot more space below the "Start Proctored Exam" button than above it. Not critical, obviously, but adding a bit more space above the button and reducing the space below might improve appearances.

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -53,7 +89,20 @@
 
 <script type="text/javascript">
 
-  var _waiting_for_proctored_interval = null;
+  var accessible_error = function(message, callback) {
+      $("#accessibile-error-modal .cancel-button").click(function(){
```
andy-armstrong: Adding the click handlers each time the error is shown could cause trouble, because the code is reusing the same element. When the error is shown a second time, there will be two click handlers active for each button, and they might interfere with each other. I would recommend moving this code out of the function so that the handlers are always in place. They won't cause any trouble if the modal hasn't been shown.

Another option would be to clean up the handlers when the modal is hidden, but that is probably too much work.

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -53,7 +89,20 @@
...
+          $("#accessibile-error-modal .close-modal").click();
+      });
+      $("#accessibile-error-modal .ok-button").click(function(){
+          $("#accessibile-error-modal .close-modal").click();
+          callback();
+      });
+
+      accessible_modal("#accessibile-error-modal #confirm_open_button", "#accessibile-error-modal .close-modal", "#accessibile-error-modal", ".content-wrapper");
```
andy-armstrong: Wrap this line as it is too long as it stands.

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -53,7 +89,20 @@
...
+      $("#accessibile-error-modal #confirm_open_button").click();
```
andy-armstrong: What is this click for? It is somewhat confusing to simulate a click to force the modal to open. Is this how other modal code is working?

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -87,27 +136,25 @@
...
   );
 
-  $(document).ready(function(){
-    _waiting_for_proctored_interval = setInterval(
-      poll_exam_started,
-      5000
-    );
-  });
 
   function poll_exam_started() {
```
andy-armstrong: I'd rename this as it is no longer polling.

andy-armstrong:
@dianakhuang This looks great and will vastly improve the performance. I have a few comments so let me know when you're ready for me to re-review.

edx_proctoring/templates/proctored_exam/instructions.html
dianakhuang: It is there to simulate a click for the modal code to work. I think it's a weird thing that these modals do specifically.

dianakhuang:
@andy-armstrong I believe I have addressed your feedback. Still building a new sandbox right now,  but you can take a look at the new code.

andy-armstrong:
@dianakhuang Looks great, thanks. 👍 

BTW, is there somewhere to put an integration test for this new behavior? Does that have to go into edx-platform?

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -53,7 +89,14 @@
 
 <script type="text/javascript">
 
-  var _waiting_for_proctored_interval = null;
+  var accessible_error = function(message) {
+
+      accessible_modal("#accessibile-error-modal #confirm_open_button",
+              "#accessibile-error-modal .close-modal", "#accessibile-error-modal", ".content-wrapper");
```
bjacobel: `accessibile` is spelled with an extra `i` throughout this file

dianakhuang:
Addressed comments and updated sandbox.

bjacobel:
:thumbsup:

dianakhuang:
@catong Could you do another quick pass through the user-facing strings? Thanks!

catong:
@dianakhuang This looks great! 👍

marcotuts:
👍  thanks @dianakhuang (found this PR after all.) JIRA doesn't seem to link to PRs in repos that aren't edx-platform. That explains the non-link in JIRA.

PR #303, opened 2016-08-25
Don't error out if the credit service returns None.

dianakhuang:
https://openedx.atlassian.net/browse/TNL-5307

Small fix. Prevents code from erroring out when a `credit_service` is not found.

Reviewers: @robrap @bjacobel

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -899,9 +899,19 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, 
             return_course_info=True
         )
 
+        default_name = _('your course')
+        if credit_state:
+            course_name = credit_state.get('course_name', default_name)
+        else:
+            course_name = default_name
+            log.info(
+                "Could not find credit_state for user id %r in the course %r.",
+                exam_attempt_obj.user_id,
+                exam_attempt_obj.proctored_exam.course_id
+            )
         send_proctoring_attempt_status_email(
             exam_attempt_obj,
-            credit_state.get('course_name', _('your course'))
+            course_name
         )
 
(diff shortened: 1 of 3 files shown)


bjacobel:
:shipit:

edx_proctoring/api.py
```diff
@@ -899,10 +899,17 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
...
-            credit_state.get('course_name', _('your course'))
-        )
+        if credit_state:
+            send_proctoring_attempt_status_email(
+                exam_attempt_obj,
+                credit_state.get('course_name', _('your course'))
+            )
+        else:
+            log.error(
+                "Could not retrieve the credit state for the user with the user id %r in the course %r.",
```
robrap: Are we fixing an actual issue?  It's unclear if this should be an error, warning, or info.  What are the cases that the credit state service would not be available?

Also, the message seems misleading, because we never even get this far.  Maybe:
```
The credit state service is not available.  Skipping proctoring attempt status email for the user with the user id %r in the course %r.
```
?

dianakhuang: The credit service returns None if the student is not enrolled in the course, or the course doesn't have credit. I figured that it was a little difficult to fix this entirely without more information, which is why I decided that logging the issue would be more helpful for now.

I'll make the error message more specific now.

robrap: Discussed in person.  I thought this was the `credit_service` and not `credit_state`, so part of my comment doesn't make sense.  I would switch this to info log as discussed.

dianakhuang:
@robrap I updated the code as discussed. Could you take another look?

robrap:
👍

PR #308, opened 2016-09-22
Clean up the Python tests

andy-armstrong:
This change cleans up the Python tests to make them easier to maintain. It also adds missing requirements necessary to run the tests within a clean virtual environment.

@dianakhuang @robrap please review.


edx_proctoring/tests/test_api.py
```diff
@@ -1,12 +1,10 @@
 # coding=utf-8
```
andy-armstrong: Reviewers: GitHub is making this look complicated, but all I did is moved a number of tests out into two new test files, to make this file more manageable.

robrap:
@andy-armstrong: You probably noticed, but this change exposed all the existing code to pep8.  You've got fixes needed there that are clearly a pre-req for merging.

local_requirements.txt
```diff
@@ -5,6 +5,15 @@ djangorestframework>=3.1,<3.2
...
 pytz>=2012h
 pycrypto>=2.6
+python-dateutil==2.1
 
+# edX packages
 -e git+https://github.com/edx/event-tracking.git@0.2.1#egg=event-tracking==0.2.1
--e git+https://github.com/edx/opaque-keys.git@27dc382ea587483b1e3889a3d19cbd90b9023a06#egg=opaque-keys
+edx-opaque-keys==0.3.4
+
+# Used for testing
```
dianakhuang: Why do we need this here as well as in `test_requirements.txt`? Maybe we just need to update the documentation to say that those requirements need to be installed as well for local testing.

andy-armstrong: That's a great point. I've removed the test changes and updated the wiki instead to say to install the test requirements:

https://openedx.atlassian.net/wiki/display/TNL/Proctoring+FAQ

andy-armstrong:
@dianakhuang I've addressed your feedback so please re-review. Thanks.

dianakhuang:
👍  from me now!

PR #310, opened 2016-09-26
Create a mock backend that bypasses proctoring step.

dianakhuang:
This became much easier when I discovered the `threading` module.

Sandbox not available yet, but I'll add it as soon as it's ready.

https://openedx.atlassian.net/browse/TNL-5129

Reviewers: @andy-armstrong @robrap

### DIFF
--- a/edx_proctoring/backends/mock.py
+++ b/edx_proctoring/backends/mock.py
@@ -1,0 +1,91 @@
+"""
+Implements a mock proctoring backend provider to be used for testing,
+which doesn't require the setup and configuration of the Software Secure backend provider.
+"""
+import threading
+from edx_proctoring.callbacks import start_exam_callback
+from edx_proctoring.backends.backend import ProctoringBackendProvider
+
+
+class MockProctoringBackendProvider(ProctoringBackendProvider):
+    """
+    Implementation of the ProctoringBackendProvider that bypasses proctoring setup.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def register_exam_attempt(self, exam, context):
+        """
+        Called when the exam attempt has been created but not started
+
+        Args
+            exam (dict) - a dictionary containing information about the exam
+               keys:
+                 exam_name (str) - the name of the exam
+                 course_id (str) - serialized course key of the course that the exam belongs to
+                 id (str) - the id of the given exam
+            context (dict) - a dictionary containing information about the current exam context
+               keys:
+                 attempt_code (str) - code that represents this current attempt
+                 time_limit_mins (str) - time limit of exam in minutes
+                 is_sample_attempt (bool) - True if this is a practice exam
+                 callback_url (str) - url that we would like the external grader to call back to
+                 full_name (str) - full name of the student taking the exam
+                 review_policy (dict) - the policy for reviewing this exam
+                 review_policy_exception (dict) - an exceptions that may exist for this review
+                 email (str) - the email address of the student
+
+        Returns
+            string that corresponds to the backend record locator of the attempt
+        """
+        # Since the code expects this callback to be coming from the external software, we wait
+        # to call it so that the attempt is created properly before the callback is called.
+        timer = threading.Timer(1.0, start_exam_callback, args=(None, context['attempt_code']))
+        timer.start()
+        return context['attempt_code']
+
+    def start_exam_attempt(self, exam, attempt):
+        """
+        Method that is responsible for communicating with the backend provider
+        to establish a new proctored exam
+        """
+        return None
+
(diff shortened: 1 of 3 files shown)


edx_proctoring/backends/mock.py
```diff
@@ -0,0 +1,90 @@
...
+                 review_policy (dict) - the policy for reviewing this exam
+                 review_policy_exception (dict) - an exceptions that may exist for this review
+                 email (str) - the email address of the student
+
+        Returns
+            string that corresponds to the backend record locator of the attempt
+        """
+        # Since the code expects this callback to be coming from the external software, we wait
+        # to call it so that the attempt is created properly before the callback is called.
+        timer = threading.Timer(2.0, start_exam_callback, args=(None, context['attempt_code']))
```
robrap: When you call `start_exam_callback` to early it presumably gives an exception.  If you want to make this code a bit more resilient, you could wrap start_exam_callback that makes a couple of attempts, until it might finally fail/give up.  In that way, you might be able to:
1. Shrink the 2 second wait to 1 second, if that matters, and
2. Have a call that is more likely to succeed under a load test if the server does get backed up, assuming that Software Secure would retry if it had problems connecting to the callback url.

I can see an argument for putting this off unless we have an issue.  At the same time, if you start getting errors in load testing, it may not be clear that it is due to false starts on this callback, and you may save time down the line.

edx_proctoring/backends/tests/test_backend.py
```diff
@@ -149,3 +150,23 @@ def test_null_provider(self):
...
+        Test that the mock backend provider does what we expect it to do.
+        """
+        provider = MockBackendProvider()
+        attempt_code = "test_code"
+        self.assertEqual(
+            attempt_code,
+            provider.register_exam_attempt(None, {'attempt_code': attempt_code})
+        )
+
+        self.assertEqual(
```
robrap: Shouldn't this also assert that `start_exam_callback` is eventually called, or is that too difficult?

edx_proctoring/backends/mock.py
```diff
@@ -0,0 +1,90 @@
+"""
+Implementation of a backend provider, which immitates a software secure backend
```
robrap: typo: imitates

robrap: What about something like this:
```
Implements a mock proctoring backend provider to be used for testing, which doesn't require the setup and configuration of the Software Secure backend provider.
```

robrap:
@dianakhuang: This is nice and simple.  It's surprising that this is all that is needed to get through a whole proctored exam with a mock, but that would be great.  Thanks.  Done with pass.

dianakhuang:
Sandbox that uses this backend: https://proctoring-mock-backend.sandbox.edx.org

edx_proctoring/backends/mock.py
```diff
@@ -0,0 +1,91 @@
...
+        timer = threading.Timer(1.0, start_exam_callback, args=(None, context['attempt_code']))
+        timer.start()
+        return context['attempt_code']
+
+    def start_exam_attempt(self, exam, attempt):
+        """
+        Method that is responsible for communicating with the backend provider
+        to establish a new proctored exam
+        """
+        return None
```
andy-armstrong: I wonder if this backend should track whether the exam has been started or not. It seems that that might be useful when debugging.

edx_proctoring/backends/mock.py
```diff
@@ -0,0 +1,91 @@
...
+        Args
+            payload (dict) -
+                the payload from the external service - is service dependent
+
+        Returns nothing
+
+        Side Effects
+            - updates review status
+        """
+        return None
```
andy-armstrong: Should this have the documented side effect of updating the review status?

edx_proctoring/backends/mock.py
dianakhuang: This is a method that's part of the `Backend` interface that gets used literally nowhere in the code (the Software Secure implementation also just returns `None`). I left it in to match the other backend implementations.

dianakhuang:
@andy-armstrong I think we might need to talk during standup or via HipChat about our assumptions about the purposes of the mock backend. I mostly wrote this with the intention of unblocking the load testing/sandbox setup flows. We can add review updates/interactions later if we need that functionality, but I don't think it's necessary for this iteration of the mock backend.

edx_proctoring/backends/mock.py
andy-armstrong: Bizarre...

andy-armstrong:
FYI here's the ticket I created to enhance the mock backend beyond this PR: https://openedx.atlassian.net/browse/TNL-5662

robrap:
👍

PR #311, opened 2016-09-30
Make the launch callback polling interval variable.

dianakhuang:
Suggestions on better naming / initial values for the constants would be helpful. Let me know if you want a sandbox, but the user-visible behavior isn't that easy to see.

Reviewers @andy-armstrong @robrap


robrap:
@andy-armstrong: FYI: I have second thoughts about our approach and have been discussing with @dianakhuang.  We may set up a call with Software Secure.  It may not make sense for you to dig too deeply into this review until we can discuss more on Monday.

robrap:
@dianakhuang: Here are my current thoughts:
1. I don't think we'll be able to rely on getting any new endpoints from SS in the short-term.
2. Starting with 30s instead of 60s for SS Client polling bumps the estimate from 7275 to 9650 req/min.  Given the conservative nature of the estimates, I think we should go to 30s.
3. When determining our grace periods (or whatever we call them all), I think we should account for some number of seconds of delay per poll to account for slow connections.  Maybe 5s?  Maybe talk to Dave about this?  (This may not affect this PR.  This is just my overall thinking.)
4. The end exam time period would thus be 30 + 15 + 2x5 = 55s.  This is longer than Marco had hoped, but may have to do for now.

Let me know your thoughts.  I'm going to continue the review assuming the above.  Sorry to have blocked this work.

edx_proctoring/callbacks.py
```diff
@@ -152,12 +155,17 @@ def get(self, request, attempt_code):  # pylint: disable=unused-argument
             )
 
         update_exam_attempt(attempt['id'], last_poll_timestamp=timestamp, last_poll_ipaddr=ip_address)
+        time_remaining_seconds = get_time_remaining_for_attempt(attempt)
+        polling_interval = constants.DEFAULT_CLIENT_POLLING_INTERVAL
+        if time_remaining_seconds < constants.EXAM_CONCLUDING_INTERVAL:
```
robrap: I think we need to switch to reduced polling when status is submitted in place of checking for time remaining.  Let's discuss if need be.

This may not affect anything, but what happens when time runs out in the exam?  Do they have to submit before end of exam, or is it auto-submitted when time runs out?

edx_proctoring/constants.py
```diff
@@ -55,4 +55,23 @@
...
+    else getattr(settings, 'DEFAULT_CLIENT_POLLING_INTERVAL', 30)
+)
+
+REDUCED_CLIENT_POLLING_INTERVAL = (
+    settings.PROCTORING_SETTINGS['REDUCED_CLIENT_POLLING_INTERVAL'] if
+    'REDUCED_CLIENT_POLLING_INTERVAL' in settings.PROCTORING_SETTINGS
+    else getattr(settings, 'REDUCED_CLIENT_POLLING_INTERVAL', 10)
+)
+
+EXAM_CONCLUDING_INTERVAL = (
```
robrap: I think this will go away.

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -190,10 +190,7 @@
       var _poll_interval = null;
```
robrap: I think I'd adjust this code as follows:
1.  Add the following:
```
var _poll_interval_seconds = null;
```
2. Add a `set_poll_interval(new_polling_interval)` function that will set the interval if _poll_interval is null, and clear and set if _poll_interval_seconds is different from new_polling_interval.  We only want to clear for a change.
3. In `ready()`, I'd make an initial call to `set_poll_interval({{default polling interval}})` before calling `poll_exam_status`.
4. Adjust `poll_exam_status` to call `set_poll_interval`.

edx_proctoring/tests/test_views.py
```diff
@@ -835,6 +836,29 @@ def test_attempt_status_for_exception(self):
                 )
                 self.assertEqual(response.status_code, 200)
 
+    @patch('edx_proctoring.callbacks.get_time_remaining_for_attempt')
+    def test_attempt_status_poll_interval(self, time_remaining_mock):
```
robrap: May be based on status, and not time remaining.

edx_proctoring/constants.py
```diff
@@ -46,7 +46,7 @@
 SOFTWARE_SECURE_CLIENT_TIMEOUT = (
     settings.PROCTORING_SETTINGS['SOFTWARE_SECURE_CLIENT_TIMEOUT'] if
     'SOFTWARE_SECURE_CLIENT_TIMEOUT' in settings.PROCTORING_SETTINGS
-    else getattr(settings, 'SOFTWARE_SECURE_CLIENT_TIMEOUT', 30)
+    else getattr(settings, 'SOFTWARE_SECURE_CLIENT_TIMEOUT', 60)
 )
 
 SOFTWARE_SECURE_SHUT_DOWN_GRACEPERIOD = (
```
robrap: Can we add a comment to describe what this is?  Also, was it meant to be 2 heartbeats (e.g. 2x5s)?  Would it not be DEFAULT_CLIENT_POLLING_INTERVAL + REDUCED_CLIENT_POLLING_INTERVAL, or 30s + 15s = 45s?  Does it need a little extra wiggle room, like fixed 5-10s, or 3 heartbeats of DEFAULT_CLIENT_POLLING_INTERVAL + 2xREDUCED_CLIENT_POLLING_INTERVAL, or 30s + 2x15s = 60s.

edx_proctoring/constants.py
```diff
@@ -55,4 +55,23 @@
...
+DEFAULT_CLIENT_POLLING_INTERVAL = (
+    settings.PROCTORING_SETTINGS['DEFAULT_CLIENT_POLLING_INTERVAL'] if
+    'DEFAULT_CLIENT_POLLING_INTERVAL' in settings.PROCTORING_SETTINGS
+    else getattr(settings, 'DEFAULT_CLIENT_POLLING_INTERVAL', 30)
+)
+
+REDUCED_CLIENT_POLLING_INTERVAL = (
+    settings.PROCTORING_SETTINGS['REDUCED_CLIENT_POLLING_INTERVAL'] if
+    'REDUCED_CLIENT_POLLING_INTERVAL' in settings.PROCTORING_SETTINGS
+    else getattr(settings, 'REDUCED_CLIENT_POLLING_INTERVAL', 10)
```
robrap: In the spreadsheet, we have this as 15s.

edx_proctoring/callbacks.py
```diff
@@ -152,12 +155,17 @@ def get(self, request, attempt_code):  # pylint: disable=unused-argument
             )
 
         update_exam_attempt(attempt['id'], last_poll_timestamp=timestamp, last_poll_ipaddr=ip_address)
+        time_remaining_seconds = get_time_remaining_for_attempt(attempt)
```
dianakhuang: This was one of my initial concerns with this story. It stops at `0`, even before the exam begins, etc. We could change it to return `None` if it doesn't make sense that there's time remaining.

edx_proctoring/callbacks.py
dianakhuang: When time runs out, the status becomes `timed_out`, I believe. There's a way of generalizing all the statuses into 'complete' statuses. There's a `ProctoredExamStudentAttemptStatus.is_completed_status` that we can use.

edx_proctoring/constants.py
```diff
@@ -55,4 +55,23 @@
     else getattr(settings, 'SOFTWARE_SECURE_SHUT_DOWN_GRACEPERIOD', 10)
 )
 
+DEFAULT_CLIENT_POLLING_INTERVAL = (
+    settings.PROCTORING_SETTINGS['DEFAULT_CLIENT_POLLING_INTERVAL'] if
+    'DEFAULT_CLIENT_POLLING_INTERVAL' in settings.PROCTORING_SETTINGS
+    else getattr(settings, 'DEFAULT_CLIENT_POLLING_INTERVAL', 30)
```
dianakhuang: Yes, the idea was that the numbers would get 'fixed' for this PR when it was complete. I wanted to make sure the code was correct first, which seems like a good idea in retrospect.

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -187,12 +187,25 @@
...
 
       $(document).ready(function() {
-        poll_exam_status();
+          set_poll_interval({{default_polling_interval}});
+          poll_exam_status();
       });
 
+      function set_poll_interval(new_polling_interval) {
+
+          // The polling interval will be in seconds, so we need to multiply by 1000 to get the interval in miliseconds.
```
robrap: typo: milliseconds

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -187,12 +187,25 @@
...
+          new_pollint_interval = new_polling_interval * 1000;
+          if (new_polling_interval != _current_interval) {
+              clearInterval(_poll_interval_id);
```
robrap:
```
_current_interval = new_polling_interval;
```

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -187,12 +187,25 @@
...
       $(document).ready(function() {
-        poll_exam_status();
+          set_poll_interval({{default_polling_interval}});
+          poll_exam_status();
       });
 
+      function set_poll_interval(new_polling_interval) {
+
+          // The polling interval will be in seconds, so we need to multiply by 1000 to get the interval in miliseconds.
+          new_pollint_interval = new_polling_interval * 1000;
```
robrap: Oops.  I added these as commit comments.  I am going to leave them.

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -206,6 +217,9 @@
             // NOTE: This is per the API documentation from SoftwareSecure
             window.external.quitApplication();
           }
+
+          // The polling interval is returned in seconds, so we need to multiply by 1000.
```
robrap: You can drop this comment here.  It moved.

edx_proctoring/templates/proctored_exam/proctoring_launch_callback.html
```diff
@@ -206,6 +217,9 @@
...
+          set_poll_interval(parseInt(data.polling_intervall));
```
robrap: typo: polling_intervall => polling_interval

edx_proctoring/constants.py
```diff
@@ -46,7 +46,7 @@
 SOFTWARE_SECURE_CLIENT_TIMEOUT = (
     settings.PROCTORING_SETTINGS['SOFTWARE_SECURE_CLIENT_TIMEOUT'] if
     'SOFTWARE_SECURE_CLIENT_TIMEOUT' in settings.PROCTORING_SETTINGS
-    else getattr(settings, 'SOFTWARE_SECURE_CLIENT_TIMEOUT', 30)
+    else getattr(settings, 'SOFTWARE_SECURE_CLIENT_TIMEOUT', 60)
```
robrap: @dianakhuang: One thought is you add a commit now for the final constants.  In other words, revert this back to 30 and make `DEFAULT_CLIENT_POLLING_INTERVAL` and `REDUCED_CLIENT_POLLING_INTERVAL` both 5.  It may be quick to get this landed and do the more complex testing with the other constant changes in the other story, rather than doing complex testing twice.

Thoughts?

robrap:
Looks good.  Just waiting on the sandbox.

PR #312, opened 2016-10-04
Release 0.14.0

andy-armstrong:
@dianakhuang @robrap please review this new edx-proctoring release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,14 +34,14 @@ def load_requirements(*requirements_paths):
 
 setup(
     name='edx-proctoring',
-    version='0.13.0',
+    version='0.14.0',
     description='Proctoring subsystem for Open edX',
     long_description=open('README.md').read(),
     author='edX',
     url='https://github.com/edx/edx-proctoring',
     license='AGPL',
     classifiers=[
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
         'Environment :: Web Environment',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -34,14 +34,14 @@ def load_requirements(*requirements_paths):
...
-    version='0.13.0',
+    version='0.14.0',
     description='Proctoring subsystem for Open edX',
     long_description=open('README.md').read(),
     author='edX',
     url='https://github.com/edx/edx-proctoring',
     license='AGPL',
     classifiers=[
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
```
robrap: Out of curiosity, do you feel this is just entering Beta, or are you just getting this in line with where you think it has been?  Do we docs on this, or do we just wing it?

robrap:
👍

setup.py
andy-armstrong: Good question. I found the list of valid classifiers here:

https://pypi.python.org/pypi?%3Aaction=list_classifiers

My thinking is that there is no way this is an alpha since we're using it in production. I was tempted to go with ``Development Status :: 5 - Production/Stable`` but then figured that should probably wait until we fix the polling issues. At that point I think we should switch to version 1.0.0 and mark it as production/stable. 

Does that make sense?

robrap: Yes.  Thank you.

PR #313, opened 2016-10-19
WiP: Add a button to the proctoring flow if clint app

dianakhuang:
needs to be closed.


robrap:
@dianakhuang: How many times can that Proctored exam be taken?  Is only 1 reviewer allowed to review in the sandbox?

dianakhuang:
@robrap the staff user can reset the verified user's exam attempt status, allowing multiple reviewers to review the sandbox.

chris-mike:
@dianakhuang I guess I'm a little out of the loop. Can you either send screenshots for what I should be looking at or steps to produce the new button?

dianakhuang:
Sorry about forgetting the screenshots, @chris-mike ! It's really just this new button:

Button screenshot

edx_proctoring/tests/test_utils.py
```diff
@@ -66,3 +70,18 @@ def test_emit_event(self):
                 'one': 'two'
             }
         )
+
+    def test_has_client_app_shutdown(self):
```
robrap: I don't get where this test is coming from and for, given that we are removing polling?  Was this moved, but I'm not seeing the original?  Other?

edx_proctoring/views.py
```diff
@@ -422,12 +422,18 @@ def put(self, request, attempt_id):
                     exam_id=attempt['proctored_exam']['id'],
                     user_id=request.user.id
                 )
-            elif action == 'submit':
+            elif action == 'timed_submit' or action == 'client_close':
```
robrap:
```
elif action in ['timed_submit', 'client_close']:
```

edx_proctoring/views.py
```diff
@@ -422,12 +422,18 @@ def put(self, request, attempt_id):
...
                 exam_attempt_id = update_attempt_status(
                     attempt['proctored_exam']['id'],
```
robrap: Why would `timed_submit` above be used with proctored_exam attempt?  Or does `timed_submit` not mean what I think?

dianakhuang: There used to be one `submit` which I now have split into `timed_submit` and `proctored_submit` depending on whether it's a timed exam or a proctored exam, because the transitions after a timed exam behave slightly differently than the transitions after a proctored exam now.

PR #314, opened 2016-10-21
Delete in-client polling.

dianakhuang:
### Description

Removes polling from the proctoring client and the 'waiting for app shutdown' step from the end of exam. Also adds text to the final 'submitted' page to alert users that they will need to close the application.

### Sandbox
- [x] https://dianakhuang.sandbox.edx.org/ - the verified user is set up to take the 'proctored' exam.

To view the (former) polling page, it can be accessed at this url: https://dianakhuang.sandbox.edx.org/api/edx_proctoring/proctoring_launch_callback/start_exam/[exam_code]

### Testing
- [x] i18n
- [x] Unit, integration, acceptance tests as appropriate

### Reviewers
- [x] Code review: @andy-armstrong 
- [x] Code review: @robrap 
- [x] Doc Review: @catong 
- [ ] Product review: @marcotuts 

FYI: Tag anyone who might be interested in this PR here.

### Post-review


robrap:
@dianakhuang: On the sandbox, I could not start an exam because "Start System Check" gave a 404 for `/courses/course-v1:Proctoring+Proctoring+Proctoring/courseware/3b02972013a049a08badae8d7f4aab05/79af1cd0f5e4483a8a0f5d655f0ea848/mockurl`.

dianakhuang:
@robrap oh, the sandbox is set up with the mock proctoring backend instead of the software secure one. You don't have to do the System Check step.

edx_proctoring/templates/proctored_exam/submitted.html
```diff
@@ -7,6 +7,11 @@
   </h3>
   <p>
     {% blocktrans %}
+    If your proctoring client is still open, please close it now.
```
robrap: This is more a question for @catong, but we may want to use "proctoring session" consistently.

catong: @dianakhuang Here is my suggestion, to more closely match what's in the current learner doc. Let me know what you think:
"If the proctoring software window is still open, you can close it now. Confirm that you want to quit the application when you are prompted."

edx_proctoring/tests/test_student_view.py
```diff
@@ -717,40 +716,26 @@ def test_get_studentview_submitted_status(self):
                 'default_time_limit_mins': 90
             }
         )
-        self.assertIn(self.proctored_exam_waiting_for_app_shutdown_msg, rendered_response)
+        self.assertIn(self.proctored_exam_submitted_msg, rendered_response)
 
-        reset_time = datetime.now(pytz.UTC) + timedelta(minutes=2)
```
robrap: Search for `last_poll_timestamp`, like above, and delete code that is setting (and probably using) it.  Thanks!

edx_proctoring/views.py
```diff
@@ -336,27 +331,6 @@ def get(self, request, attempt_id):
             # if we never heard from the client, then we assume it is shut down
             attempt['client_has_shutdown'] = last_poll_timestamp is None
```
robrap: Can we remove this line and `last_poll_timestamp = ...` line above (with comments)?  Can you also search and make sure `client_has_shutdown` and `last_poll_timestamp` are basically never used?

robrap:
@dianakhuang: Thanks for the clean-up.  Minor comments.

Also, can you

robrap:
@andy-armstrong: FYI: We may need to hotfix this tomorrow or Thursday.

robrap:
@andy-armstrong: What we turned off was the auto-shutdown of an exam due to a missing heartbeat.  However, the polling itself is still active.  We suspect, but are unable to prove, that the traffic from the polling for one learner is too much for her system to handle and is making it so that submitting answers to problems is sometimes failing.

It does raise the question of whether or not we should make it more clear that an answer didn't make it through and let the user know what to do in this case.

dianakhuang:
Sandbox has been updated with the latest changes.

edx_proctoring/tests/test_views.py
```diff
@@ -901,77 +808,6 @@ def test_attempt_status_stickiness(self):
...
-    @ddt.data(
-        ProctoredExamStudentAttemptStatus.created,
-        ProctoredExamStudentAttemptStatus.ready_to_start,
-        ProctoredExamStudentAttemptStatus.started,
-        ProctoredExamStudentAttemptStatus.ready_to_submit
-    )
-    def test_attempt_callback_timeout(self, running_status):
-        """
-        Ensures that the polling from the client will cause the
-        server to transition to timed_out if the user runs out of time
```
robrap: Is this in addition to the banner polling?  Wouldn't this have already happened with the banner polling?  Can we create a 1 minute exam and test?

edx_proctoring/views.py
```diff
@@ -328,35 +323,6 @@ def get(self, request, attempt_id):
...
-                                attempt['proctored_exam']['id'],
-                                attempt['user']['id'],
-                                ProctoredExamStudentAttemptStatus.error
-                            )
-                            attempt['status'] = ProctoredExamStudentAttemptStatus.error
-                        except ProctoredExamIllegalStatusTransition:
-                            # don't transition a completed state to an error state
-                            pass
-
             # add in the computed time remaining as a helper to a client app
```
robrap: Do you know what this comment means and if there is additional clean-up needed around ` attempt['time_remaining_seconds']`?  Is `attempt['accessibility_time_string']` being used?  If one or both of these are used, maybe this comment just needs to be improved or removed.

dianakhuang: This endpoint is the one that gets used by the banner polling, so I assume that these numbers are used by the banner, not by the SS client.

robrap: The term "client app" is confusing in the case of proctoring.  I'd just delete this comment altogether.  Feel free to do that with the squash.

robrap:
👍 once you remove the misleading comment.

robrap:
Nice work on the deep cleanse too.  Thanks!

PR #318, opened 2016-10-31
Fix icon alignment for practice exams

andy-armstrong:
@marcotuts noticed an alignment bug for practice exams which I'd forgotten to verify. This fix moves the icon span before the paragraph tag, in order to make the right alignment work as expected.

@dianakhuang @robrap please review.

P.S. I'm updating the sandbox now.


andy-armstrong:
The sandbox has now updated, and I've verified that the fix worked.

robrap:
👍  Code looks ok, but I did not have time to review the sandbox.  More importantly would be that it actually fixes what you wanted and causes no other issue.

dianakhuang:
👍

PR #319, opened 2016-11-03
Increment banner polling from 30 seconds to 60 seconds.

dianakhuang:
Reviewers: @robrap @andy-armstrong 

Look, there's even a test!

### DIFF
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
@@ -20,6 +20,7 @@ var edx = edx || {};
             this.secondsLeft = 0;
             /* give an extra 5 seconds where the timer holds at 00:00 before page refreshes */
             this.grace_period_secs = 5;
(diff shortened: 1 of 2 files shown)


edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -136,7 +136,7 @@ var edx = edx || {};
         updateRemainingTime: function (self) {
             self.timerTick ++;
             self.secondsLeft --;
-            if (self.timerTick % 30 === 0){
+            if (self.timerTick % 60 === 0){
```
andy-armstrong: Nit: I think it would be clearer if this was a constant that was named as to its purpose.

andy-armstrong: I also noticed that this line doesn't match our ESLint rules, so I guess that isn't applied in this repo! There should be a space before the opening ``{``.

robrap:
👍

PR #322, opened 2016-12-16
Update messaging to mention MicroMasters

andy-armstrong:
https://openedx.atlassian.net/browse/TNL-6113

This is a one line text change, along with corresponding test fixes.

@catong @dianakhuang, please review.

### DIFF
--- a/edx_proctoring/templates/proctored_exam/entrance.html
+++ b/edx_proctoring/templates/proctored_exam/entrance.html
@@ -7,7 +7,7 @@
   </h3>
   <p>
     {% blocktrans %}
-      To be eligible to earn credit for this course, you must take and pass the proctoring review for this exam.
+      To be eligible for course credit or for a MicroMasters credential, you must pass the proctoring review for this exam.
     {% endblocktrans %}
   </p>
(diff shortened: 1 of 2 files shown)


edx_proctoring/templates/proctored_exam/entrance.html
```diff
@@ -7,7 +7,7 @@
   </h3>
   <p>
     {% blocktrans %}
-      To be eligible to earn credit for this course, you must take and pass the proctoring review for this exam.
+      To be eligible for course credit or for a MicroMasters credential, you must pass the proctoring review for this exam.
```
catong: Should we make "MicroMasters" a variable?

catong:
👍 with my question about hardcoding "MicroMasters".

edx_proctoring/templates/proctored_exam/entrance.html
andy-armstrong: I considered that, but then saw that it is hard-coded in other strings in edX platform. It is an interesting point that this message is very edX-specific now, which isn't ideal for an open source library. @dianakhuang do you have any recommendations?

dianakhuang:
👍  from me as well.

PR #324, opened 2016-12-20
Upgrade proctoring to use Transifex

andy-armstrong:
### Description

This change upgrades edx-proctoring to support localization through Transifex. In order to achieve this, the library was regenerated using https://github.com/edx/cookiecutter-django-app which adds Transifex support as well as adopting edX best practices for building a Django app. 

### Sandbox
- [ ] _TBD_

### Testing
- [ ] Manual verification of localization on my sandbox

FYI: @rlucioni

### Post-review


andy-armstrong:
@rmoch @laurentdavid I am starting to switch the edx-proctoring repository over to use Transifex for localization. As part of this process, I will be replacing your hand-crafted French .po file:

https://github.com/edx/edx-proctoring/blob/master/locale/fr/LC_MESSAGES/django.po

Before this merges, we need to move all the translations into Transifex. Is that something you would be comfortable doing? I can attempt to do it myself, but I will almost certainly make mistakes.

laurentdavid:
Hi Andy,

Thanks for reaching us. Happy to do it over the next few weeks.

Laurent

.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # Installation artifacts
 src
+requirements/*.txt
```
dianakhuang: Are these generated by something else? It seems like we don't want to ignore these?

edx_proctoring/tests/test_views.py
```diff
@@ -2,14 +2,18 @@
 """
 All tests for the proctored_exams.py
 """
+
+from __future__ import absolute_import
+
+from datetime import datetime, timedelta
 import json
-import pytz
+from string import Template  # pylint: disable=deprecated-module
```
dianakhuang: Is there a way of converting this to not using a deprecated module? It just makes me nervous about not going back to clean it up later.

.gitignore
andy-armstrong: Yes, these are generated by pip-tools through the make command:

    make upgrade

The idea is that it is easy to pin the requirements but also easy to refresh them all as needed:

https://github.com/nvie/pip-tools

andy-armstrong: BTW, this is a standard part of the Django app cookicutter: https://github.com/edx/cookiecutter-django-app

edx_proctoring/tests/test_views.py
andy-armstrong: Good call. I've converted this weird template logic to instead use a parameterized function.

andy-armstrong:
@dianakhuang @robrap this is ready for re-review. I'm going to update my sandbox now.

edx_proctoring/__init__.py
```diff
@@ -1,3 +1,9 @@
 """
-Init file for the top level edx_proctoring directory
+The exam proctoring subsystem for the Open edX platform.
 """
+
+from __future__ import absolute_import
+
+__version__ = '0.1.0'
```
robrap: Is the version that should be pulled in in setup.py?

setup.py
```diff
@@ -1,60 +1,69 @@
...
+    os.system("git push --tags")
+    sys.exit()
+
+README = open(os.path.join(os.path.dirname(__file__), 'README.rst')).read()
+CHANGELOG = open(os.path.join(os.path.dirname(__file__), 'CHANGELOG.rst')).read()
 
 setup(
     name='edx-proctoring',
-    version='0.16.2',
+    version='0.17.0',
```
robrap: Should this be using `VERSION` above that comes from `__init__.py`?

edx_proctoring/models.py
```diff
@@ -173,11 +177,14 @@ class ProctoredExamStudentAttemptStatus(object):
...
     status_alias_mapping = {
-        submitted: _('pending'),
-        verified: _('satisfactory'),
-        rejected: _('unsatisfactory')
+        submitted: ugettext_noop('pending'),
+        verified: ugettext_noop('satisfactory'),
+        rejected: ugettext_noop('unsatisfactory')
     }
 
+    def __unicode__(self):
```
andy-armstrong: I added these because there was a PEP8 failure for each model that didn't have a ``__unicode__`` method. The code in edx-platform seems to be very inconsistent about which one to use. 

It seems that probably ``__repr__`` is the better choice, so I'll try converting these and see what happens:

http://stackoverflow.com/questions/1984162/purpose-of-pythons-repr

edx_proctoring/__init__.py
andy-armstrong: Great catch. I've updated this to be the true version number, and fixed setup.py to use it.

setup.py
andy-armstrong: Yes, it should. Fixed.

bjacobel:
Codecov is working, build is failing because of high threshold but the good news is between the python (previously 100% coverage) and the JS (previously untracked in coveralls, around 70% coverage) we're pulling 90%. May want to adjust the thresholds?

edx_proctoring/__init__.py
robrap: @andy-armstrong: I'm not seeing any change.  Do you still need to push something?

edx_proctoring/models.py
robrap: Same issue here. I was asked to re-review in stand-up, but looks like that may have been premature.  Can you ping me when it is ready?  Thanks.

README.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Overview
+--------
+
+Proctored exams are exams with time limits that learners complete while online
+proctoring software monitors their computers and behavior for activity that
+might be evidence of cheating. This Python library provides the proctoring
+implementation used by Open edX.
+
+Documentation
```
andy-armstrong: I've added documentation for the basic make commands. I decided against documenting all of the Transifex commands beyond the help command.

andy-armstrong:
@robrap I've got a green build now! Please re-review at your convenience.

andy-armstrong:
@dianakhuang @robrap I have everything working, *except* that I can't get the translations to show up on my sandbox (they work correctly on my devstack). I think that I'm not packaging the library correctly, but nothing that I've tried has worked. Let's discuss it in the morning:

https://andy-armstrong.sandbox.edx.org/courses/course-v1:AndyA+AAPTC+1/courseware/9b7aeac146d54fc7a552f2feb2b00f51/be3702e5245a4d6f96194449303bdbf3/

andy-armstrong:
@dianakhuang @robrap Well, I realized why my sandbox wasn't updating. It was because the version number wasn't changing, so pip install didn't think it needed to install the latest version. I pip uninstalled and reinstalled, and now the translations are working:

https://andy-armstrong.sandbox.edx.org/courses/course-v1:AndyA+AAPTC+1/courseware/9b7aeac146d54fc7a552f2feb2b00f51/be3702e5245a4d6f96194449303bdbf3/

PR #327, opened 2017-01-11
Add new character to list of invalid characters.

dianakhuang:
Tests exist here: https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/backends/tests/test_software_secure.py#L246

### DIFF
--- a/edx_proctoring/backends/software_secure.py
+++ b/edx_proctoring/backends/software_secure.py
@@ -38,7 +38,7 @@ from edx_proctoring.serializers import (
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


andy-armstrong:
@dianakhuang Does it make sense to write a unit test for this? Have you determined what the behavior is currently if an ! is found?

dianakhuang:
In the current behavior, the `!` is left in the payload. When we add it to this list, it is replaced by a `_`. 

There already is a test (that I linked to) that iterates through the list of invalid characters and verifies that the invalid character is not present in the output. I don't know what other testing would be useful.

PR #328, opened 2017-01-17
Support Django app directory structures

andy-armstrong:
This commit updates edx_proctoring to support the Transifex automation described here: https://openedx.atlassian.net/wiki/display/ECOM/Automated+Translation+Management

It depends upon the following change to i18n-tools which fixes the library to understand the typical layout of a Django app library (https://github.com/edx/i18n-tools/pull/52).

### DIFF
--- a/edx_proctoring/locale/config.yaml
+++ b/edx_proctoring/locale/config.yaml
@@ -71,14 +71,3 @@ locales:
 # The locales used for fake-accented English, for testing.
 dummy_locales:
     - eo
-
-# Directories we don't search for strings.
-ignore_dirs:
-    - docs
-    - edx-proctoring
-    - logs
-    - node_modules
-    - performance
-    - requirements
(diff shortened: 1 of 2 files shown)


edx_proctoring/locale/config.yaml
```diff
@@ -71,14 +71,3 @@ locales:
...
-# Directories we don't search for strings.
-ignore_dirs:
-    - docs
-    - edx-proctoring
-    - logs
-    - node_modules
-    - performance
-    - requirements
-    - scripts
-    - settings
```
dianakhuang: Why aren't we excluding these directories anymore?

andy-armstrong: Good question. It is because we are now cd'ing to the Django app directory and running the scripts there. These directories are all in the root, so they aren't seen now.

PR #334, opened 2017-02-15
Use expire time for timeouts instead of now.

dianakhuang:
### Description

Mark an exam as 'completed' at the point of timeout rather than when the status update occurs.

### Sandbox
- [ ] https://dianakhuang.sandbox.edx.org/courses/course-v1:Proctoring+Proctoring+Proctored/courseware/faaa46a1e2a7461b9acf667c761b38d3/24b5e9edf5e64d96bb2962ec5aa20967/

Steps to reproduce:
1. Log in as the verified user
2. Start the exam
3. Close the tab
4. Leave the tab closed for > 1 minute
5. Open up the exam again
6. Verify that the 'exam submitted' page shows
7. Log in as the staff user
8. Check to see that the instructor dashboard only claims the verified user only used 1 minute to complete the exam.

### Testing
- [ ] Unit, integration, acceptance tests as appropriate

### Reviewers
- [ ] Code review: @andy-armstrong 
- [ ] Code review: @robrap 

### Post-review

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -391,7 +391,8 @@ def _check_for_attempt_timeout(attempt):
             update_attempt_status(
                 attempt['proctored_exam']['id'],
                 attempt['user']['id'],
-                ProctoredExamStudentAttemptStatus.timed_out
+                ProctoredExamStudentAttemptStatus.timed_out,
+                timeout_timestamp=expires_at
             )
             attempt = get_exam_attempt_by_id(attempt['id'])
 
@@ -715,7 +716,8 @@ def mark_exam_attempt_as_ready(exam_id, user_id):
     return update_attempt_status(exam_id, user_id, ProctoredExamStudentAttemptStatus.ready_to_start)
 
 
-def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True, cascade_effects=True):
+def update_attempt_status(exam_id, user_id, to_status,
+                          raise_if_not_found=True, cascade_effects=True, timeout_timestamp=None):
     """
     Internal helper to handle state transitions of attempt status
     """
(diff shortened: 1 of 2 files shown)


edx_proctoring/api.py
```diff
@@ -792,6 +794,8 @@ def update_attempt_status(exam_id, user_id, to_status, raise_if_not_found=True,
     )
     if add_start_time:
         exam_attempt_obj.started_at = datetime.now(pytz.UTC)
+    elif alias_timeout:
```
andy-armstrong: Nit: while you're in here, could you rename this variable as I found it very hard to understand what it means. In what way is it an alias?

dianakhuang: It was added originally over here: https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/api.py#L733

It seems to be referring to the fact that it's aliasing the timeout state to the completed state.

robrap: How about renaming it to `treat_timeout_as_submitted`?

edx_proctoring/tests/test_api.py
```diff
@@ -950,10 +950,12 @@ def test_alias_timed_out(self):
         """
 
         exam_attempt = self._create_started_exam_attempt()
+        timeout_timestamp = datetime.now(pytz.UTC) - timedelta(minutes=1)
```
robrap: This might just be me, but I was trying to determine the significance of the 1 minute delta.  Could you:
1. Rename `timeout_timestamp` to `random_timestamp`, and
2. Subtract 3 hours, or something seemingly more random.

robrap: What makes `alias_timeout` true for this test?  Do we need one for when it is false?

PR #346, opened 2017-04-05
README had wrong repo name at top

bjacobel:


andy-armstrong:
Retroactive 👍 . Thanks for fixing my screw-up.

PR #349, opened 2017-05-09
Fix quality failures with new edx-lint version

andy-armstrong:
This is a simple fix to get edx-proctoring builds to pass quality tests now that edx-lint has upgraded its version of pylint.


andy-armstrong:
@robrap I didn't think of the amnesty. Is that something we can use against any repo?

robrap:
@andy-armstrong I'm not sure.  I meant this to be a question for @cpennington. :)

PR #354, opened 2017-06-14
Only request reviewed translations from transifex

amangano-edx:
This adds the `--mode reviewed` option to the transifex command to ensure that we only pull translations that have been reviewed into the codebase from transifex. This should reduce the likelihood that bad translations make it into the application.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ detect_changed_source_translations:
 	cd edx_proctoring && i18n_tool changed
 
 pull_translations: ## pull translations from Transifex
-	tx pull -af
+	tx pull -af --mode reviewed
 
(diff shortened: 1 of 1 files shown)


rlucioni:
@amangano-edx, in case you don't know what @andy-armstrong's referring to: https://github.com/edx/cookiecutter-django-app. It might also be good to update https://github.com/edx/cookiecutter-django-ida while you're at it; the updates will look similar.

andy-armstrong:
Thanks @rlucioni. I'd intended to include a link but forgot to:

https://github.com/edx/cookiecutter-django-app/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/Makefile#L87

PR #360, opened 2017-07-03
Update for django 1.11 compatibility.

dsjen:
[EDUCATOR-766](https://openedx.atlassian.net/browse/EDUCATOR-766): Upgrade edx-proctoring to Django 1.11:
* Added django 1.11 to tox and added it to the `envlist`. This will run tests against django 1.8 & 1.11
* To support django 1.11, management commands are updated to use `parser.add_arguments()` instead of `make_option`, `patterns` is removed, and test settings are updated to indicate the template backend and that this is an app (e.g. installed w/in the platform)

Please review, @edx/educator-dahlia

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,4 +3,4 @@ Muhammad Shoaib <mshoaib@edx.org>
 Afzal Wali <afzal@edx.org>
 Mushtaq Ali <mushtaak@gmail.com>
 Christina Roberts <christina@edx.org>
-
+Dennis Jen <djen@edx.org>
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -4,6 +4,6 @@ The exam proctoring subsystem for the Open edX platform.
 
 from __future__ import absolute_import
 
-__version__ = '0.18.1'
+__version__ = '0.18.2'
 
 default_app_config = 'edx_proctoring.apps.EdxProctoringConfig'  # pylint: disable=invalid-name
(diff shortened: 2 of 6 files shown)


dsjen:
@efischer19 -- yep! I confirmed that the args work as expected (e.g. `./manage.py set_attempt_status --user 5 --exam 192 --to rejected`. I'm surprised at the drop in coverage (didn't pay it much heed earlier).

I think in order to merge this, I'll need whoever "owns" this repo to update the status checks to allow merge if the patch coverage decreases. @efischer19, do you have permissions (assuming you're ok with the drop)?

efischer19:
I am okay with the drop, do not have the needed permissions though. You'll likely need to talk to ops/IT support for that.

cahrens:
@dsjen should you also update this line in setup.py? I'm wondering about the same thing in my PR.

https://github.com/edx/edx-proctoring/blob/master/setup.py#L63

thallada:
@cahrens good point, I created a PR to update it. https://github.com/edx/edx-proctoring/pull/361

cahrens:
Another thing I'm confused about-- repositories created from the cookie cutter seem to list the different versions of django in their travis file. I don't know why this is.

For example,
https://github.com/edx/web-fragments/blob/master/.travis.yml#L5

efischer19:
That's setting up tox to run tests in parallel against each different django version.

See https://travis-ci.org/edx/web-fragments/builds/230438809 as an example from that repo

cahrens:
So as long as the test suite is short, we don't care about running in parallel, I suppose.

efischer19:
Yep. No (or very little) time penalty, lots of confidence when the time comes to upgrade django (or python) versions

PR #361, opened 2017-07-07
Support Django 1.11 in setup.py requirements

thallada:
@cahrens pointed this out. This repo now supports Django 1.11, so `setup.py` should reflect that too.

@edx/educator-dahlia

### DIFF
--- a/edx_proctoring/backends/tests/test_software_secure.py
+++ b/edx_proctoring/backends/tests/test_software_secure.py
@@ -8,7 +8,7 @@ from __future__ import absolute_import
 
 import json
 import ddt
-from mock import patch
+from mock import MagicMock, patch
 from httmock import all_requests, HTTMock
 
 from django.test import TestCase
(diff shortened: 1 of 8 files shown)


thallada:
Looks like the Django 1.11 test was actually installing Django 1.10 before this change, so there's some failures here due to Django 1.11 breaking changes. I'll try and fix them up, but it looks like we are actually only supporting 1.10 at the moment.

thallada:
Okay, @edx/educator-dahlia I fixed the Django 1.11 build.

edx_proctoring/backends/tests/test_software_secure.py
```diff
@@ -87,6 +87,7 @@ def mock_response_error(url, request):  # pylint: disable=unused-argument
         }
     }
 )
+@patch('django.core.urlresolvers.reverse', MagicMock)
```
efischer19: nice

setup.py
```diff
@@ -60,6 +63,6 @@ def get_version(*file_paths):
     ],
     include_package_data=True,
     install_requires=[
-        "Django>=1.8,<1.11"
+        "Django>=1.8,<1.12"
```
cahrens: We've standardized on <2.0 (see conversation on my milestones PR).

PR #365, opened 2017-08-02
EDUCATOR-927 Override grade to zero when exam attempt is rejected

thallada:
This is the first part of [EDUCATOR-409](https://openedx.atlassian.net/browse/EDUCATOR-409): By default, a failed proctoring review should make assessment grade go to zero.

### Part 2 (edx-platform)

This will use a new service in the edx-platform called GradesService. The implementation for that is [in my PR in edx-platform](https://github.com/edx/edx-platform/pull/15726). Once this is merged I will release a new version of edx-proctoring and then reference that version in the requirements.txt in my PR for edx-platform.

### How I tested this

In addition to the unit tests I created, I manually tested the workflow in a sandbox.

1. Created a sandbox and followed these directions to set up a proctored exam and enroll a verified user to submit an attempt: https://openedx.atlassian.net/wiki/spaces/EDUCATOR/pages/161049499/Proctoring+Dev+Workflow
2. In the django shell, I checked that the subsection grade for that user on that exam was graded.
3. Went into the admin panel and rejected the exam attempt.
4. Saw that the grade override and grade recalculate signal was logged to `/edx/var/log/lms/edx.log`
5. Saw tracking events `edx.grades.subsection.score_overridden` followed by `edx.grades.subsection.grade_calculated` and `edx.graded.course.grade_calculated` with `"event_transaction_type"` set to `edx.grades.subsection.score_overridden` in `/edx/log/var/tracking/tracking.log`
6. In the django shell, I checked the grade again and saw that both `earned_all` and `earned_graded` were now `0.0`.
7. In the admin panel, moved the attempt from rejected to verified.
8. Saw that the grade override deletion and grade recalculate signal was logged to `/edx/var/log/lms/edx.log`
9. Saw tracking events `edx.grades.subsection.score_overridden`, except this time `"override_deleted"` was `true`. Which was followed by `edx.grades.subsection.grade_calculated` and `edx.graded.course.grade_calculated` with `"event_transaction_type"` set to `edx.grades.subsection.score_overridden` in `/edx/log/var/tracking/tracking.log`
10. In the django shell, I checked that the grade for the user on the exam returned back to their original score.

We can disable the grade override on a course-by-course basis with the `rejected_exam_overrides_grade` course waffle flag, but this will be enabled for all courses by default.

### Eventing

This adds a new event `edx.grades.subsection.score_overridden` which is documented at https://openedx.atlassian.net/wiki/spaces/AN/pages/125927581/Grading+Event+Design.

This shouldn't affect any existing events besides now `edx.grades.subsection.grade_calculated` and `edx.graded.course.grade_calculated` events are sent with `"event_transaction_type"` set to `edx.grades.subsection.score_overridden`.

### Documentation

@srpearce Before part 2 of this is merged and goes live, the following needs to be documented:

1. Student facing documentation about how their grade will be set to zero automatically if they fail the proctoring review and how to appeal it.
2. Instructor facing documentation about this behavior. Also, they should know that rejected exam attempts previous to the date that this work is merged will not be affected (they won't override the learner's grade).
3. Research guide updates for the new event mentioned above.

### Open Questions

@katymyw Note that this is what the user's exam grade ("edX Exams" is the proctored exam) will look like in the Course Progress for Student panel after the override:

screenshot-20170726165124-1008x337

Will it be clear to users that an override is applied?

@edx/educator-dahlia 

FYI: @sanfordstudent


edx_proctoring/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import
 
-__version__ = '0.18.2'
+__version__ = '0.18.3'
```
thallada: Does a patch version bump sound right to people? I feel like I always do semantic versioning wrong.

efischer19: I'd argue that this is a big enough feature addition to justify a minor version increase.

I'd really like to argue that having fully-used-in-production versions that are < 1 is kind of silly, and we should just bite the bullet and declare a 1.0.0 release one of these days, but that may be overkill.

thallada: Okay, I wasn't sure since the last patch version seemed pretty big to me (django upgrade).

1.0.0 sounds good to me :tada: :champagne: :fireworks: :confetti_ball:. If no one else has objections, I'll go bump this to 1.0.0.

edx_proctoring/tests/test_api.py
```diff
@@ -912,6 +914,169 @@ def test_cascading(self, to_status, create_attempt, second_attempt_status, expec
...
+        )
+
+        # Grade has still the original score
+        self.assertDictEqual({
+            'earned_all': grade.earned_all,
+            'earned_graded': grade.earned_graded
+        }, {
+            'earned_all': 5.0,
+            'earned_graded': 5.0
+        })
```
efischer19: These tests are great - super straightforward. One could argue that they need to be smaller unit-style tests, but we want to assert things at a lot of places in the process (any failure = whole thing broken), so i think these longer acceptance-style scenario tests are the best way to do that without a whole mess of repetitive setup.

edx_proctoring/tests/test_api.py
```diff
@@ -1141,6 +1306,7 @@ def test_attempt_status_summary(self, status, expected):
         Assert that we get the expected status summaries
         """
 
+        set_runtime_service('grades', MockGradesService())
```
efischer19: Why did the service need to be added here? It's not used in this test

srpearce:
@thallada Great—I just started on the docs.

@katymyw and @thallada re: the Course Progress for Student panel: I'd definitely suggest messaging there, and on the learner's Progress page. Otherwise, learners and course teams might think we've lost the exam or something has gone wrong with grading, for example. If @thallada can add a screen shot of the learner's Progress page, I can suggest messaging for both places.

edx_proctoring/tests/test_api.py
thallada: No idea, I must have pasted this here by mistake. I'll remove it. Thanks for catching it.

srpearce:
@thallada and @katymyw: I've added a PR for the event documentation that I could use some eyes on: https://github.com/edx/edx-documentation/pull/1562

thallada:
I'm going to merge this. This is the plan for releasing this work: 

1. Mark a new 1.0.0 release of edx-proctoring.
2. Update platform PR to point to it
3. Merge platform PR and release to stage & prod. This will not override any grades, only log messages.
4. Analyze the logs to see if the code is behaving like we expect.
5. Merge another platform PR that adds back in the grade override code. Release to stage & prod. At this point the code will affect new proctored exams unless we disable the overrides on a course with the course waffle flag.

PR #372, opened 2017-08-14
add report generation endpoint + tests

arizzitano:
Adds an endpoint `get_exam_violation_report` to generate an exam violation report.

Details of the product request are in https://openedx.atlassian.net/browse/EDUCATOR-411.

Sandbox available at https://dahlia.sandbox.edx.org.

See also https://github.com/edx/edx-platform/pull/15806.

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -37,6 +37,7 @@ from edx_proctoring.models import (
     ProctoredExamStudentAttempt,
     ProctoredExamStudentAttemptStatus,
     ProctoredExamReviewPolicy,
+    ProctoredExamSoftwareSecureReview,
 )
 from edx_proctoring.serializers import (
     ProctoredExamSerializer,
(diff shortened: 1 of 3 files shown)


edx_proctoring/api.py
```diff
@@ -1947,3 +1948,40 @@ def get_student_view(user_id, course_id, content_id,
...
+            'is_sample_attempt': attempt['is_sample_attempt'],
+            'started_at': attempt['started_at'],
+            'completed_at': attempt['completed_at'],
+            'status': attempt['status'],
+        } for attempt in get_all_exam_attempts(course_id)
+    }
+
+    comments = ProctoredExamSoftwareSecureComment.objects.filter(
+        review__exam__course_id=course_id,
+        review__exam__is_practice_exam=False
```
efischer19: I don't think it will ever happen in practice, but to be clear - are there any scenarios where the user would want to have the option of including practice exams? If so, we could build a defaults-to-False `include_practice_exams` parameter into this method.

arizzitano: Good point -- just added it.

edx_proctoring/api.py
```diff
@@ -1947,3 +1948,46 @@ def get_student_view(user_id, course_id, content_id,
...
+            'completed_at': attempt['completed_at'],
+            'status': attempt['status'],
+            'review_status': None
+        } for attempt in get_all_exam_attempts(course_id)
+    }
+
+    reviews = ProctoredExamSoftwareSecureReview.objects.select_related().filter(
+        exam__course_id=course_id,
+        exam__is_practice_exam=include_practice_exams
+    )
```
arizzitano: ^this was the main thing I changed. I had to fetch the full list of reviews as well, since if we just fetched comments, reviews with no comments wouldn't get their review status set. I think select_related joins the related comments, so we aren't making additional queries for each review's comments.

arizzitano:
@efischer19 I made a few changes since you last looked -- could I trouble you for one more once-over?

edx_proctoring/api.py
efischer19: I *think* we want to specify `proctoredexamsoftwaresecurecomment_set` inside the `select_related()` to avoid fetching more data than we need, but I'm not 100% sure about that. I'll dig a little here.

efischer19: Resulting SQL:

```
SELECT
  `proctoring_proctoredexamsoftwaresecurereview`.`id`,
  `proctoring_proctoredexamsoftwaresecurereview`.`created`,
  `proctoring_proctoredexamsoftwaresecurereview`.`modified`,
  `proctoring_proctoredexamsoftwaresecurereview`.`attempt_code`,
  `proctoring_proctoredexamsoftwaresecurereview`.`review_status`,
  `proctoring_proctoredexamsoftwaresecurereview`.`raw_data`,
  `proctoring_proctoredexamsoftwaresecurereview`.`video_url`,
  `proctoring_proctoredexamsoftwaresecurereview`.`reviewed_by_id`,
  `proctoring_proctoredexamsoftwaresecurereview`.`student_id`,
  `proctoring_proctoredexamsoftwaresecurereview`.`exam_id`
FROM
  `proctoring_proctoredexamsoftwaresecurereview`
INNER JOIN
  `proctoring_proctoredexam`
  ON (
    `proctoring_proctoredexamsoftwaresecurereview`.`exam_id` = `proctoring_proctoredexam`.`id`
)
WHERE(
  `proctoring_proctoredexam`.`course_id` = 'course-v1:edX+DemoX+Demo_Course'
  AND
    `proctoring_proctoredexam`.`is_practice_exam` = 0
)
```

PR #373, opened 2017-08-16
 EDUCATOR-1104 Fix 500 error in course page.

attiyaIshaque:
### Description
This PR fixes that when the user is not enrolled in the course and credit state is None, it wouldn't raise the 500 error.

### How to Test?

**Production** 
- https://courses.edx.org/courses/course-v1:ASUx+SOC101x+2174C/course/

**Sandbox**
- https://proctoring-app.sandbox.edx.org/courses/course-v1:edx+PC101+2017/course/

### Tests
 - [x] Unit Test

### Reviewers
- [x] @afzaledx 
- [x] @noraiz-anwar 

### Post-review

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1235,7 +1235,7 @@ def _check_eligibility_of_enrollment_mode(credit_state):
     # Also make an exception for the honor students to take the "practice exam" as a proctored exam.
     # For the rest of the enrollment modes, None is returned which shows the exam content
     # to the student rather than the proctoring prompt.
-    return credit_state['enrollment_mode'] == 'verified'
+    return credit_state and credit_state['enrollment_mode'] == 'verified'
 
 
 def _get_ordered_prerequisites(prerequisites_statuses, filter_out_namespaces=None):
--- a/edx_proctoring/tests/test_api.py
+++ b/edx_proctoring/tests/test_api.py
@@ -1648,3 +1648,18 @@ class ProctoredExamApiTests(ProctoredExamTestCase):
         self.assertEqual(len(results['failed_prerequisites']), expected_len_failed_prerequisites)
         self.assertEqual(len(results['pending_prerequisites']), expected_len_pending_prerequisites)
         self.assertEqual(len(results['declined_prerequisites']), expected_len_declined_prerequisites)
+
+    def test_summary_without_credit_state(self):
+        """
+        Test that attempt status summary is None for users who are not enrolled.
+        """
+        exam_id = self._create_exam_with_due_time()
+        set_runtime_service('credit', MockCreditServiceNone())
+
+        timed_exam = get_exam_by_id(exam_id)
+        summary = get_attempt_status_summary(
+            self.user.id,
+            timed_exam['course_id'],
+            timed_exam['content_id']
+        )
+        self.assertIsNone(summary)


edx_proctoring/api.py
```diff
@@ -1235,7 +1235,7 @@ def _check_eligibility_of_enrollment_mode(credit_state):
     # Also make an exception for the honor students to take the "practice exam" as a proctored exam.
     # For the rest of the enrollment modes, None is returned which shows the exam content
     # to the student rather than the proctoring prompt.
-    return credit_state['enrollment_mode'] == 'verified'
+    return credit_state and credit_state['enrollment_mode'] == 'verified'
```
noraiz-anwar: We can enhance readability of this by resolving `credit_state and credit_state['enrollment_mode'] == 'verified'` into a separate variable and then return it.
Something like:
``` python 
my_variable = credit_state and credit_state['enrollment_mode'] == 'verified'
return my_variable
```

edx_proctoring/tests/test_api.py
```diff
@@ -1648,3 +1648,19 @@ def test_declined_prerequisites(self, content_id,
         self.assertEqual(len(results['failed_prerequisites']), expected_len_failed_prerequisites)
         self.assertEqual(len(results['pending_prerequisites']), expected_len_pending_prerequisites)
         self.assertEqual(len(results['declined_prerequisites']), expected_len_declined_prerequisites)
+
+    def test_summary_without_credit_state(self):
+        """
+        Test that status summary is None for unenrolled user
+        that does not have a corresponding credit state.
+        """
```
noraiz-anwar: Sorry I didn't get it. What summary we are discussing here? and corresponding credit state?

noraiz-anwar:
What are steps to observe your fix on sandbox?

edx_proctoring/api.py
attiyaIshaque: To enhance readability its good but when `credit_state` is `None` its check `credit_state['enrollment_mode'] `also. and it raised this error
```
'NoneType' object has no attribute '__getitem__'
```
but in this case, if `credit_state` is `None` its returns in this point `return credit_state`
Make sense?

edx_proctoring/tests/test_api.py
attiyaIshaque: Can you please see this method [get_attempt_status_summary](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/api.py#L1452)? You will better understand.

attiyaIshaque:
@noraiz-anwar Just observe course home page would not raise server error.

edx_proctoring/api.py
noraiz-anwar: If `credit_state` is `None` it will not check `credit_state['enrollment_mode']` in 
``` python 
my_variable = credit_state and credit_state['enrollment_mode'] == 'verified'
```

edx_proctoring/tests/test_api.py
noraiz-anwar: We can reword this doc string and specify that we are going to check **attempt status summary**. Also we can specify about corresponding credit state, is this state related to enrollment? If yes, then we can simply say 'test that attempt status summary is None for users who are not enrolled'.

edx_proctoring/api.py
afzaledx: Will the credit_state check be more appropriate here: https://github.com/edx/edx-proctoring/blob/2bee14371e8024e9bca0f7460c0df8538a147e59/edx_proctoring/api.py#L1494

noraiz-anwar: @afzaledx thanks for sharing this. 
@attiyaIshaque You can refactor `_check_eligibility_of_enrollment_mode` before using it. credit_state can be checked in this method before doing `credit_state['enrollment_mode'] == 'verified'`.

attiyaIshaque: @afzaledx Its better to check credit_state within the method `_check_eligibility_of_enrollment_mode`

edx_proctoring/tests/test_api.py
attiyaIshaque: @noraiz-anwar i will change the doc string :) thanks

edx_proctoring/api.py
```diff
@@ -1235,7 +1235,9 @@ def _check_eligibility_of_enrollment_mode(credit_state):
     # Also make an exception for the honor students to take the "practice exam" as a proctored exam.
     # For the rest of the enrollment modes, None is returned which shows the exam content
     # to the student rather than the proctoring prompt.
-    return credit_state['enrollment_mode'] == 'verified'
+    if credit_state:
+        credit_state['enrollment_mode'] == 'verified'
+    return credit_state
```
noraiz-anwar: I think it is better to explicitly return `False` here. Caller of this function would be expecting a boolean. Some caller of this function may use this function like
``` python 
if _check_eligibility_of_enrollment_mode == True/False
```

attiyaIshaque:
@noraiz-anwar I am sticking with my this code 
```
 return credit_state and credit_state['enrollment_mode'] == 'verified'
```
Now i am merging it.

PR #375, opened 2017-08-30
EDUCATOR-1256 Fix 500 error when user is not enrolled.

attiyaIshaque:
### Description
This PR fixes that when the user is not enrolled in the course and credit state is None, it wouldn't raise the 500 error.

### Tests
 - [x] Unit Test

### Reviewers
- [x] @afzaledx 
- [x] @Rabia23 

### Post-review

### DIFF
--- a/edx_proctoring/tests/test_student_view.py
+++ b/edx_proctoring/tests/test_student_view.py
@@ -29,7 +29,7 @@ from edx_proctoring.models import (
 )
 from edx_proctoring.runtime import set_runtime_service
 
-from .test_services import MockCreditServiceWithCourseEndDate
+from .test_services import MockCreditServiceWithCourseEndDate, MockCreditServiceNone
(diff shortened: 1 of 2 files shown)


edx_proctoring/tests/test_student_view.py
```diff
@@ -388,6 +388,24 @@ def test_get_disabled_student_view(self):
...
+            course_id=self.course_id,
+            content_id=self.content_id,
+            context={
+                'is_proctored': True,
+                'display_name': self.exam_name,
+                'default_time_limit_mins': 90
+            },
+            user_role='student'
+        )
+        self.assertIsNotNone(rendered_response)
```
awaisdar001: We should check that after the fix `get_student_view` is not raising 500.

attiyaIshaque: Sure @awaisdar001 I will change it.

PR #380, opened 2017-09-12
Fix Travis builds

efischer19:
Taken from https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Using-Gulp

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,12 +7,15 @@
   "devDependencies": {
     "gulp": "^3.9.0",
     "gulp-karma": "0.0.1",
+    "jasmine-core": "^2.8.0",
     "karma": "^0.13.0",
     "karma-chrome-launcher": "^0.2.0",
-    "karma-coverage": "latest",
-    "karma-firefox-launcher": "latest",
+    "karma-coverage": "^1.1.1",
+    "karma-phantomjs-launcher": "^1.0.4",
     "karma-jasmine": "^0.3.6",
     "karma-jasmine-jquery": "0.1.1",
-    "karma-sinon": "latest"
+    "karma-sinon": "^1.0.5",
+    "phantomjs-prebuilt": "^2.1.14",
+    "sinon": "^3.2.1"
   }
 }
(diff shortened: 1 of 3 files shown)


efischer19:
FYI @awaisdar001 @thallada @dsjen 

AFAICT, the big issue was some sort of change to the travis base that made `gulp` not a command that exists. I got that sorted, and have just been trying to get Travis to run from there. I've added 2 dependencies to package.json, but there's still an error and I'm not sure why.

Further, the [last successful run of these tests](https://travis-ci.org/edx/edx-proctoring/builds/270352234) has a line `Firefox 31.0.0 (Linux 0.0.0): Executed 0 of 15 SUCCESS (0 secs / 0 secs)`, which would seem to indicate that tests didn't run. I think they are though - as there's coverage data afterward.

Right now we're in a state where I have tests passing locally, but not on travis, and no obvious error messages to follow. I'll keep looking tomorrow.

dsjen:
Not that I know why the current travis build is timing out after the
updates, but it might be something about firefox. Maybe you can swap it out
for phantomsjs? Here's how it's set up in Insights,
https://github.com/edx/edx-analytics-dashboard/search?utf8=%E2%9C%93&q=phantom&type=
.

package.json
```diff
@@ -7,12 +7,15 @@
   "devDependencies": {
     "gulp": "^3.9.0",
     "gulp-karma": "0.0.1",
+    "jasmine-core": "^2.8.0",
     "karma": "^0.13.0",
     "karma-chrome-launcher": "^0.2.0",
     "karma-coverage": "latest",
-    "karma-firefox-launcher": "latest",
+    "karma-phantomjs-launcher": "latest",
```
thallada: Putting "latest" here and for the package below worries me because a breaking change could be released for these packages and break our builds. Using the carrot will upgrade everything except new major versions, which is what I think we want.

efischer19: Roger, I'll update before merging

PR #383, opened 2017-10-03
Add logging to find out the cause of error.

attiyaIshaque:
# [EDUCATOR-1441](https://openedx.atlassian.net/browse/EDUCATOR-1441)
### Description:
I have added the logs to find out the root cause of problem because currently, we don't have enough information that in which courses `ProctoredExamNotFoundException` exception is raises

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -313,6 +313,10 @@ def get_exam_by_content_id(course_id, content_id):
     """
     proctored_exam = ProctoredExam.get_exam_by_content_id(course_id, content_id)
     if proctored_exam is None:
+        log.exception(
+            'Cannot find the proctored exams in this course %s with content_id: %s',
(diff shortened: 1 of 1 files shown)


edx_proctoring/api.py
```diff
@@ -313,6 +313,12 @@ def get_exam_by_content_id(course_id, content_id):
     """
     proctored_exam = ProctoredExam.get_exam_by_content_id(course_id, content_id)
     if proctored_exam is None:
+        log.exception(
```
awaisdar001: Please: https://openedx.atlassian.net/wiki/spaces/SUST/pages/125206530/Escalation+Coding+Practices

attiyaIshaque: Thanks @awaisdar001. I have change it according to the doc link.

PR #400, opened 2018-01-04
invalidate certificate when marked proctored exam as suspicious

Rabia23:
# [Learners who drop below passing threshold due to Suspicious proctored exams should not receive certificates - EDUCATOR-2016](https://openedx.atlassian.net/browse/EDUCATOR-2016)

### Description
In the past, because of proctoring, a learners have temporarily passing score and certificate, and later have their score lowered due to being flagged as suspicious. So, they have `certificate delivered = Y` in the grade report, and  `Download Now` link when viewed in the course tools. 

Now the learners will not be able to receive certificates after lowering their score due to "suspicious" flag and grade report will have `certificate delivered = N`.

### Testing
- [x] Unit test

### Reviewers
- [x] Code review: @noraiz-anwar 
- [x] Code review: @thallada 

### Post-review

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -936,6 +936,22 @@ def update_attempt_status(exam_id, user_id, to_status,
                 earned_graded=REJECTED_GRADE_OVERRIDE_EARNED
             )
 
+            certificates_service = get_runtime_service('certificates')
+
+            log.info(
+                'Invalidating certificate for user_id {user_id} in course {course_id} whose '
+                'grade dropped below passing threshold due to suspicious proctored exam'.format(
+                    user_id=exam_attempt_obj.user_id,
+                    course_id=exam['course_id']
+                )
+            )
+
+            # invalidate certificate after overriding subsection grade
+            certificates_service.invalidate_certificate(
+                user_id=exam_attempt_obj.user_id,
+                course_key_or_id=exam['course_id']
+            )
+
     if (to_status == ProctoredExamStudentAttemptStatus.verified and
             ProctoredExamStudentAttemptStatus.needs_grade_override(from_status)):
         grades_service = get_runtime_service('grades')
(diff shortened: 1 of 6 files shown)


edx_proctoring/tests/test_services.py
```diff
@@ -239,3 +259,23 @@ def undo_override_subsection_grade(self, user_id, course_key_or_id, usage_key_or
...
+        self.generated_certificate = {}
+
+    def invalidate_certificate(self, user_id, course_key_or_id):
+        """
+        Get the generated certificate for key (user_id + course_key) and invalidate certificate
+        when dropped below passing threshold due to suspicious proctored exam
+        """
+        key = str(user_id) + str(course_key_or_id)
+        self.generated_certificate[key] = MockGeneratedCertificate()
+        self.generated_certificate[key].mock_invalidate()
```
noraiz-anwar: `mock_invalidate` method simply override all  fields of certificate.  I didn't get how we are testing certificate invalidation here. Lets meet in person and discuss this.

thallada: This is just a mock service. We still need a test in [`test_api.py`](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/tests/test_api.py) checking that when a grade is overridden this invalidate_certificate function is called.

edx_proctoring/api.py
```diff
@@ -936,6 +936,14 @@ def update_attempt_status(exam_id, user_id, to_status,
...
             )
 
+            certificates_service = get_runtime_service('certificates')
+
+            # invalidate certificate after overriding subsection grade
+            certificates_service.invalidate_certificate(
+                user_id=exam_attempt_obj.user_id,
+                course_key_or_id=exam['course_id']
+            )
+
```
thallada: I think this would be a good place to add a log with the message similar to what you had in your other PR: "Invalidating certificate for user %d in course %s whose grade dropped below passing threshold due to suspicious proctored exam"

Rabia23: done

Rabia23:
@thallada / @noraiz-anwar I have addressed your comments. Please have a look at it again.

PR #407, opened 2018-01-22
Update edx-lint dependency to v0.5.5

estute:


jzoldak:
This removes the pylint version dependency incompatibilities and get master back to green.

jzoldak:
This removes the pylint version dependency incompatibilities and gets master back to green.

PR #409, opened 2018-01-31
handle grading service failure

noraiz-anwar:
Continue deleting special exam attempt if grade override undo operation is failed.

[undo_override_subsection_grade](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/services.py#L98) is failed if related PersistentSubsectionGrade entry does not exist.
PersistentSubsectionGrade does not exist in a case where a user started/submitted special exam but did not submit any problem.

FYI: @awaisdar001 / @efischer19

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1120,11 +1120,14 @@ def remove_exam_attempt(attempt_id, requesting_user):
         instructor_service.delete_student_attempt(username, course_id, content_id, requesting_user=requesting_user)
     if grades_service:
         # EDUCATOR-2141: Also remove any grade overrides that may exist
-        grades_service.undo_override_subsection_grade(
-            user_id=user_id,
-            course_key_or_id=course_id,
-            usage_key_or_id=content_id,
-        )
+        try:
+            grades_service.undo_override_subsection_grade(
+                user_id=user_id,
+                course_key_or_id=course_id,
+                usage_key_or_id=content_id,
+            )
+        except Exception as ex:  # pylint: disable=broad-except
+            log.exception(ex)
 
     # see if the status transition this changes credit requirement status
     if ProctoredExamStudentAttemptStatus.needs_credit_status_update(to_status):


efischer19:
I'm not a huge fan of the "broad exception, log it and carry on" approach to this problem - could we instead clean up the grade service method to not throw an exception in these cases? Something like this: https://github.com/edx/edx-platform/compare/efischer/proctoring_idea

noraiz-anwar:
@eric Thanks for looking at it. 
I am not sure if not throwing exception in such cases is ok with respect to grading implementation as it is explicitly mentioned in [get_subsection_grade_override](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/services.py#L39) that an exception is raised if PersistentSubsectionGrade does not exist.
At one other place in [code(recalculation of subsection grade implementation)](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/tasks.py#L226) it is expected that grade override returned by `get_subsection_grade_override` would be None only if score was deleted.
If it makes sense to not throw an exception then definitely your idea is better approach to this problem. What are your thoughts on this?

noraiz-anwar:
@efischer19 Can we do something like https://github.com/edx/edx-platform/compare/noraiz/EDUCATOR-2228? This way we do not need to change `get_subsection_grade_override` implementation. 
Will do some refactoring and doc string updates if overall approach makes sense.
Thanks.

efischer19:
Oh good catch, that would be problematic. I'm still not a fan of having that exception cross the boundary; it ought to remain contained in GradesService. Could we try something like this? (updated code, same link): https://github.com/edx/edx-platform/compare/efischer/proctoring_idea

efischer19:
Ha, great minds think alike - your suggestion is pretty much the same idea I had.

👍 Carry on!

noraiz-anwar:
Thanks @efischer19 for your time. Really appreciate it. 👍

noraiz-anwar:
Closing this PR. Will be filing another PR in edx-platform.

PR #412, opened 2018-02-20
Proctored exam allowed time calculation clean up

noraiz-anwar:
### Description
This PR address the following discrepancy in calculation and usage of allowed time for a proctored exam.
- Allowed time is calculated and stored at time of entrance into proctored exam.
- When a learner starts exam, starting time is used as a benchmark to provide learner with previously calculated time.
- At time of starting exam, learner is nearer to deadline from when allowed time was calculated (at time of entrance into exam).
- Consider a scenario where learner is provided with time past deadline. (This is possible by following above given sequence)
- Deadline is reached but provided time is still running.
- After deadline, submission is not possible as no submit button will be shown and timeout will not be considered a submit.

This is illustrated by following diagram:
proctoring - page 1 2

This PR moves allowed time calculation to 'Exam Started' step i.e time will be calculated and stored when learner starts exam. 

### Reviewers
- [ ] Code review: @efischer19 
- [x] Code review: @awaisdar001 

FYI: @sstack22 

### Post-review

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -553,25 +553,14 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
 
             raise StudentExamAttemptAlreadyExistsException(err_msg)
 
-    allowed_time_limit_mins = exam['time_limit_mins']
-
-    # add in the allowed additional time
-    allowance_extra_mins = ProctoredExamStudentAllowance.get_additional_time_granted(exam_id, user_id)
-    if allowance_extra_mins:
-        allowed_time_limit_mins += allowance_extra_mins
-
-    allowed_time_limit_mins, is_exam_past_due_date = _calculate_allowed_mins(
-        exam['due_date'],
-        allowed_time_limit_mins
-    )
-
+    allowed_time_limit_mins = _calculate_allowed_mins(exam, user_id)
     attempt_code = unicode(uuid.uuid4()).upper()
 
     external_id = None
     review_policy = ProctoredExamReviewPolicy.get_review_policy_for_exam(exam_id)
     review_policy_exception = ProctoredExamStudentAllowance.get_review_policy_exception(exam_id, user_id)
 
-    if not is_exam_past_due_date and taking_as_proctored:
+    if not has_due_date_passed(exam['due_date']) and taking_as_proctored:
         scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
         callback_url = '{scheme}://{hostname}{path}'.format(
             scheme=scheme,
@@ -626,7 +615,6 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
         exam_id,
         user_id,
         '',  # student name is TBD
-        allowed_time_limit_mins,
         attempt_code,
         taking_as_proctored,
         exam['is_practice_exam'],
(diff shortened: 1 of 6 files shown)


edx_proctoring/api.py
```diff
@@ -553,25 +553,14 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
-    allowance_extra_mins = ProctoredExamStudentAllowance.get_additional_time_granted(exam_id, user_id)
-    if allowance_extra_mins:
-        allowed_time_limit_mins += allowance_extra_mins
-
-    allowed_time_limit_mins, is_exam_past_due_date = _calculate_allowed_mins(
-        exam['due_date'],
-        allowed_time_limit_mins
-    )
-
+    allowed_time_limit_mins = _calculate_allowed_mins(exam_id, user_id)
```
awaisdar001: can we pass `exam` here to avoid fetching it from db again? 

e.g. to avoid `get_exam_by_id(exam_id)`

awaisdar001: We are passing `allowed_time_limit_mins ` in context at L#585, will that start the timer for the learner?

noraiz-anwar: `allowed_time_limit_mins ` passed in context is for third party software. We pass this time at time of registering exam with proctoring software.

noraiz-anwar: @efischer19 do you have any idea why we pass `allowed_time_limit_mins` to proctoring software while registering exam attempt?

edx_proctoring/migrations/0007_auto_20180220_1409.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
efischer19: This isn't a blocker if it's too much trouble, but could you attempt to re-create these migrations as a single file, with a descriptive name?

How-to:
1. unapply migrations locally (migrate `edx_proctoring` back to 0005)
2. delete these 2 migration files
3. re-run `makemigrations`, but use the `-n NAME` flag to give the operations a useful name (https://docs.djangoproject.com/en/1.11/ref/django-admin/#django-admin-makemigrations)

noraiz-anwar:
Thank you @efischer19. I just updated my PR according to your suggestions.

noraiz-anwar:
@efischer19 Can you please help me finding out why we [send 'allowed time'](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/api.py#L596) to third party while registering an exam attempt?
We need to know in what cases (if any) third party proctoring software use this sent allowed time.
We are concerned about it because an exam attempt is registered at time of creating exam attempt so proctoring software would have allowed time with respect to exam creation time.

efischer19:
Hey sorry @noraiz-anwar, I didn't get a chance to dig deeply here today. I think we send that along so that the proctoring service can "fail" an exam if the video provided is longer than the "exam time", but I'll have to double-check to confirm that.

I'm going to be digging a bit deeper into proctoring over the next few days; for right now my gut feeling is that we're OK to continue sending the value along. I think the fact that we close off submission after a learner's time is up means this will never really be an issue we see in practice with the proctoring service.

noraiz-anwar:
@efischer19 No problem.
Thank you! I agree to continue sending "exam time" value along.

PR #420, opened 2018-03-07
Updated travis validate translations

Ayub-Khan:
Updated travis to use the i18n-tools validate
to validate translations in pullrequests.
LEARNER-2556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
 before_script:
 - npm install -g gulp-cli
 script:
+- make validate_translations
 - make test-all
 - make test-js
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,12 @@ pull_translations: ## pull translations from Transifex
 push_translations: ## push source translation files (.po) from Transifex
 	tx push -s
 
+validate_translations: ## Test translation files
+	cd edx_proctoring && i18n_tool validate -v
+
 dummy_translations: ## generate dummy translation (.po) files
 	cd edx_proctoring && i18n_tool dummy
 
 build_dummy_translations: extract_translations dummy_translations compile_translations ## generate and compile dummy translation files
 
-validate_translations: build_dummy_translations detect_changed_source_translations ## validate translations
+check_translations_up_to_date: build_dummy_translations detect_changed_source_translations ## validate translations
--- a/requirements/dev.in
+++ b/requirements/dev.in
@@ -2,7 +2,7 @@
 
 diff-cover                # Changeset diff test coverage
 edx_lint==0.5.5           # Python code linting rules
-edx-i18n-tools            # For i18n_tool dummy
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -69,6 +69,7 @@ test-js:
 test-all: ## run tests on every supported Python/Django combination
 	tox -e quality
 	tox
+	cd edx_proctoring && i18n_tool validate
```
amangano-edx: Why do we do this in the `Makefile` instead of `.travis.yml` like we're doing in the other repos?

amangano-edx:
Please validate that this works by creating a test branch, committing a bad translation to it, and confirming that the build fails.

amangano-edx:
Also, please confirm that the correct version of edx-i18n-tools is installed (>= 0.4.0) in the travis environment.

Makefile
Ayub-Khan: @amangano-edx  i have no idea. should i update this.

amangano-edx: I would make this a separate make target (`make extract_translations`) and then run it in the script component of travis.yml, since that's how it's run in the other repos.

Ayub-Khan:
All the translation issues were fixed in https://github.com/edx/edx-proctoring/pull/425.

Makefile
```diff
@@ -91,6 +91,9 @@ pull_translations: ## pull translations from Transifex
 push_translations: ## push source translation files (.po) from Transifex
 	tx push -s
 
+test_translations: ## Test translation files
+	cd edx_proctoring && i18n_tool validate
```
amangano-edx: Why not include this in the existing `validate_translations` target?

Ayub-Khan: validate_translations uses other make targets too and using it will enforce other checks too. thats why i avoided it as its not being used for testing before.

.travis.yml
```diff
@@ -16,6 +16,7 @@ before_script:
 script:
 - make test-all
 - make test-js
+- make validate_translations
```
amangano-edx: Nit: It might be better to run this on line 17, before test-all and test-js. I think it would be better to find out about translation issues up front, rather than having to wait for all the tests to finish.

Makefile
```diff
@@ -91,9 +91,12 @@ pull_translations: ## pull translations from Transifex
 push_translations: ## push source translation files (.po) from Transifex
 	tx push -s
 
+validate_translations: ## Test translation files
+	cd edx_proctoring && i18n_tool validate -v
```
amangano-edx: What does that -v option do? I don't see that documented here: https://github.com/edx/i18n-tools/blob/master/i18n/validate.py#L208

Ayub-Khan: @amangano-edx  it is from parent class https://github.com/edx/i18n-tools/blob/master/i18n/__init__.py#L25

PR #425, opened 2018-03-19
Fix translations

Ayub-Khan:
-updated header and removed existing translations
-updated to pull atleast 1% translated languages
-new translations

fyi @fysheets


Ayub-Khan:
@efischer19 I noticed Codecov project tests always fail. is it ok to merge even if they are failing ?

efischer19:
Yeah, go ahead and merge this

edx_proctoring/locale/ar/LC_MESSAGES/django.po
```diff
@@ -1,15 +1,23 @@
...
+# Nabeel El-Dughailib <nabeel@qordoba.com>, 2016
+# may <may@qordoba.com>, 2016
+# shefaa abu jabel <shefaa.aj@gmail.com>, 2016
+# Najwan Al Rousan <najwanrousan@gmail.com>, 2016
+# Omar Al-Ithawi <oithawi@qrf.org>, 2016
+# Ahmed Jazzar <ajazzar@qrf.org>, 2016
+# Andy Armstrong <andya@edx.org>, 2017
+msgid ""
+msgstr ""
+"Project-Id-Version: edx-proctoring\n"
```
amangano-edx: Curiously, this is set to `0.1a` in https://github.com/edx/edx-platform/blob/master/conf/locale/en/LC_MESSAGES/djangojs.po#L27, but it's set to `edx-platform` in https://github.com/edx/edx-platform/blob/master/conf/locale/es_419/LC_MESSAGES/djangojs.po#L141

I'm not sure which one is correct, as the gettext docs are a little unclear about what this should be. It's probably fine to leave what you have, but for reference, it might be useful to have a look at the docs: https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html

PR #429, opened 2018-04-17
Pin pip version to 9.0.3 for Travis/Tox

fysheets:
Pinning pip version because pip 10 was recently released, and tox-battery doesn't support it yet: https://github.com/signalpillar/tox-battery/issues/14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
 before_install:
 - export DISPLAY=:99.0
 - sh -e /etc/init.d/xvfb start
-- pip install --upgrade pip  # pip-tools requires pip >= 6.1
+- pip install --upgrade pip==9.0.3  # pip-tools requires pip >= 6.1 & pin to 9.0.3 until tox-battery upgrades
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ env:
 before_install:
 - export DISPLAY=:99.0
 - sh -e /etc/init.d/xvfb start
-- pip install --upgrade pip  # pip-tools requires pip >= 6.1
+- pip install --upgrade pip==9.0.3  # pip-tools requires pip >= 6.1 & pin to 9.1.3 until tox-battery upgrades
```
jmbowman: It looks like I originally gave you the wrong version number, you fixed it in the code, but left it unchanged in the comment and PR title.  I accept blame for this, but can we make these consistent? 😃 Otherwise looks fine.

fysheets: Haha yes, good catch :) I blame copy/paste/commiting 5 minutes before a meeting.

PR #448, opened 2018-10-03
WIP switch to using stevedore for locating backends

davestgermain:


edx_proctoring/api.py
```diff
@@ -85,6 +86,7 @@ def create_exam(course_id, content_id, exam_name, time_limit_mins, due_date=None
         is_practice_exam=is_practice_exam,
         is_active=is_active,
         hide_after_due=hide_after_due,
+        backend=backend
```
nedbat: nit: use a trailing comma.  (I know you aren't a fan of super-shallow review... There will also be more substantive comments.)

edx_proctoring/apps.py
```diff
@@ -14,3 +18,31 @@ class EdxProctoringConfig(AppConfig):
...
+        if name is None:
+            try:
+                name = config['DEFAULT']
+            except KeyError:
+                raise ImproperlyConfigured("No default proctoring backend set in settings.PROCTORING_BACKEND_PROVIDERS")
+        try:
+            options = options or config[name]
+            return self.backends[name].plugin(**options)
+        except KeyError:
+            raise NotImplementedError("No proctoring backend configured for '{}'.  "\
```
nedbat: Nit: no need for backslash.

edx_proctoring/api.py
```diff
@@ -63,7 +65,8 @@
 
 
 def create_exam(course_id, content_id, exam_name, time_limit_mins, due_date=None,
-                is_proctored=True, is_practice_exam=False, external_id=None, is_active=True, hide_after_due=False):
+                is_proctored=True, is_practice_exam=False, external_id=None, is_active=True, hide_after_due=False,
+                backend=getattr(settings, 'PROCTORING_BACKENDS', {}).get('DEFAULT', None)):
```
nedbat: You've isolated yourself from settings that might not be available, but don't we want to read the settings at the time the function is called? Why are we working so hard to squeeze this into a parameter default?

edx_proctoring/views.py
```diff
@@ -786,5 +803,221 @@ def put(self, request, attempt_id):     # pylint: disable=unused-argument
...
+                             request.data,
+                             backend=provider)
+        except ProctoredBaseException as ex:
+            LOG.exception(ex)
+            return Response(
+                data={
+                    'reason': six.text_type(ex)
+                },
+                status=400
+            )
```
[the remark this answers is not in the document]
davestgermain: Just adapting to the style in the file...
Yes, it's not easy to read.

edx_proctoring/views.py
```diff
@@ -786,5 +803,221 @@ def put(self, request, attempt_id):     # pylint: disable=unused-argument
...
+        attempt_status = (
+            ProctoredExamStudentAttemptStatus.verified
+            if review.review_status in ['pass', 'verified']
+            else (
+                # if we are not allowed to store 'rejected' on this
+                # code path, then put status into 'second_review_required'
+                ProctoredExamStudentAttemptStatus.rejected if allow_rejects else
+                ProctoredExamStudentAttemptStatus.second_review_required
+            )
+        )
```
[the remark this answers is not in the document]
davestgermain: I'm not responsible for writing that. (But I am responsible for looking the other way instead of changing it)

edx_proctoring/backends/tests/test_backend.py
```diff
@@ -114,6 +117,9 @@ def on_review_saved(self, review):
         """
         return super(PassthroughBackendProvider, self).on_review_saved(review)
 
+    def on_exam_saved(self, exam):
+        return super(PassthroughBackendProvider, self).on_exam_saved(exam)
+
```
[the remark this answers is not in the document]
davestgermain: That's for the tests, to verify that the abstract base class raises NotImplementedError for all of the methods.

edx_proctoring/api.py
davestgermain: I moved it into the function.

edx_proctoring/views.py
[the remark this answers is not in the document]
davestgermain: refactored

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+
+SSL is required for all (production environment) requests.
+
+All requests and responses in this API are formatted as JSON objects.
+
+
+Proctoring System configuration endpoint
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    /v1/config/
```
matthugs: can we have these be a sub-entity to `config` called `rules`? I feel like we're using that framing a lot in conversation around this issue, and it's possible there may be non-rule-oriented configs for other providers in the future

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+        "download_url": "https://my.provider.com/download/software",
+        "instructions": [
+            "Step one, download the software",
+            "Next, run the software",
+            ...
+        ]
+    }
+
+The keys in the config object should be machine readable. The values are human readable. PS should respect the HTTP request ``Accept-Language``
+header and translate all human readable configuration options into the requested language, if possible.
```
matthugs: Just wanted to confirm we're confident that Verificient can handle this i18n subtlety for us. Do we need to raise this in a communication with them?

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+
+    /v1/config/
+
+``GET``: returns an object of the available configuration options and metadata for the proctoring service.::
+
+    {
+        "config": {
+            "allow_multiple": "Allow multiple monitors",
+            "allow_notes": "Allow paper notes",
+            "allow_apps": "Allow other applications to be running",
```
matthugs: As written each of these assume that all rule configurations are boolean values. Can we
1) minimally, document the fact that our api assumes all rules can be expressed as booleans, xor
2) complicate this slightly to e.g. `"allow_apps": {"name": "Allow other applications to be running", "type": "boolean" },` or something else which enables more rule configuration types in the future?

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+            ...
+        },
+        "name": "My Proctoring Provider",
+        "download_url": "https://my.provider.com/download/software",
```
matthugs: This verb could probably be made more clear, since it's likely that the url will likely encompass a more general user action than just "download"ing in the future. Would "launch" be a good alternative?

edx_proctoring/management/commands/create_oauth_credentials.py
```diff
@@ -0,0 +1,46 @@
...
+    def create_oauth_credentials(self, name):
+        """
+        Automatically create oauth credentials for REST backends
+        """
+        username = 'proctor_api_%s' % name
+        try:
+            auth_user = get_user_model().objects.create_user(username)
+        except IntegrityError:
+            auth_user = get_user_model().objects.get(username=username)
+        app, created = apps.get_model('oauth2_provider.Application').objects.get_or_create(
```
davestgermain: @nedbat This command is not easily testable because it's looking for a django app which edx_proctoring does not depend on. The command really belongs somewhere else in edx-platform. I floated the idea but didn't get much interest in even having this command. Maybe it should be removed.

docs/backends.rst
davestgermain: Do you mean rename ```config``` to ```rules``` in the JSON response?

docs/backends.rst
davestgermain: I agree that settings may need to be more flexible in the future. I suggest that you take this task on after this PR has been merged. That would involve changing this doc, the proctoring api, and the mockprock server.

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+The PS system should respond with an object containing at least the following fields::
+
+    {
+        "id": "<some opaque id for the attempt>",
+    }
+
+
+    /v1/exam/{exam_id}/attempt/{attempt_id}/
```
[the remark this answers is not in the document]
davestgermain: Yes, but if I unindent it, it won't match the formatting of the other urls. Do you know what a better restructuredtext strategy is for these things?

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
+===================================
+ Proctoring backend implementation
+===================================
+
+Proctoring providers (PS) who wish to integrate with Open edX should implement a `REST API`_ and a thin `Python wrapper`_, as described below.
```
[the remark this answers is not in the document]
davestgermain: Oops. I did have them all as PP, but decided to change it to PS for "proctoring service", to avoid the scatological references everywhere. But then I didn't change the word it stood for! haha.

edx_proctoring/management/commands/create_oauth_credentials.py
davestgermain: I removed it from the repo.

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+        "id": "<some opaque id for the exam>"
+    }
+
+
+Exam attempt endpoint
+^^^^^^^^^^^^^^^^^^^^^
+
+    /v1/exam/{exam_id}/attempt/
+
+``{exam_id}`` and ``{attempt_id}`` are the ids returned by the PS during creation.
```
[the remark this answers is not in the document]
davestgermain: done

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+header and translate all human readable configuration options into the requested language, if possible. 
+
+If a download_url is included in the response, Open edX will redirect learners to the the address before the proctoring session starts. The address will include ``attempt={attempt_id}`` in the query string.
+
+Exam endpoint
+^^^^^^^^^^^^^
+
+    /v1/exam/{exam_id}/
+
+``GET``: returns an object about the exam. If no exam exists, return 404 error.::
```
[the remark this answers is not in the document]
davestgermain: done

edx_proctoring/backends/rest.py
```diff
@@ -0,0 +1,166 @@
...
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+        self.session = OAuthAPIClient(self.base_url, self.client_id, self.client_secret)
+
+    def get_javascript(self):
+        """
+        Returns the backend javascript to embed on each proctoring page
+        """
+        package = self.__class__.__module__.split('.')[0]
+        return pkg_resources.resource_string(package, 'backend.js')
```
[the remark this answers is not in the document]
matthugs: We spoke through 2. in person yesterday; I'm not adding any new concerns from the concern that `server+backend` should probably become `server...       +         ... backend` in some way, based on the differing levels of scrutiny we'll apply to each. I think I captured that in [EDUCATOR-3664](https://openedx.atlassian.net/browse/EDUCATOR-3664).

I'll try to remember to capture 1. in a ticket; waiting on one from @scottrish encompassing JS API clarification work.

docs/backends.rst
matthugs: Can you capture it in a ticket/subtask of something, then, to help ensure it doesn't get lost in the mail?

docs/backends.rst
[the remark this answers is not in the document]
nedbat: Could each URL be a new heading?

docs/backends.rst
```diff
@@ -0,0 +1,211 @@
...
+``POST``: may be used to create the exam on the PS, by sending an object like this::
+
+    {
+        "config": {
+            "allow_notes": false,
+            "allow_multiple": true
+        },
+        "course_id": "myOrgX:Course101",
+        "is_practice": false,
+        "is_proctored": true,
```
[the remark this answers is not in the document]
davestgermain: Actually, I don't think there is a purpose to that. I think it mirrors the edx model, where ``is_proctored = False`` means that it's a Timed exam.

edx_proctoring/backends/software_secure.py
```diff
@@ -360,7 +209,7 @@ def _create_zendesk_ticket(self, review, serialized_exam_object, serialized_atte
         """
```
nedbat: It looks like this pull request removes the one call to _create_zendesk_ticket, while also changing the function?

edx_proctoring/backends/software_secure.py
```diff
@@ -147,186 +151,31 @@ def on_review_callback(self, payload):
...
+        SoftwareSecureReviewStatus.validate(payload['reviewStatus'])
+        review_status = SoftwareSecureReviewStatus.to_standard_status.get(payload['reviewStatus'], None)
+
+        comments = []
+        converted = {
+            'status': review_status,
+            'comments': comments,
+            'payload': payload
         }
+        for comment in payload.get('webCamComments', []) + payload.get('desktopComments', []):
```
nedbat: I see that this will work fine, but it will be clearer if you build `comments` fully before using it in `converted`.

setup.py
```diff
@@ -63,18 +63,28 @@ def get_version(*file_paths):
     ],
```
nedbat: Eventually: update the classifiers (not Alpha, Django 1.11 only, supports 3.6)

edx_proctoring/api.py
```diff
@@ -453,6 +467,14 @@ def get_exam_attempt_by_id(attempt_id):
     return _get_exam_attempt(exam_attempt_obj)
 
 
+def get_exam_attempt_by_external_id(external_id):
+    """
+    Return an existing exam attempt for the given student
```
nedbat: There's no student given.

It would be great if we could update the docstrings to be clear about what is happening. "Attempt" is used to mean (I think) three different things: our attempt id, their attempt id, and our attempt object.

edx_proctoring/backends/software_secure.py
```diff
@@ -129,13 +120,26 @@ def get_software_download_url(self):
...
+    def on_review_callback(self, attempt, payload):
         """
         Called when the reviewing 3rd party service posts back the results
 
         Documentation on the data format can be found from SoftwareSecure's
         documentation named "Reviewer Data Transfer"
         """
+        received_id = payload['examMetaData']['ssiRecordLocator'].lower()
+        if attempt['external_id'].lower() != received_id and not\
+                settings.PROCTORING_SETTINGS.get('ALLOW_CALLBACK_SIMULATION', False):
```
nedbat: I found the original form of this test easier to understand (fewer negations, and split out into a variable):
```
        match = (
            attempt_obj.external_id.lower() == external_id.lower() or
            settings.PROCTORING_SETTINGS.get('ALLOW_CALLBACK_SIMULATION', False)
        )
        if not match:
```

edx_proctoring/backends/software_secure.py
davestgermain: I've moved the zendesk ticket creation to the review endpoint.

edx_proctoring/views.py
```diff
@@ -852,15 +852,16 @@ def make_review(self, attempt, data, backend=None):
...
-        # updating attempt status will trigger workflow
-        # (i.e. updating credit eligibility table)
-
-        update_attempt_status(
-            attempt['proctored_exam']['id'],
-            attempt['user']['id'],
-            attempt_status,
-            raise_if_not_found=False
-        )
+        if not attempt.get('is_archived', False):
```
davestgermain: @nedbat it turns out that archived attempts are used a lot. I restored the check, to prevent status changes on archived attempts.
Every save to the attempt table duplicates the data in the attempt history table. We should revisit this to figure out whether it's necessary to keep all of this duplicated data around.

edx_proctoring/backends/software_secure.py
```diff
@@ -147,186 +151,31 @@ def on_review_callback(self, payload):
...
         """
 
-        (attempt_obj, is_archived_attempt) = locate_attempt_by_attempt_code(review.attempt_code)
-
-        if not attempt_obj:
-            # This should not happen, but it is logged in the help
-            # method
-            return
-
-        if is_archived_attempt:
```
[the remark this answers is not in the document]
davestgermain: restored in https://github.com/edx/edx-proctoring/pull/448/commits/43ef1bca3ae2f5e9d14c0c42f7c257b033f871f1

edx_proctoring/backends/software_secure.py
```diff
@@ -147,186 +151,31 @@ def on_review_callback(self, payload):
...
-    def on_review_saved(self, review, allow_rejects=False):  # pylint: disable=arguments-differ
+    def on_exam_saved(self, exam):
         """
-        called when a review has been save - either through API (on_review_callback) or via Django Admin panel
-        in order to trigger any workflow associated with proctoring review results
+        Called after an exam is saved.
         """
 
-        (attempt_obj, is_archived_attempt) = locate_attempt_by_attempt_code(review.attempt_code)
-
```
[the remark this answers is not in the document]
davestgermain: restored in https://github.com/edx/edx-proctoring/pull/448/commits/43ef1bca3ae2f5e9d14c0c42f7c257b033f871f1

PR #452, opened 2018-11-01
Trying to speed up travis builds

davestgermain:


davestgermain:
@nedbat does this make sense?

nedbat:
I don't understand these changes, or why they make things faster.

PR #455, opened 2018-11-08
Correct .rst errors so README is readable on PyPI

nedbat:
(also, what's with the changelog?)


nedbat:
@davestgermain simple.

PR #456, opened 2018-11-08
Reinstated the button to continue to exam

davestgermain:
[EDUCATOR-3688](https://openedx.atlassian.net/browse/EDUCATOR-3688)


nedbat:
I don't see anything here about continuing.  The start button will now be available even if there are backend instructions?

davestgermain:
@nedbat The current production proctoring flow has a start button on the page that you have to manually click to continue. I previously removed that button and had the page poll for status changes and automatically reload when the exam attempt is ready. This PR reinstates the start button, to make it consistent with the production experience.

PR #457, opened 2018-11-09
Save the review policy with the exam on the backend

davestgermain:
Whenever the ProctoredExam or ProctoredExamReviewPolicy models change, the backend `on_exam_saved` method will be called, passing in the correct serialized exam.

This will simplify the next step, which is to add the exam rules as a JSON blob.

@MichaelRoytman @nedbat

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -16,6 +16,8 @@ import six
 from django.utils.translation import ugettext as _, ugettext_noop
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.dispatch import receiver
+from django.db.models.signals import post_save
 from django.template import loader
 from django.urls import reverse, NoReverseMatch
 from django.core.mail.message import EmailMessage
@@ -108,10 +110,6 @@ def create_exam(course_id, content_id, exam_name, time_limit_mins, due_date=None
     # read back exam so we can emit an event on it
     exam = get_exam_by_id(proctored_exam.id)
     emit_event(exam, 'created')
-    external_id = get_backend_provider(exam).on_exam_saved(exam)
-    if external_id:
-        proctored_exam.external_id = external_id
-        proctored_exam.save()
     return proctored_exam.id
 
 
@@ -209,6 +207,33 @@ def remove_review_policy(exam_id):
     exam_review_policy.delete()
 
 
+@receiver(post_save, sender=ProctoredExamReviewPolicy)
+@receiver(post_save, sender=ProctoredExam)
+def _save_exam_on_backend(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Save the exam to the backend provider when our model changes.
+    It also combines the review policy into the exam when saving to the backend
+    """
+    if sender == ProctoredExam:
+        exam_obj = instance
+        review_policy = ProctoredExamReviewPolicy.get_review_policy_for_exam(instance.id)
+    else:
+        exam_obj = instance.proctored_exam
+        review_policy = instance
+    if exam_obj.is_proctored:
+        exam = ProctoredExamSerializer(exam_obj).data
+        if review_policy:
+            exam['rule_summary'] = review_policy.review_policy
+            # When the rules are defined as boolean options,
+            # save them here
+            exam['rules'] = getattr(review_policy, 'rules', {})
+        backend = get_backend_provider(exam)
+        external_id = backend.on_exam_saved(exam)
+        if external_id and external_id != exam_obj.external_id:
+            exam_obj.external_id = external_id
+            exam_obj.save()
+
+
 def get_review_policy_by_exam_id(exam_id):
     """
     Looks up exam by the Primary Key. Raises exception if not found.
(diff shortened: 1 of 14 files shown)


edx_proctoring/api.py
```diff
@@ -193,6 +191,34 @@ def remove_review_policy(exam_id):
...
+        exam_obj = instance.proctored_exam
+        review_policy = instance
+    if exam_obj.is_proctored:
+        exam = ProctoredExamSerializer(exam_obj).data
+        if review_policy:
+            exam['rule_summary'] = review_policy.review_policy
+            # When the rules are defined as boolean options,
+            # save them here
+            if hasattr(review_policy, 'rules'):  # pragma: no cover
+                exam['rules'] = review_policy.rules  # pragma: no cover
```
nedbat: Why do we think these lines can't be tested?

davestgermain: @nedbat because @MichaelRoytman hasn't finished his PR, which will create the rules field on the review policy model. I assume that he will remove the pragma comment and the hasattr check in his PR.

edx_proctoring/apps.py
```diff
@@ -23,33 +25,41 @@ def get_backend_choices(self):
         Returns an iterator of available backends:
         backend_name, verbose name
         """
-        for extension in self.backends:
-            yield extension.name, getattr(extension.plugin, 'verbose_name', u'Unknown')
+        for name, backend in self.backends.items():
```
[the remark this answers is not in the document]
davestgermain: ```items()``` is compatible with python 2 and 3. On 2, it returns a list, while on 3 it returns an iterator. In this case, the difference is irrelevant.

edx_proctoring/apps.py
```diff
@@ -23,33 +25,41 @@ def get_backend_choices(self):
...
             raise NotImplementedError("No proctoring backend configured for '{}'.  "
-                                      "Available: {} {}".format(name, self.backends.names(), config))
+                                      "Available: {}".format(name, list(self.backends)))
 
     def ready(self):
         """
         Loads the available proctoring backends
         """
-        self.backends = ExtensionManager(namespace='openedx.proctoring')  # pylint: disable=W0201
+        config = getattr(settings, 'PROCTORING_BACKENDS', None)  # pylint: disable=literal-used-as-attribute
```
nedbat: Did pylint complain about this line? You shouldn't need that comment.

edx_proctoring/backends/rest.py
```diff
@@ -142,9 +142,9 @@ def on_exam_saved(self, exam):
         """
         Called after an exam is saved.
         """
-        if 'config' not in exam and self.default_config:
+        if self.default_rules and not exam.get('rules', None):
```
nedbat: The exam should get the default rules if it explicitly defines `rules: {}` ?

edx_proctoring/tests/__init__.py
```diff
@@ -13,15 +12,9 @@ def setup_test_backends():
...
-        Extension('test', 'openedx.proctoring', TestBackendProvider, None),
-        Extension('null', 'openedx.proctoring', NullBackendProvider, None),
-        Extension('mock', 'openedx.proctoring', MockProctoringBackendProvider, None),
-        Extension('software_secure', 'openedx.proctoring', SoftwareSecureBackendProvider, None)
-    ]
-
-    config.backends = ExtensionManager.make_test_instance(extensions)
+    config.backends['test'] = TestBackendProvider()
+    config.backends['null'] = NullBackendProvider()
+    config.backends['mock'] = MockProctoringBackendProvider()
```
nedbat: We don't need to clean these up when the test is done?

davestgermain: There's probably a better way to do this. The function should run one time before the tests run. I could put it a setupClass method somewhere, but the various test suites don't inherit from a single base class.

Do you have a suggestion?

edx_proctoring/backends/rest.py
davestgermain: That's my assumption for the moment. There are always going to be default rules, whether they're defined on the platform level (which eventually percolate to self.default_rules) or whether they're defined on the provider's side (in which case, we don't know exactly what the defaults are). For now I think we can assume that all of the backend defaults will be defined on the platform level explicitly.

edx_proctoring/tests/__init__.py
nedbat: My question was really, is it OK that these things will be in effect for all the tests?  If that's OK, then we can leave this as it is.

Pytest has a bunch of ways to manage these things if you need to clean them up.

edx_proctoring/backends/rest.py
nedbat: Sorry, I was unclear. You're saying there are always default rules, but I'm talking about what is on the exam.  If the exam says `{"rules": {}}`, then it will get the defaults.  The old code wouldn't have applied defaults if the exam had `{"config": {}}`

davestgermain: That's right. I think this is the better approach, to ensure that at least the platform default rules get sent to the backend. We could certainly revisit it once we have a real workflow going.

edx_proctoring/tests/__init__.py
davestgermain: Yes, they apply to all tests. This monkeypatching is needed because stevedore loads the backends, and only if edx_proctoring is installed.

PR #459, opened 2018-11-14
Modernize the requirements files.

nedbat:


nedbat:
BTW: the author should merge, not the approver.

davestgermain:
@nedbat check the travis logs. You may need to use an absolute path to requirements/base.in.

setup.py
```diff
@@ -61,29 +90,12 @@ def get_version(*file_paths):
...
-        "pycryptodomex>=3.4.7",
-        "python-dateutil>=2.1",
-        "requests",
-        "stevedore",
-        "six",
-    ],
-    dependency_links=[
-        "git+https://github.com/edx/event-tracking.git@0.2.2#egg=event-tracking==0.2.2",
-    ],
+    install_requires=load_requirements("requirements/base.in"),
```
davestgermain: For completeness' sake, should there be a ``tests_require``?

davestgermain: ...And then maybe integrate the [pytest runner](https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner) here?

nedbat:
Jenkins is now happy.  I'm not going to add stuff to support "setup.py test".  I don't think people do that these days.

davestgermain:
👍

PR #460, opened 2018-11-14
add rules field to ProctoredExamReviewPolicy and ProctoredExamReviewP…

MichaelRoytman:
…olicyHistory models as JSONField


MichaelRoytman:
Re: https://openedx.atlassian.net/browse/EDUCATOR-3580
Related to: https://github.com/edx/edx-proctoring/pull/457

edx_proctoring/api.py
```diff
@@ -115,58 +115,61 @@ def create_exam(course_id, content_id, exam_name, time_limit_mins, due_date=None
...
+            rules=rules
         )
     )
     log.info(log_msg)
     exam_review_policy = ProctoredExamReviewPolicy.get_review_policy_for_exam(exam_id)
     if exam_review_policy is None:
         raise ProctoredExamReviewPolicyNotFoundException
 
-    if review_policy:
+    if review_policy or rules:
```
MichaelRoytman: TODO: This has the potential to save None as the review policy, which the ProctoredExamReviewPolicy model does not support (i.e. `review_policy = models.TextField()`). I would have to do allow null values. Is this the right approach?

edx_proctoring/models.py
```diff
@@ -256,6 +257,9 @@ class ProctoredExamReviewPolicy(TimeStampedModel):
     # policy that will be passed to reviewers
     review_policy = models.TextField()
 
+    # JSON rules that will be passed to reviewers
+    rules = JSONField(null=True)
```
MichaelRoytman: `null=true` because not all proctoring providers support JSON rules (e.g. RPNow).

edx_proctoring/api.py
davestgermain: How about ``review_policy = models.TextField(default='')``? I don't think that would require a migration.

MichaelRoytman: It did create a migration, actually. Does that change your suggestion?

davestgermain: No, I guess that's ok.

requirements/test.txt
```diff
@@ -30,6 +30,7 @@ future==0.17.1            # via pyjwkest
 httmock==1.2.6
 httpretty==0.9.6
 idna==2.7                 # via requests
+jsonfield==2.0.2
```
davestgermain: @MichaelRoytman I think with @nedbat's change, you don't need to put jsonfield in the test/doc/etc requirements files. Just the base file will do the trick.

nedbat: @MichaelRoytman You should edit just one .in file to add the requirement (base.in).  Then "make upgrade" will update all the .txt files.

requirements/base.txt
```diff
@@ -4,6 +4,9 @@
 #
 #    pip-compile --output-file requirements/base.txt requirements/base.in
 #
+--index-url http://edx.devstack.devpi:3141/root/pypi/+simple/
```
davestgermain: You should remove these lines before checking in the file. It's a devstack artifact.

davestgermain:
👍

PR #461, opened 2018-11-19
EDUCATOR-3703: proctoring js arch proof of concept

matthugs:
https://openedx.atlassian.net/browse/EDUCATOR-3703

ping/echo is as-yet uncalled

This approach involves the edx-proctoring npm package being imported by each provider's npm package. We'll want to get this up quickly so I can publish to `npm` and update the dependencies downstream to pull down from the registry (since I've been developing with them specified by local file paths, which will need to be corrected before we can merge those changes).

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1807,7 +1807,7 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
         student_view_template = 'proctored_exam/ready_to_submit.html'
 
     if student_view_template:
-        context['backend_js'] = provider.get_javascript()
+        context['backend_js_bundle'] = provider.get_javascript()
         template = loader.get_template(student_view_template)
         context.update(_get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_exam=True))
         return template.render(context)
@@ -1955,7 +1955,7 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
         student_view_template = 'proctored_exam/ready_to_submit.html'
 
     if student_view_template:
-        context['backend_js'] = provider.get_javascript()
+        context['backend_js_bundle'] = provider.get_javascript()
         template = loader.get_template(student_view_template)
         context.update(_get_proctored_exam_context(exam, attempt, user_id, course_id))
         return template.render(context)
@@ -2012,7 +2012,7 @@ def get_student_view(user_id, course_id, content_id,
             is_proctored=context.get('is_proctored', False),
             is_practice_exam=context.get('is_practice_exam', False),
             due_date=context.get('due_date', None),
-            hide_after_due=context.get('hide_after_due', None),
+            hide_after_due=context.get('hide_after_due', False),
         )
         exam = get_exam_by_content_id(course_id, content_id)
 
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -5,7 +5,8 @@ docs/backends.rst
 import logging
 import time
 import uuid
-import pkg_resources
+
+from webpack_loader.utils import get_files
 
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus
@@ -72,10 +73,13 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
 
     def get_javascript(self):
         """
-        Returns the backend javascript to embed on each proctoring page
+        Returns the url of the javascript bundle into which the provider's JS will be loaded
         """
         package = self.__class__.__module__.split('.')[0]
-        return pkg_resources.resource_string(package, 'backend.js')
+        bundle_chunks = get_files(package, config="WORKERS")
+        if bundle_chunks:
+            return bundle_chunks[0]["url"]
+        return ''
 
     def get_software_download_url(self):
         """
--- a/edx_proctoring/backends/tests/test_rest.py
+++ b/edx_proctoring/backends/tests/test_rest.py
@@ -6,6 +6,7 @@ import json
 import jwt
 
 import responses
+from mock import patch
 
 from django.test import TestCase
 from django.utils.translation import activate
@@ -216,9 +217,22 @@ class RESTBackendTests(TestCase):
         self.assertEqual(payload, new_payload)
 
     def test_get_javascript(self):
-        # A real backend would return real javascript from backend.js
+        # A real backend would return a real bundle url from webpack
+        # but in this context we'll fail looking up webpack's stats file
         with self.assertRaises(IOError):
             self.provider.get_javascript()
+
+    @patch('edx_proctoring.backends.rest.get_files')
+    def test_get_javascript_bundle(self, get_files_mock):
+        get_files_mock.return_value = [{'name': 'rest', 'url': '/there/it/is'}]
+        javascript_url = self.provider.get_javascript()
+        self.assertEqual(javascript_url, '/there/it/is')
+
+    @patch('edx_proctoring.backends.rest.get_files')
+    def test_get_javascript_empty_bundle(self, get_files_mock):
+        get_files_mock.return_value = []
+        javascript_url = self.provider.get_javascript()
+        self.assertEqual(javascript_url, '')
 
     def test_instructor_url(self):
         user = {
--- a/edx_proctoring/static/index.js
+++ b/edx_proctoring/static/index.js
@@ -1,0 +1,32 @@
+export const handlerWrapper = (Handler) => {
+   let handler = new Handler({});
+
+  self.addEventListener("message", (message) => {
+    switch(message.data.type) {
+      case 'config': {
+        handler = new Handler(message.data.options);
+        break;
+      }
+      case 'startExamAttempt': {
+        if(handler.onStartExamAttempt) {
+          handler.onStartExamAttempt().then(() => self.postMessage({type: 'examAttemptStarted'}))
+        }
+        break;
+      }
+      case 'endExamAttempt': {
+        if(handler.onEndExamAttempt) {
+          handler.onEndExamAttempt().then(() => self.postMessage({type: 'examAttemptEnded'}))
+        }
+        break;
+      }
+      case 'ping': {
+        if(handler.onPing) {
+          handler.onPing().then(() => self.postMessage({type: 'echo'}))
+        }
+        break;
+      }
+    }
+  });
+
+}
+export default handlerWrapper;
--- a/edx_proctoring/static/proctoring/js/exam_action_handler.js
+++ b/edx_proctoring/static/proctoring/js/exam_action_handler.js
@@ -1,0 +1,95 @@
+var edx = edx || {};
+
+(function($) {
+  'use strict';
+
+  var actionToMessageTypesMap = {
+    'submit': {
+      promptEventName: 'endExamAttempt',
+      responseEventName: 'examAttemptEnded'
+    },
+    'start': {
+      promptEventName: 'startExamAttempt',
+      responseEventName: 'examAttemptStarted'
+    }
+  };
+
+  function workerPromiseForEventNames(eventNames) {
+    return function() {
+      var proctoringBackendWorker = new Worker(edx.courseware.proctored_exam.configuredWorkerURL);
+      return new Promise(function(resolve) {
+        var responseHandler = function(e) {
+          if (e.data.type === eventNames.responseEventName) {
+            proctoringBackendWorker.removeEventListener('message', responseHandler);
+            proctoringBackendWorker.terminate();
+            resolve();
+          }
+        };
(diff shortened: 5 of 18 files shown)


edx_proctoring/templates/proctored_exam/ready_to_start.html
```diff
@@ -56,16 +56,33 @@ <h3>
...
+        var proctoringBackendWorker = new Worker(edx.courseware.proctored_exam.configuredWorkerURL);
+        new Promise((resolve) => {
+            var responseHandler = (e) => {
+                if (e.data.type === "examAttemptStarted") {
+                    proctoringBackendWorker.removeEventListener('message', responseHandler);
+                    proctoringBackendWorker.terminate();
+                    resolve();
+                }
+            };
+            proctoringBackendWorker.addEventListener('message', responseHandler);
```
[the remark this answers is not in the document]
matthugs: `static/index.js` currently only contains code that gets bundled into the webworker bundle; there isn't a simple way to get code from there into the webpage's main JS `window` context as of right now.

The more available alternatives are to add another file into https://github.com/edx/edx-platform/blob/6c5ee7209c9c0d62b143a93dc43475806c44ec90/lms/envs/common.py#L1351, which would put our JS on every single LMS page whenever proctoring is enabled. That'd be pretty glaringly not ideal, so I'm being glaring in my hackaround.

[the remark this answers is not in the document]
davestgermain: What about ``proctoring/js/proctored_app.js``? That will be on every proctored page, but not every LMS page, right?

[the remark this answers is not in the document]
matthugs: Nope, every single file in that list is loaded on every page of a course with proctoring enabled: https://github.com/edx/edx-platform/blob/master/lms/templates/courseware/course_navigation.html#L26

Fixing that situation seemed like it'd inflate scope, but we can talk through it synchronously.

edx_proctoring/backends/rest.py
```diff
@@ -59,10 +60,11 @@ def __init__(self, client_id=None, client_secret=None, **kwargs):
 
     def get_javascript(self):
         """
-        Returns the backend javascript to embed on each proctoring page
+        Returns the url of the javascript bundle into which the provider's JS will be loaded
```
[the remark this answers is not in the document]
matthugs: It returns nothing; tested

edx_proctoring/backends/tests/test_rest.py
```diff
@@ -182,6 +182,7 @@ def test_on_review_callback(self):
         self.assertEqual(payload, new_payload)
 
     def test_get_javascript(self):
-        # A real backend would return real javascript from backend.js
+        # A real backend would return a real bundle url from webpack
+        # but in this context we'll fail looking up webpack's stats file
         with self.assertRaises(IOError):
```
davestgermain: If you changed the code to return an empty string if there's no package, then the code will not raise IOError, in which case, this test should fail. Does it fail?

edx_proctoring/templates/proctored_exam/ready_to_start.html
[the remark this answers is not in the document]
davestgermain: There is a footer.html template included on every proctored page. It already has javascript and could be used for these event handlers. Ultimately, I think the js is going to have to be refactored into the static/index.js or another file, because once the proctored exam starts, we're in the regular courseware context, without any templates available from edx-proctoring. We won't be able to create the event listener for the application heartbeat, unless the code is refactored.

edx_proctoring/backends/tests/test_rest.py
matthugs: There are two files loaded when looking up a bundle: first the library looks up the stats file (which in this test case, fails with an IOError). If that were to succeed, then we'd look up the bundle's location from inside the stats file. In the case that the bundle's name isn't found in the stats file, get_files will return an empty list.

I hardened the code and added a test illustrating this, but the test we're talking about here does pass.

edx_proctoring/templates/proctored_exam/ready_to_start.html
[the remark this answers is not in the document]
matthugs: referenced in https://openedx.atlassian.net/browse/EDUCATOR-3690

georgebabey:
We should be using https://github.com/edx/eslint-config-edx and the style defined there. I know this might be a bit controversial (I believe you've mentioned being a fan of prettier) but right now thats our standards laid out in OEP-11

edx_proctoring/static/index.js
```diff
@@ -0,0 +1,32 @@
+export const handlerWrapper = (Handler) => {
+   let handler = new Handler({});
+
+  self.addEventListener("message", (message) => {
+    switch(message.data.type) {
+      case 'config': {
```
georgebabey: Documentation on this event in the backends.rst file would be helpful, even if its just to state we will call your initializer with the options passed into this event.

edx_proctoring/static/index.js
```diff
@@ -0,0 +1,32 @@
...
+        handler = new Handler(message.data.options);
+        break;
+      }
+      case 'startExamAttempt': {
+        if(handler.onStartExamAttempt) {
+          handler.onStartExamAttempt().then(() => self.postMessage({type: 'examAttemptStarted'}))
```
georgebabey: I would pull these type strings out into constants so they can be imported for use here and in the action handler

edx_proctoring/static/index.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+      case 'endExamAttempt': {
+        if(handler.onEndExamAttempt) {
+          handler.onEndExamAttempt().then(() => self.postMessage({type: 'examAttemptEnded'}))
+        }
+        break;
+      }
+      case 'ping': {
+        if(handler.onPing) {
+          handler.onPing().then(() => self.postMessage({type: 'echo'}))
```
georgebabey: Doesn't seem to be used in the action handler at all, but you may want to be consistent and call this type 'ping'

edx_proctoring/static/index.js
georgebabey: I'm also not clear on how this interfaces with the proctoring tool but you may want to accept some data object that can be passed to the action handler. For example would they need some identifier of the exam being taken? Again this is my first time looking at this and I don't think I have the full picture so please ignore if this doesn't make any sense.

edx_proctoring/static/index.js
```diff
@@ -0,0 +1,32 @@
...
+        handler = new Handler(message.data.options);
+        break;
+      }
+      case 'startExamAttempt': {
+        if(handler.onStartExamAttempt) {
+          handler.onStartExamAttempt().then(() => self.postMessage({type: 'examAttemptStarted'}))
+        }
+        break;
+      }
+      case 'endExamAttempt': {
```
georgebabey: Would also pull these message types into a constant to import here and the action handlers `actionToMessageTypesMap`

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
+var edx = edx || {};
+
+(function($) {
```
georgebabey: Using jquery here seems a little heavy. I'd recommend axios for the ajax call and vanilla js (`getDocumentById` or `querySelector` w/ `.dataset`)

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
...
+  'use strict';
+
+  var actionToMessageTypesMap = {
+    'submit': {
```
georgebabey: See above, I'd love to see these magic strings pulled into a shared constant module

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
...
+    },
+    'start': {
+      promptEventName: 'startExamAttempt',
+      responseEventName: 'examAttemptStarted'
+    }
+  };
+
+  function workerPromiseForEventNames(eventNames) {
+    return function() {
+      var proctoringBackendWorker = new Worker(edx.courseware.proctored_exam.configuredWorkerURL);
```
georgebabey: may want to check for window.Worker... although if they've gotten this far on a browser that doesn't support web workers I'd be impressed :)

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
...
+      return new Promise(function(resolve) {
```
georgebabey: `(resolve) =>`

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
...
+        var responseHandler = function(e) {
+          if (e.data.type === eventNames.responseEventName) {
+            proctoringBackendWorker.removeEventListener('message', responseHandler);
+            proctoringBackendWorker.terminate();
+            resolve();
+          }
+        };
+        proctoringBackendWorker.addEventListener('message', responseHandler);
```
georgebabey: Why not use the event name here for the eventListener rather than the generic 'message' - it may make debugging easier in the future and would allow active listeners for multiple event types to be created in the future.

edx_proctoring/static/proctoring/js/exam_action_handler.js
georgebabey: Nevermind, I see further down you check in `shouldUseWorker`

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -0,0 +1,95 @@
...
+    if(shouldUseWorker) {
+      workerPromiseForEventNames(actionToMessageTypesMap[action])()
+        .then(updateExamAttemptStatusPromise(actionUrl, action))
+        .then(reloadPage);
+    } else {
+      updateExamAttemptStatusPromise(actionUrl, action)()
+        .then(reloadPage);
+    }
+  };
+  edx.courseware.proctored_exam.examEndHandler = function() {
```
georgebabey: nit: some weird newlines here (and probably should add one before this function)

edx_proctoring/api.py
```diff
@@ -1804,7 +1804,7 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
         student_view_template = 'proctored_exam/ready_to_submit.html'
 
     if student_view_template:
-        context['backend_js'] = provider.get_javascript()
+        context['backend_js_bundle'] = provider.get_javascript()
```
georgebabey: nit: I find this name a bit confusing as its just the bundle url, not the javascript itself.

edx_proctoring/static/index.js
matthugs: I'm not sure I have a good mechanism for doing that. Are you proposing an es6 module with those constants, to be imported both here and in the JS that runs outside the web worker? None of the latter are es6 modules, so I'm not sure I see how that could be achieved.

edx_proctoring/static/index.js
matthugs: I pushed it off, but the implementation is in progress: https://github.com/edx/edx-proctoring/pull/466

When this was first pitched to me it was described as a "heartbeat" mechanism, and I found the ping/echo language more descriptive when googling around to understand the rationale for implementing the pattern: http://etutorials.org/Programming/Software+architecture+in+practice,+second+edition/Part+Two+Creating+an+Architecture/Chapter+5.+Achieving+Qualities/5.2+Availability+Tactics/

edx_proctoring/static/proctoring/js/exam_action_handler.js
matthugs: this code doesn't run through babel, and I'm a bit anxious about browser compatibility

edx_proctoring/static/proctoring/js/exam_action_handler.js
matthugs: IIUC this is standard webworker boilerplate; `MessageEvent`s are all that are propagated between a page and the web workers they spawn. I don't think I can listen to any other types of events here.

edx_proctoring/static/index.js
matthugs: You're completely correct that this is weird; I only have this in here because of a dev-only need. We won't be able to set up a sandbox server for our testing-only reference implementation without this in here. I didn't document it because I don't actually want vendor implementations to depend on it.

georgebabey: Makes sense - but maybe add a comment anyways so others (internally) don't accidentally build on it

edx_proctoring/static/index.js
georgebabey: Agreed the ping/echo makes sense, but it seems the type of message would be `ping` while the message itself, if we were sending data along, would be `echo`, I was just getting hung up on the `onPing()` responding by sending an `echo` type

PR #466, opened 2018-11-27
ping-echo implementation

matthugs:
TODO: think farther through the passing of the backend's JS URL via JS
global; can the two approaches be reconciled?

### DIFF
--- a/edx_proctoring/static/index.js
+++ b/edx_proctoring/static/index.js
@@ -1,38 +1,40 @@
 export const handlerWrapper = (Handler) => {
-   let handler = new Handler({});
+  let handler = new Handler({});
 
-  self.addEventListener("message", (message) => {
-    switch(message.data.type) {
+  self.addEventListener('message', (message) => {
+    switch (message.data.type) {
       case 'config': {
         handler = new Handler(message.data.options);
         break;
       }
       case 'startExamAttempt': {
-        if(handler.onStartExamAttempt) {
+        if (handler.onStartExamAttempt) {
           handler.onStartExamAttempt()
-                 .then(() => self.postMessage({type: 'examAttemptStarted'}))
-                 .catch(() => self.postMessage({type: 'examAttemptStartFailed'}))
+            .then(() => self.postMessage({ type: 'examAttemptStarted' }))
+            .catch(() => self.postMessage({ type: 'examAttemptStartFailed' }));
         }
         break;
       }
       case 'endExamAttempt': {
-        if(handler.onEndExamAttempt) {
+        if (handler.onEndExamAttempt) {
           handler.onEndExamAttempt()
-                 .then(() => self.postMessage({type: 'examAttemptEnded'}))
-                 .catch(() => self.postMessage({type: 'examAttemptEndFailed'}))
+            .then(() => self.postMessage({ type: 'examAttemptEnded' }))
+            .catch(() => self.postMessage({ type: 'examAttemptEndFailed' }));
         }
         break;
       }
       case 'ping': {
-        if(handler.onPing) {
+        if (handler.onPing) {
           handler.onPing()
-                 .then(() => self.postMessage({type: 'echo'}))
-                 .catch(() => self.postMessage({type: 'pingFailed'}))
+            .then(() => self.postMessage({ type: 'echo' }))
+            .catch(() => self.postMessage({ type: 'pingFailed' }));
         }
+        break;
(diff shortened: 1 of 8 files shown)


edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -78,6 +78,10 @@ var edx = edx || {};
                 // remove callback on unload event
                 $(window).unbind('beforeunload', this.unloadMessage);
             }
+            var desktopApplicationJsUrl = this.model.get('desktop_application_js_url');
+            if (desktopApplicationJsUrl && !edx.courseware.proctored_exam.configuredWorkerURL) {
```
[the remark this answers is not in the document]
matthugs: For starting and ending the exam, the global variable `configuredWorkerURL` is set directly before `examStartHandler` and `examEndHandler` (which your link points to) are attached as listeners to DOM events. That setting happens in django templates that are specific to the start- and end-specific pages of the proctoring flow.

For the purposes of ping, we can't apply a similar approach, since ping needs to happen all over courseware. That's why we set it here as well. In the special case of the ready_to_start and ready_to_submit pages, we'll see `edx.courseware.proctored_exam.configuredWorkerURL` already set and not try to set it again.

edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -140,6 +144,9 @@ var edx = edx || {};
         updateRemainingTime: function (self) {
             self.timerTick ++;
             self.secondsLeft --;
+            if (self.timerTick % self.poll_interval === self.poll_interval / 2) {
+              edx.courseware.proctored_exam.pingApplication().catch(self.submitExamForFailureState.bind(self));
```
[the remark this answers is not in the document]
matthugs: `edx.courseware.proctored_exam.pingApplication` is again a JS global; it’s the only means I have of passing values between different JS files in edx-proctoring.

`catch` is a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) thing.

edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -175,6 +182,18 @@ var edx = edx || {};
                 self.reloadPage();
             }
         },
+        submitExamForFailureState: function () {
```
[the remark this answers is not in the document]
matthugs: that's actually why we have `location.reload()` here; to reload the page after the exam's status has been changed. That'll effectively look like navigating to the interstitial from the user's perspective (and is the approach applied for the timeout condition as well).

edx_proctoring/static/index.js
```diff
@@ -9,19 +9,25 @@ export const handlerWrapper = (Handler) => {
       }
       case 'startExamAttempt': {
         if(handler.onStartExamAttempt) {
-          handler.onStartExamAttempt().then(() => self.postMessage({type: 'examAttemptStarted'}))
+          handler.onStartExamAttempt()
+                 .then(() => self.postMessage({type: 'examAttemptStarted'}))
+                 .catch(() => self.postMessage({type: 'examAttemptStartFailed'}))
```
nedbat: A very minor point, but do we have formatting guidelines that cover whether to use semicolons to terminate statements, and whether to use a space in `if (` ?  The files in this PR differ.

edx_proctoring/tests/test_views.py
```diff
@@ -1107,7 +1107,8 @@ def test_download_software_clicked_action(self):
 
     @ddt.data(
         ('submit', ProctoredExamStudentAttemptStatus.submitted),
-        ('decline', ProctoredExamStudentAttemptStatus.declined)
+        ('decline', ProctoredExamStudentAttemptStatus.declined),
+        ('error', ProctoredExamStudentAttemptStatus.error)
```
nedbat: In Python files (at least) you can use trailing commas to keep future diffs cleaner.

edx_proctoring/static/index.js
MichaelRoytman: Can we use https://github.com/edx/eslint-config-edx for this?

matthugs: We sure do; they're both implemented in eslint configs we don't have set up for this repo. Thanks for pointing these out.

PR #468, opened 2018-11-28
Ensure that the instructions are translated on the backend

davestgermain:
[EDUCATOR-3687](https://openedx.atlassian.net/browse/EDUCATOR-3687)

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -6,6 +6,7 @@ import logging
 import time
 import uuid
 import pkg_resources
+
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus
 from edx_rest_api_client.client import OAuthAPIClient
@@ -89,7 +90,7 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
         """
         url = self.config_url
         log.debug('Requesting config from %r', url)
-        response = self.session.get(url).json()
+        response = self.session.get(url, headers=self._get_language_headers()).json()
         return response
 
     def get_exam(self, exam):
@@ -200,6 +201,20 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
         log.debug('Created instructor url for %r %r %r', course_id, exam_id, attempt_id)
         return url
 
+    def _get_language_headers(self):
+        """
+        Makes the Accept-Language header based on the currently active language
+        """
+        from django.conf import settings
+        from django.utils.translation import get_language
+
+        headers = {}
+        current_lang = get_language()
+        default_lang = settings.LANGUAGE_CODE
+        if current_lang and current_lang != default_lang:
+            headers['Accept-Language'] = '{};{}'.format(current_lang, default_lang)
+        return headers
+
     def _make_attempt_request(self, exam, attempt, method='POST', status=None, **payload):
         """
         Calls backend attempt API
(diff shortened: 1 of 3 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -200,6 +201,20 @@ def get_instructor_url(self, course_id, user, exam_id=None, attempt_id=None):
         log.debug('Created instructor url for %r %r %r', course_id, exam_id, attempt_id)
         return url
 
+    def _get_language_headers(self):
+        """
+        Makes the Accept-Language header based on the currently active language
+        """
+        from django.conf import settings
+        from django.utils.translation import get_language
```
nedbat: Importing django here is because we want to wait to import them?  Wouldn't it work to have them at the top?

edx_proctoring/backends/rest.py
```diff
@@ -200,6 +201,20 @@ def get_instructor_url(self, course_id, user, exam_id=None, attempt_id=None):
...
+
+        headers = {}
+        current_lang = get_language()
+        default_lang = settings.LANGUAGE_CODE
+        if current_lang and current_lang != default_lang:
```
nedbat: Does it hurt to send the header even if the current language is the default?  Being explicit with a third-party service would be better, no?

edx_proctoring/backends/rest.py
```diff
@@ -200,6 +201,20 @@ def get_instructor_url(self, course_id, user, exam_id=None, attempt_id=None):
         log.debug('Created instructor url for %r %r %r', course_id, exam_id, attempt_id)
         return url
 
+    def _get_language_headers(self):
+        """
+        Makes the Accept-Language header based on the currently active language
```
nedbat: "Returns a dictionary of headers for the Accept-Language headers..."

edx_proctoring/backends/tests/test_rest.py
```diff
@@ -89,6 +89,27 @@ def test_get_attempt(self):
         external_attempt = self.provider.get_attempt(attempt)
         self.assertEqual(external_attempt, attempt)
 
+    @responses.activate
+    def test_get_attempt_i18n(self):
+        from django.utils.translation import activate
```
nedbat: Again: why is this imported here?

edx_proctoring/backends/rest.py
davestgermain: Yes, developers of the subclass may want to import the class, without having to have django installed and configured. Perhaps this function belongs somewhere else?

edx_proctoring/backends/tests/test_rest.py
davestgermain: That one was an oversight.

edx_proctoring/backends/rest.py
davestgermain: I thought about putting the function elsewhere, but then I'd have to pass in the language to get_attempt, which would require changing the signatures of the other backends. It just seemed a little messier that way.

edx_proctoring/backends/rest.py
davestgermain: It'll be sent all the time now.

edx_proctoring/backends/rest.py
nedbat: OK, put a comment here explaining the "why".

edx_proctoring/backends/rest.py
nedbat: Cool.

PR #469, opened 2018-11-28
Ensure that timed exams do not attempt to update the backend

davestgermain:


edx_proctoring/backends/__init__.py
```diff
@@ -9,6 +9,10 @@ def get_backend_provider(exam=None):
     Returns an instance of the configured backend provider
     """
     backend_name = None
-    if exam and exam['backend']:
-        backend_name = exam['backend']
+    if exam:
+        if exam['backend']:
+            backend_name = exam['backend']
+        elif 'is_proctored' in exam and not exam['is_proctored']:
```
nedbat: This triggers only if there's an explicit 'is_proctored':False item in the exam.  Is that what you wanted?  A missing 'is_proctored' key will continue on.

edx_proctoring/backends/__init__.py
```diff
@@ -9,6 +9,10 @@ def get_backend_provider(exam=None):
...
+            # timed exams don't have a backend
+            return None
     return apps.get_app_config('edx_proctoring').get_backend(name=backend_name)
```
nedbat: What does this line return if backend_name is None?

edx_proctoring/tests/test_api.py
```diff
@@ -1654,6 +1654,25 @@ def test_status_summary_bad(self):
 
         self.assertIsNone(summary)
 
+    @patch('edx_proctoring.backends.tests.test_backend.TestBackendProvider')
+    def test_update_exam_attempt_timed(self, mock_backend):
```
nedbat: I'm suspicious that this mock isn't doing what you want.  tests/__init__.py makes one instance of TestBackendProvider, and uses that for all the tests, I think.  "assert_not_called" may be true just because the mock isn't installed properly.

edx_proctoring/backends/__init__.py
davestgermain: yes, that's what I intended, because there's a case where a caller calls:
`get_backend_provider({'backend': 'software_secure'})`

edx_proctoring/backends/__init__.py
davestgermain: It returns the platform default

PR #470, opened 2018-11-29
Add JS deploy step for specially marked npmvX.X.X tags

matthugs:
To enable us to get our js code into npm, we need to set up our CI to push to the npm artifact registry.

Awaiting a pull request from @feanil to add authentication keys.


feanil:
Added the encrypted key as a new commit on the PR.

PR #472, opened 2018-11-30
When there is no instructor dashboard for the backend, return a 404 instead of a 500

davestgermain:
This slightly improves the LMS instructor dashboard experience when the course proctoring backend does not offer a review dashboard. The accordion option will still appear, but the iframe will show a 404 message instead of a 500 error...

### DIFF
--- a/edx_proctoring/apps.py
+++ b/edx_proctoring/apps.py
@@ -53,11 +53,14 @@ class EdxProctoringConfig(AppConfig):
         config = settings.PROCTORING_BACKENDS
 
         self.backends = {}  # pylint: disable=W0201
+        not_found = []
         for extension in ExtensionManager(namespace='openedx.proctoring'):
             name = extension.name
             try:
                 options = config[name]
                 self.backends[name] = extension.plugin(**options)
             except KeyError:
-                warnings.warn("No proctoring backend configured for '{}'.  "
-                              "Available: {}".format(name, list(self.backends)))
+                not_found.append(name)
+        if not_found:  # pragma: no branch
+            warnings.warn("No proctoring backend configured for '{}'.  "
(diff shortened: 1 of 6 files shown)


edx_proctoring/backends/backend.py
```diff
@@ -82,3 +82,10 @@ def get_attempt(self, attempt):
             dict: backend exam attempt object
         """
         return attempt
+
+    # pylint: disable=unused-argument
+    def get_instructor_url(self, course_id, user, exam_id=None, attempt_id=None):
```
MichaelRoytman: Why isn't this an abstract method? Because this is the default behavior?

edx_proctoring/apps.py
```diff
@@ -53,11 +53,14 @@ def ready(self):
...
         for extension in ExtensionManager(namespace='openedx.proctoring'):
             name = extension.name
             try:
                 options = config[name]
                 self.backends[name] = extension.plugin(**options)
             except KeyError:
-                warnings.warn("No proctoring backend configured for '{}'.  "
-                              "Available: {}".format(name, list(self.backends)))
+                not_found.append(name)
+        if not_found:  # pragma: no branch
```
MichaelRoytman: Why not test this?

davestgermain: It turns out to be very difficult to test both cases (when some backend settings exist, and when all backend settings exist) because this function gets called once by the django app loading machinery.

edx_proctoring/backends/backend.py
davestgermain: Yeah, it'll be the default behavior to do nothing. Making it an abstract method is a philosophical decision. Personally, I think having this base class be abstract doesn't help very much.

PR #474, opened 2018-12-04
Improve testing around absent bundles

matthugs:
I mucked up my tests around these cases; this fixes that up.

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -3,10 +3,12 @@ Base implementation of a REST backend, following the API documented in
 docs/backends.rst
 """
 import logging
+import warnings
 import time
 import uuid
 
 from webpack_loader.utils import get_files
+from webpack_loader.exceptions import BaseWebpackLoaderException, WebpackBundleLookupError
 
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus
@@ -76,10 +78,33 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
         Returns the url of the javascript bundle into which the provider's JS will be loaded
         """
         package = self.__class__.__module__.split('.')[0]
-        bundle_chunks = get_files(package, config="WORKERS")
-        if bundle_chunks:
-            return bundle_chunks[0]["url"]
-        return ''
+        js_url = ''
+        try:
+            bundle_chunks = get_files(package, config="WORKERS")
+            # still necessary to check, since webpack_loader can be
+            # configured to ignore all matching packages
+            if bundle_chunks:
+                js_url = bundle_chunks[0]["url"]
+
+        except WebpackBundleLookupError:
+            warnings.warn(
+                u'Could not find webpack bundle for proctoring backend {package}.'
+                u' Check whether webpack is configured to build such a bundle'.format(
+                    package=package
+                )
+            )
+        except BaseWebpackLoaderException:
+            warnings.warn(
+                u'Could not find webpack bundle for proctoring backend {package}.'.format(
+                    package=package
+                )
+            )
+        except IOError as err:
+            warnings.warn(
+                u'Webpack stats file corresponding to WebWorkers not found: {}'
+                .format(str(err))
+            )
+        return js_url
 
     def get_software_download_url(self):
         """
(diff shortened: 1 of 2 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -76,7 +77,11 @@ def get_javascript(self):
         Returns the url of the javascript bundle into which the provider's JS will be loaded
         """
         package = self.__class__.__module__.split('.')[0]
-        bundle_chunks = get_files(package, config="WORKERS")
+        bundle_chunks = ''
+        try:
+            bundle_chunks = get_files(package, config="WORKERS")
+        except BaseWebpackLoaderException:
+            pass
```
davestgermain: It's generally not great to swallow exceptions with `pass`. You can emit a warning, instead, so that somebody looking at the logs can understand what might be going on.

And some comments explaining what's going on would be good.

Also, you can do `try: ... except: ... else:` here

matthugs: I don't actually see a benefit of try-except-else, but I appreciate you letting me know that that's a thing.

edx_proctoring/backends/rest.py
```diff
@@ -76,7 +77,22 @@ def get_javascript(self):
         Returns the url of the javascript bundle into which the provider's JS will be loaded
         """
         package = self.__class__.__module__.split('.')[0]
-        bundle_chunks = get_files(package, config="WORKERS")
+        bundle_chunks = ''
+        try:
+            bundle_chunks = get_files(package, config="WORKERS")
+        except WebpackBundleLookupError:
+            log.warning(
+                u'Could not find webpack bundle for proctoring backend %s.' +
```
davestgermain: + isn't required for string continuations. Also, perhaps `warnings.warn(message)` would be better than `log.warning(message)`, because it'll only appear once in the logs.

edx_proctoring/backends/rest.py
```diff
@@ -76,7 +77,22 @@ def get_javascript(self):
...
+                u' Check whether webpack is configured to build such a bundle',
+                package
+            )
+        except BaseWebpackLoaderException:
+            log.warning(
+                u'Could not find webpack bundle for proctoring backend %s.',
+                package
+            )
+        # still necessary to check, since webpack_loader can be
```
davestgermain: Make the rest of this in an else on the try/except block.

edx_proctoring/backends/tests/test_rest.py
```diff
@@ -229,11 +230,23 @@ def test_get_javascript_bundle(self, get_files_mock):
         self.assertEqual(javascript_url, '/there/it/is')
 
     @patch('edx_proctoring.backends.rest.get_files')
-    def test_get_javascript_empty_bundle(self, get_files_mock):
+    def test_get_javascript_empty_return(self, get_files_mock):
         get_files_mock.return_value = []
         javascript_url = self.provider.get_javascript()
         self.assertEqual(javascript_url, '')
 
+    @patch.object(webpack_loader.loader.WebpackLoader, 'get_assets')
```
davestgermain: You don't have to import the object in order to patch it. `@patch('webpack_loader.loader.WebpackLoader.get_assets')` should work.

edx_proctoring/backends/rest.py
matthugs: warnings.warn messages are probably excessively noisy as output from running unit tests. They may cause them to fail, as far as CI is concerned. Let's find out!

edx_proctoring/backends/rest.py
```diff
@@ -76,10 +78,28 @@ def get_javascript(self):
         Returns the url of the javascript bundle into which the provider's JS will be loaded
         """
         package = self.__class__.__module__.split('.')[0]
-        bundle_chunks = get_files(package, config="WORKERS")
-        if bundle_chunks:
-            return bundle_chunks[0]["url"]
-        return ''
+        js_url = ''
+        try:
+            bundle_chunks = get_files(package, config="WORKERS")
```
davestgermain: Is it still possible for this method to raise IOError? I know the tests still have a case expecting to catch IOError, but this method should probably be resilient enough to handle misconfiguration or missing npm packages, etc. (as long as the warning messages are clear for each edge case)

davestgermain:
👍

PR #476, opened 2018-12-06
Use instructions from the subclass instead of from the API request

davestgermain:
In order to ensure that instructions are reviewable and translated, we want to have them statically defined. Backends will override the `proctoring_instructions` property, returning a list of strings wrapped with `gettext`. 

In edx-platform, add the backend module name to `conf/locale/config.yaml` (third_party section), and our i18n_tool will automatically find and merge their strings.

[EDUCATOR-3767](https://openedx.atlassian.net/browse/EDUCATOR-3767)


edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -195,7 +195,7 @@ <h2 id="acessible-error-title">{% trans "Cannot Start Proctored Exam" %}</h2>
       },
       success: function (data) {
         newWindow.location = "{{software_download_url|safe}}";
-        setTimeout(poll_until_ready, 15000);
+        // setTimeout(poll_until_ready, 15000);
```
matthugs: probably better to delete this altogether. Why is this necessary?

davestgermain: not sure if we're going to want to revive it. but I'll remove it for now.

matthugs: I'd remove the dead code as well; it's better habit to be in to not leave this stuff in place, since it inflates what needs to be sent to the browser.

PR #479, opened 2018-12-07
remove rules as parameter for create and update of exam review policy

MichaelRoytman:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@edx/edx-proctoring",
   "//": "Be sure to update the version number in edx_proctoring/__init__.py",
-  "version": "1.5.0-rc.1",
+  "version": "1.5.0rc2",
```
matthugs: The dash is necessary; we should keep it in the same format it was in before. As is I think this would fail to publish, as according to [semver](https://github.com/npm/node-semver) it's not a valid version number.

This is my bad, but would you mind adding another (weird) json comment-ish key-value pair above this to point out that the formats of the python and npm versions are expected to be slightly different? You can add it as another "//"-keyed string, like the one above this line.

matthugs:
```sh
C02TP0LBG8WL% npm install semver
+ semver@5.6.0
added 1 package and audited 1 package in 1.084s
found 0 vulnerabilities

C02TP0LBG8WL% node
> var semver = require('semver');
undefined
> semver.valid('1.5.0rc2');
null
> semver.valid('1.5.0-rc.2');
'1.5.0-rc.2'
```

MichaelRoytman: Thank you for pointing this out. I will fix the version number and will clarify this with a weird comment-ish thing.

PR #480, opened 2018-12-07
Fix bug wherein non-JS proctoring providers attempt to ping

matthugs:
... leading them to always fail/time out


davestgermain:
👍

PR #487, opened 2018-12-11
Enables the use of the Open edX Pluggable Django App framework

davestgermain:
The [Pluggable Django App framework](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/plugins/README.rst) allows Open edX Django apps to be more self-contained and flexible. 

The bulk of this diff refactors the urls to be properly namespaced. There's no change in behavior, but moving forward, it should be easier to make settings changes by isolating them to PRs in this repo.


Makefile
```diff
@@ -22,12 +22,10 @@ help: ## display this help message
...
 clean: ## remove generated byte code, coverage reports, and build artifacts
-	find . -name '*.pyc' -exec rm -f {} +
-	find . -name '*.pyo' -exec rm -f {} +
-	find . -name '*~' -exec rm -f {} +
-	rm -fr build/
-	rm -fr dist/
-	rm -fr *.egg-info
+	find . -name '*.pyc' -delete
+	find . -name '*.pyo' -delete
+	find . -name '*~' -delete
```
nedbat: I'm just curious why you changed these find commands?

edx_proctoring/backends/rest.py
```diff
@@ -161,7 +162,10 @@ def register_exam_attempt(self, exam, context):
         payload.pop('attempt_code', False)
         log.debug('Creating exam attempt for %r at %r', exam['external_id'], url)
         response = self.session.post(url, json=payload)
+        if response.status_code != 200:
+            raise BackendProvideCannotRegisterAttempt(response.content)
```
nedbat: We have a log.debug for every 200 response below.  Do we want to log something on the failures? Or do we figure that will be handled higher up?

nedbat: PS: You are missing an "r" from "BackendProvideCannot..."

edx_proctoring/settings/common.py
```diff
@@ -0,0 +1,38 @@
...
+        'DEFAULT': 'null',
+        'null': {},
+    }
+    proctoring_js = (
+        [
+            'proctoring/js/models/proctored_exam_allowance_model.js',
+            'proctoring/js/models/proctored_exam_attempt_model.js',
+            'proctoring/js/models/proctored_exam_model.js'
+        ] +
+        [
```
nedbat: Why use this [..] + [..] syntax?

nedbat: Matt points out this is the way the lines were in the settings file.  We don't need to keep them like this.

Makefile
davestgermain: `-delete` is noticeably faster. Otherwise, find has to run `rm -rf` for every file. I think that's why find has this common feature.

edx_proctoring/backends/rest.py
davestgermain: The failures will bubble up and be logged nicely. 

`BackendProvideCannotRegisterAttempt` is actually the name of the exception... I'll change that.

Makefile
nedbat: That's why they used `+` instead of `;`: find will bunch up arguments to spawn fewer processes, but OK.

edx_proctoring/settings/common.py
nedbat: :+1:

PR #491, opened 2018-12-17
ensure webworker JS will work on IE

matthugs:


davestgermain:
I don't claim to know how this pipeline stuff is supposed to work.
👍

PR #493, opened 2018-12-18
[EDUCATOR-3768]: configurability of proctoring ping interval

matthugs:
[JIRA](https://openedx.atlassian.net/browse/EDUCATOR-3768)

### DIFF
--- a/edx_proctoring/backends/backend.py
+++ b/edx_proctoring/backends/backend.py
@@ -5,12 +5,15 @@ Defines the abstract base class that all backends should derive from
 import abc
 import six
 
+from edx_proctoring import constants
+
 
 class ProctoringBackendProvider(six.with_metaclass(abc.ABCMeta)):
     """
     The base abstract class for all proctoring service providers
     """
     verbose_name = u'Unknown'
+    ping_interval = constants.DEFAULT_DESKTOP_APPLICATION_PING_INTERVAL_SECONDS
 
     @abc.abstractmethod
     def register_exam_attempt(self, exam, context):
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import
 import json
 import logging
 import six
+import waffle
 
 from django.conf import settings
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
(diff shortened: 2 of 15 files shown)


matthugs:
codecov is wrong; I do have coverage of that line

edx_proctoring/backends/backend.py
```diff
@@ -63,6 +65,22 @@ def get_javascript(self):
         """
         return ""
 
+    def get_ping_failure_should_block_user(self):
```
davestgermain: This will probably need to be determined per attempt, not per backend. If it's set globally like this, then the ping is irrelevant. 

The poll interval can be set per backend, but the failure handling could be implemented using `ProctoredExamStudentAllowance`.

edx_proctoring/views.py
```diff
@@ -506,8 +515,11 @@ def get(self, request):  # pylint: disable=unused-argument
 
             if provider:
                 desktop_application_js_url = provider.get_javascript()
+                ping_interval = provider.get_ping_interval()
             else:
                 desktop_application_js_url = ''
+                ping_interval = (
```
davestgermain: If there's no provider, then this is a timed exam, in which case, there's no desktop application to ping.

edx_proctoring/backends/backend.py
matthugs: This reminded me to update the two tickets we had subsumed into this work to reference one another; this is out of scope.

davestgermain: This isn't really part of the public backend api. Since it's only used in one place, I think the logic of checking the waffle flag should move to the caller. That would save some testing, complexity, and coverage.

edx_proctoring/backends/__init__.py
```diff
@@ -16,3 +16,6 @@ def get_backend_provider(exam=None):
         elif exam['backend']:
             backend_name = exam['backend']
     return apps.get_app_config('edx_proctoring').get_backend(name=backend_name)
+
+
+PING_FAILURE_PASSTHROUGH_TEMPLATE = 'edx_proctoring.{}_ping_failure_passthrough'
```
davestgermain: How about moving this to constants? 

Also, I just remembered that devops said they're deprecating waffle *switches*, but not flags. I'm not sure what the recommended replacement is, yet, but this should be fine for now.

davestgermain:
Also, could you increase the version number?

edx_proctoring/backends/__init__.py
matthugs: Done.

Oh gosh, I do remember that. If someone mentions it to them, we can remind them they need to update [the OEP on the subject](https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0017-bp-feature-toggles.rst#use-case-4-ops-graceful-degradation), since it's explicit about recommending the use of switches for most use cases.

PR #498, opened 2019-01-02
Block starting second proctored session

matthugs:
[EDUCATOR-3844](https://openedx.atlassian.net/browse/EDUCATOR-3844)

Prevents users from starting two independent proctored exam sessions by inserting _yet another_ interstitial in the situation in which another exam is already active at the time we go to start an exam.


edx_proctoring/templates/proctored_exam/other_exam_in_progress.html
```diff
@@ -0,0 +1,15 @@
...
+<div class="failure sequence proctored-exam" data-exam-id="{{exam_id}}">
+  <h3>
+    {% blocktrans %}
+      Complete in-progress exam
+    {% endblocktrans %}
+  </h3>
+
+  <p>
+    {% blocktrans %}
+      You must complete your currently in-progress exam before you can attempt this one.
```
davestgermain: Since we already have the running attempt available, we should be able to link back to the other exam, right?

edx_proctoring/api.py
```diff
@@ -1995,6 +2024,35 @@ def get_student_view(user_id, course_id, content_id,
...
+    try:
+        # resolve the LMS url, note we can't assume we're running in
+        # a same process as the LMS
+        exam_url = reverse('jump_to', args=[exam['course_id'], exam['content_id']])
+    except NoReverseMatch:
+        log.exception("Can't find exam url for course %s", exam['course_id'])
+    return exam_url
+
+
+def _get_running_exam_info(user_id, currently_visited_exam_id):
```
davestgermain: How about a single function like:
```python
def _get_running_exam_info(user_id, currently_visited_exam_id):
    """
    Check whether there are any other currently active exams besides
    that currently being viewed
    """
    active_exam_attempts = get_active_exams_for_user(user_id)
    if active_exam_attempts:
        exam = active_exams[0]['exam']
        is_other_exam = exam['id'] != currently_visited_exam_id
        try:
            # resolve the LMS url, note we can't assume we're running in
            # a same process as the LMS
            exam_url = reverse('jump_to', args=(exam['course_id'], exam['content_id']))
        except NoReverseMatch:
            log.exception("Can't find exam url for course %s", exam['course_id'])
            exam_url = ''
    else:
        is_other_exam, exam_url = False, ''
    return is_other_exam, exam_url
```

davestgermain:
👍

PR #500, opened 2019-01-03
EDUCATOR-3857 -- switching from proctored to timed

davestgermain:
[EDUCATOR-3857](https://openedx.atlassian.net/browse/EDUCATOR-3857)

If a proctored exam switches to being a timed exam, the backend should be notified that the exam is now "inactive".

### DIFF
--- a/edx_proctoring/signals.py
+++ b/edx_proctoring/signals.py
@@ -1,0 +1,106 @@
+"edx-proctoring signals"
+from django.db.models.signals import pre_save, post_save, pre_delete
+from django.dispatch import receiver
+from edx_proctoring import models
+from edx_proctoring.backends import get_backend_provider
+
+
+@receiver(post_save, sender=models.ProctoredExamReviewPolicy)
+@receiver(post_save, sender=models.ProctoredExam)
+def _save_exam_on_backend(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Save the exam to the backend provider when our model changes.
+    It also combines the review policy into the exam when saving to the backend
+    """
+    if sender == models.ProctoredExam:
+        exam_obj = instance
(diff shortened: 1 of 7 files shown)


edx_proctoring/signals.py
```diff
@@ -0,0 +1,124 @@
+"edx-proctoring signals"
+from django.db.models.signals import pre_save, post_save, pre_delete
+from django.dispatch import receiver
+from edx_proctoring import models
+from edx_proctoring.backends import get_backend_provider
+
+
+@receiver(pre_save, sender=models.ProctoredExam)
+def check_for_category_switch(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
davestgermain: Note: this is the only functional difference. Everything else in the PR is cleanup.

edx_proctoring/signals.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    if instance.id:
+        original = sender.objects.get(pk=instance.id)
+        if original.is_proctored and instance.is_proctored != original.is_proctored:
+            from edx_proctoring.serializers import ProctoredExamSerializer
+            exam = ProctoredExamSerializer(instance).data
+            exam['is_active'] = False
+            exam['is_proctored'] = True
+            # we have to pretend that the exam is still proctored
+            # or else we get_backend_provider will return None
```
matthugs: s/we //

matthugs: What's unclear from the code even with this comment: the is_proctored field being set here won't affect what backend.on_exam_saved does

I think this would be cleared up by eliminating the need for pretending by adding a get_backend_provider_by_name (or the like) to pull out the final line of get_backend_provider. That way there's no room to worry about whether the backend will care about whether is_proctored is set.

davestgermain: I changed get_backend_provider to return a backend for the given name. Good idea.

PR #506, opened 2019-01-09
proctortrack-specific copy changes

matthugs:
[EDUCATOR-3842](https://openedx.atlassian.net/browse/EDUCATOR-3842)

separated out the copy changes which don't touch RPNow at all, since we're more confident they can go through immediately

see also: https://github.com/joshivj/edx-proctoring-proctortrack/pull/5

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1715,7 +1715,9 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
     except NoReverseMatch:
         log.exception("Can't find progress url for course %s", course_id)
 
-    return {
+    provider = get_backend_provider(exam)
+
+    context = {
         'platform_name': settings.PLATFORM_NAME,
         'total_time': total_time,
         'exam_id': exam['id'],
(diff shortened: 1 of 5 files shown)


edx_proctoring/api.py
```diff
@@ -1695,7 +1695,7 @@ def _calculate_allowed_mins(exam, user_id):
     return allowed_time_limit_mins
 
 
-def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_exam=False):
+def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_exam=False, provider=None):
```
davestgermain: Instead of passing in the provider, create the instance of it here and add the items from `_get_proctored_exam_view` and `_get_practice_exam_view`. Those functions only create the backend in order to add stuff to the context.

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -8,13 +8,31 @@ <h3>
...
+          to verify your identity. Before you begin, make sure you are
+          on a computer with a webcam, and that you have a valid form
+          of photo identification such as a driver’s license or
+          passport.
+        {% endblocktrans %}
       </p>
-
+      <ol>
+        {% for instruction in backend_instructions %}
+          <li>{{instruction|safe}}
```
davestgermain: no closing tag

edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -8,13 +8,31 @@ <h3>
     {% endblocktrans %}
     </h3>
     {% if backend_instructions %}
-      {% for instruction in backend_instructions %}
-      <p>{{instruction|safe}}</p>
-      {% endfor %}
       <p>
-        <a href="#" id="software_download_link" data-action="click_download_software" target="_blank">{% trans "Start System Check" %}</a>
+        {% blocktrans %}
```
davestgermain: Is the intention that this changed language only applies to backends that define instructions? (in other words, the software_secure backend will still show the old instructions, and the old button/link)

matthugs: Yes, you're reading the code correctly

PR #508, opened 2019-01-10
Expose error state to proctoring rest API implementers

matthugs:
[EDUCATOR-3887](https://openedx.atlassian.net/browse/EDUCATOR-3887)


matthugs:
@davestgermain coverage addressed

PR #509, opened 2019-01-10
Added developer doc

davestgermain:


docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+    Install the proctoring package and mockprock into edx-platform in the container, for both LMS and Studio:
+
+Edit or create the ``edx-platform/requirements/private.txt`` file::
+
+    # edx-platform/requirements/private.txt
+    # This file will be used in the docker image, so use file paths that work there.
+    -e /edx/src/edx-proctoring
+    -e /edx/src/mockprock
+
+Then::
```
matthugs:
```suggestion
The packages we specified in ``private.txt`` will be automatically pulled down by::
```

docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+
+    $ make lms-shell
+    root@a7ff4f3f3f6b:/edx/app/edxapp/edx-platform# paver install_prereqs;exit
+    $ make studio-shell
+    root@a7ff4f3f3f6b:/edx/app/edxapp/edx-platform# paver install_prereqs;exit
+
+
+You can also develop edx-proctoring outside of devstack with a virtual environment::
```
matthugs: I'm not sure why in my own development process I'd want to do this; do you think there's value in including this? I'm not sure what sorts of benefit folks could glean from developing independently in this way

docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+
+    cd edx-proctoring
+    mkvirtualenv edx-proctoring
+    make upgrade requirements
+    make test-python
+
+In edx-platform/lms/envs/private.py and edx-platform/cms/envs/private.py:
```
matthugs: Can we move the rest of this section (i.e. lines 57-91) to its own section? I think a quick blurb motivating what mockprock is and why it's helpful could be useful.

docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+
+    python -m mockprock.server {client_id} {client_secret}
+
+How do I debug proctoring tests?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Set up a virtual environment, then::
+
+    cd /edx/src/edx-proctoring
+    make test-all
```
matthugs: This `make` command uses `tox`, so there shouldn't be any benefit to setting up a venv as this suggests a few lines above

matthugs: (I think we might need to replace most of this section; some of this stuff looks like it was outdated before we started this project.)

docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+* Restart Studio and LMS::
+
+    sudo /edx/bin/supervisorctl restart lms cms
+
+* Create a test course
+
+* Enroll verified@example.com in the course
+* Log in to Django admin
+* Add a verified course mode for your course
+* Update the verified user's mode to be "verified"
```
matthugs: This is actually easier said than done (the django admin section for course enrollments, where the mode this is referring to is, is hidden by a waffle flag). I'd probably recommend a new developer to use an admin user with masquerading instead, especially if they're just getting started.

docs/developing.rst
```diff
@@ -0,0 +1,185 @@
...
+Release a new version of edx-proctoring
+----------------------------------------
+
+* Update the version in ``edx_proctoring/__init__.py`` and ``package.json``
+* Create a `new release on GitHub <https://github.com/edx/edx-proctoring/releases>`_ using the version number 
+* Send an email to release-notifications@edx.org announcing the new version
+* Update edx-platform to use the new version
+    * In edx-platform, create a branch and update the requirements/edx/base.in file to reflect the new tagged branch. 
+* create a PR of this branch in edx-platform onto edx-platform:master
+* Once the PR onto edx-platform has been merged, the updated edx-proctoring will be live in production when the normally scheduled release goes.
```
matthugs:
```suggestion
* Once the PR onto edx-platform has been merged, the updated edx-proctoring will be live in production when the normally scheduled release completes.
```

docs/developing.rst
davestgermain: I run the tests outside of the docker shell. It avoids devstack flakiness. But I'll take out this section, to avoid confusion.

docs/developing.rst
davestgermain: done

docs/developing.rst
davestgermain: I replaced the "debugging tests" section with this brief "running tests" section

docs/developing.rst
matthugs: As do I! Probably worth dropping into the testing section; something like "This can be done either inside or outside your devstack environment."

PR #511, opened 2019-01-11
Add reviewed_by field to API

matthugs:
The old Software Secure logic requires two things to get an exam attempt into the "rejected" state:
1. The status used must be "suspicious", and
2. The review must have a user associated with it

We need to ask V. to do the former. This allows the latter.

[EDUCATOR-3890](https://openedx.atlassian.net/browse/EDUCATOR-3890)

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -11,6 +11,8 @@ import jwt
 from webpack_loader.utils import get_files
 from webpack_loader.exceptions import BaseWebpackLoaderException, WebpackBundleLookupError
 
+from django.contrib.auth.models import User
+
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.exceptions import BackendProviderCannotRegisterAttempt
 from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus, SoftwareSecureReviewStatus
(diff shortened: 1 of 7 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -197,6 +199,11 @@ def on_review_callback(self, attempt, payload):
         Called when the reviewing 3rd party service posts back the results
         """
         # REST backends should convert the payload into the expected data structure
+        if payload.get('reviewed_by', False):
+            try:
+                payload['reviewed_by'] = User.objects.get(email=payload['reviewed_by'])
```
matthugs: We don't currently expose the email address, but the user id to our rest api implementers (via the JWT token). That's a hole in my diff as of right now.

matthugs: Update: I'm wrong; the JWT includes both the user's id as well as their email. Implementers should be able to provide this.

edx_proctoring/backends/rest.py
```diff
@@ -198,6 +200,11 @@ def on_review_callback(self, attempt, payload):
...
```
davestgermain: I think this will ultimately fail, because the raw payload is later jsonified and saved to the db in a field in the review, and a User object can't be jsonified.

Perhaps it'd be better to do the user lookup in `views.BaseReviewCallback.make_review`

matthugs: Can we also just swap it out to be the original payload, rather than the backend-translated one? I feel like that would be less radical of a change from what I have, and would get us back to the old shape of how we used to store Software Secure reviews.

matthugs: i.e. https://github.com/edx/edx-proctoring/pull/511/commits/4b910b42da07c11ac429d8d35c62104a57b2304f

PR #513, opened 2019-01-15
Obscure the user id going to the backend

davestgermain:
[EDUCATOR-3894](https://openedx.atlassian.net/browse/EDUCATOR-3894)

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -72,6 +72,7 @@ from edx_proctoring.utils import (
     get_time_remaining_for_attempt,
     locate_attempt_by_attempt_code,
     humanized_time,
+    obscured_user_id,
 )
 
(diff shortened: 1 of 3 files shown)


edx_proctoring/views.py
```diff
@@ -1000,7 +1001,7 @@ def get(self, request, course_id, exam_id=None):
             backend = get_backend_provider(exam)
             if backend:
                 user = {
-                    'id': request.user.id,
+                    'id': obscured_user_id(request.user.id, exam['backend']),
```
MichaelRoytman: What's the purpose of including the backend in the hash?

davestgermain: It’s an extra degree of paranoia — each backend will have a different user id for the same user. (It’s a bit of a moot point, considering we’re sending the user’s email address, as well...)

PR #514, opened 2019-01-17
Refactored config dashboard url

davestgermain:
The config option will now be included in the JWT sent to the backend, rather than in a query string.

Also: prevented attempt from being requested on every page

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1753,14 +1753,17 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
         'provider_name': provider.verbose_name,
     }
     if attempt:
-        provider_attempt = provider.get_attempt(attempt)
-        download_url = provider_attempt.get('download_url', None) or provider.get_software_download_url()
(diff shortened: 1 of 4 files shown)


edx_proctoring/api.py
```diff
@@ -1753,14 +1753,17 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
         'provider_name': provider.verbose_name,
     }
     if attempt:
-        provider_attempt = provider.get_attempt(attempt)
-        download_url = provider_attempt.get('download_url', None) or provider.get_software_download_url()
+        context['exam_code'] = attempt['attempt_code']
+        if attempt['status'] in (ProctoredExamStudentAttemptStatus.created,
+                                 ProctoredExamStudentAttemptStatus.download_software_clicked):
+            # since this may make an http request, let's not include it on every page
```
MichaelRoytman: Does this occur only for REST-based backends?

davestgermain: Yes, non-REST backends don't make any request. `get_attempt` just returns the attempt passed in, unchanged.

PR #516, opened 2019-01-23
update exam ready callback documentation to clarify request structure

MichaelRoytman:


docs/backends.rst
```diff
@@ -184,6 +175,15 @@ The following fields are optional::
 
 Start and stop are seconds relative to the start of the recorded proctoring session. ``reviewed_by`` must be included whenever a specific edX user (e.g. a member of a course team) initiated the review.
```
matthugs: start and stop are (probably) meant to be embedded in comments, and to correspond to the portion of a session being commented on.

PR #517, opened 2019-01-24
Added an API for deleting user data from the configured backends.

davestgermain:
[EDUCATOR-3963](https://openedx.atlassian.net/browse/EDUCATOR-3963)

### DIFF
--- a/edx_proctoring/backends/tests/test_backend.py
+++ b/edx_proctoring/backends/tests/test_backend.py
@@ -14,6 +14,7 @@ from edx_proctoring.backends import get_backend_provider
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.backends.null import NullBackendProvider
 from edx_proctoring.backends.mock import MockProctoringBackendProvider
+from edx_proctoring.exceptions import BackendProviderCannotRetireUser
 
 # pragma pylint: disable=useless-super-delegation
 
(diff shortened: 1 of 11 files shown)


edx_proctoring/backends/tests/test_backend.py
```diff
@@ -84,6 +86,14 @@ def get_instructor_url(self, course_id, user, exam_id=None, attempt_id=None, sho
 
         return url
 
+    # pylint: disable=unused-argument
```
matthugs: Looks like this disable can be removed

edx_proctoring/tests/test_views.py
```diff
@@ -2659,3 +2660,41 @@ def test_launch_for_configuration_dashboard(self):
...
+
+    def test_can_delete_user(self):
+        deletion_url = reverse('edx_proctoring:backend_user_deletion_api', kwargs={'user_id': self.second_user.id})
+
+        response = self.client.delete(deletion_url)
+        assert response.status_code == 200
+        data = response.json()
+        assert data == {'test': True}
+        test_backend = get_backend_provider(name='test')
+        assert test_backend.last_retire_user is not None
```
matthugs: The rest of this file's test classes use `self.assetStuff` calls; is there motivation to break from that for these tests?

davestgermain: It turns out that plain asserts work better than the various assert methods when run with py.test (py.test will print more helpful error messages). I heard recently that it was recommended to switch to plain asserts across the platform, although nobody is going to go back to refactor all of the old tests.

PR #518, opened 2019-01-25
update backend.on_review_callback to not modify payload parameter

MichaelRoytman:


edx_proctoring/views.py
```diff
@@ -844,7 +845,11 @@ def make_review(self, attempt, data, backend=None):
         review.raw_data = json.dumps(data)
         review.student_id = attempt['user']['id']
         review.exam_id = attempt['proctored_exam']['id']
-        review.reviewed_by = backend_review.get('reviewed_by', None)
+
+        try:
+            review.reviewed_by = User.objects.get(email=data['reviewed_by'])
+        except (User.DoesNotExist, KeyError):
```
MichaelRoytman: I'm a little unsure about this. Is this reasonable to do? I'm doing two operations that could raise exceptions. I could substitute this with the following:

```
if review.reviewed_by:
    try:
        review.reviewed_by = User.objects.get(email=data['reviewed_by']:
    except User.DoesNotExist:
        review.reviewed_by = None
else:
    review.reviewed_by = None
```

That might be a little clearer.

davestgermain: The way you have it looks fine, but you may want to log if the user does not exist.  That would be an unusual situation.

PR #520, opened 2019-01-28
EDUCATOR-3958 -- unblock attempt errors when the onboarding review succeeds

davestgermain:


edx_proctoring/api.py
```diff
@@ -1044,7 +1044,16 @@ def update_attempt_status(exam_id, user_id, to_status,
     # call back to the backend to register the end of the exam, if necessary
     backend = get_backend_provider(exam)
     if backend:
-        # only proctored exams have a backend
+        # When onboarding exams change state to a completed status,
+        # look up any exams in the onboarding error states, and delete them.
+        # This will allow learners to return to the proctored exam and continue
+        # through the workflow.
+        if exam['is_practice_exam'] and backend.supports_onboarding and \
+                not ProctoredExamStudentAttemptStatus.is_incomplete_status(to_status):
```
matthugs:
```suggestion
                ProctoredExamStudentAttemptStatus.is_complete_status(to_status):
```

Why the double-negative?

edx_proctoring/api.py
```diff
@@ -1044,7 +1044,16 @@ def update_attempt_status(exam_id, user_id, to_status,
...
+            # find and delete any pending attempts
+            ProctoredExamStudentAttempt.objects.clear_onboarding_errors(user_id)
```
MichaelRoytman: I think this is what I was hung up on before. This implies that if an onboarding exam enters the `ProctoredExamStudentAttemptStatus. onboarding_failed` status, for example (or any "complete" status, for that matter), that we would unblock all proctored exams. Is that desirable? I feel like I'm missing a detail, because to me, we should only unblock when the onboarding exam is successfully completed. Could you clarify, please?

edx_proctoring/backends/backend.py
```diff
@@ -18,6 +18,8 @@ class ProctoringBackendProvider(six.with_metaclass(abc.ABCMeta)):
     tech_support_phone = ''
     # whether this backend supports an instructor review/configuration dashboard
     has_dashboard = False
+    # whether practice exams map to "onboarding" exams for this backend
+    supports_onboarding = False
```
MichaelRoytman: Do you need to add a corresponding property to `rest.py` of `supports_onboarding = True`?

davestgermain: It can go there, or on the subclass of the rest backend. But you're probably right that it'd be expedient to add it to the rest base class.

edx_proctoring/api.py
davestgermain: Because the error status is driven by the backend (onboarding_failed, onboarding_missing, etc), I'm treating the proctored exam attempt as a transient thing. There are several completed states that the onboarding exam could go into (including "error" if they kill the application, for instance), which would lead to different future error messages once the learner tried to retake the proctored exam. Rather than think through the different cases, I'm just unconditionally clearing the error. The worst case scenario is that the learner will immediately try to retake the proctored exam, and then the backend will return the same error, and there will be another attempt in an error state (until the onboarding exam changed state again).

MichaelRoytman: I see. That makes sense. Thank you.

edx_proctoring/backends/backend.py
MichaelRoytman: If it is to go on the subclass, we should update the docs to indicate that.

edx_proctoring/api.py
matthugs: I think the potential point of confusion here is that the `onboarding_<whatever>` statuses aren't intended to be applied to attempts on onboarding exams, which is counterintuitive. They're meant to be applied to the non-onboarding exams to which we'll be gating access. So `onboarding_missing` is not expected to _ever_ apply to an attempt for which `exam['is_practice_exam']` is true. Is that any clearer, @MichaelRoytman?

edx_proctoring/api.py
davestgermain: I've removed the double negative. :)

edx_proctoring/backends/backend.py
davestgermain: I just added it to the rest backend.

PR #522, opened 2019-01-29
LMS copy changes to support using proctortrack's onboarding exams

matthugs:
rather than using practice exams

JIRA:EDUCATOR-3960

### DIFF
--- a/edx_proctoring/templates/practice_exam/error.html
+++ b/edx_proctoring/templates/practice_exam/error.html
@@ -1,28 +1,51 @@
 {% load i18n %}
 <div class="failure sequence proctored-exam" data-exam-id="{{exam_id}}">
   <h3>
-    {% blocktrans %}
-      There was a problem with your practice proctoring session
-    {% endblocktrans %}
+    {% if provider_uses_onboarding %}
+      {% blocktrans %}
+        There was a problem with your onboarding session
+      {% endblocktrans %}
+    {% else %}
+      {% blocktrans %}
+        There was a problem with your practice proctoring session
+      {% endblocktrans %}
+    {% endif %}
   </h3>
 
   <h4>
-    {% blocktrans %}
-      Your practice proctoring results: <b class="failure"> Unsatisfactory </b>
-    {% endblocktrans %}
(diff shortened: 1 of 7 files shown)


edx_proctoring/templates/practice_exam/error.html
```diff
@@ -1,28 +1,51 @@
 {% load i18n %}
 <div class="failure sequence proctored-exam" data-exam-id="{{exam_id}}">
   <h3>
-    {% blocktrans %}
-      There was a problem with your practice proctoring session
-    {% endblocktrans %}
+    {% if provider_uses_onboarding %}
```
MichaelRoytman: I don't think I saw these changes in the copy document. I know it's just switching "practice" for "onboarding", but it might be worth putting it in there for the sake of completeness.

edx_proctoring/tests/test_student_view.py
```diff
@@ -753,7 +753,7 @@ def test_get_studentview_submitted_status_practiceexam(self):
         exam_attempt.save()
 
         rendered_response = self.render_practice_exam()
-        self.assertIn(self.practice_exam_submitted_msg, rendered_response)
+        self.assertIn(self.onboarding_exam_submitted_msg, rendered_response)
```
MichaelRoytman: Would it make sense to update the names of the functions now that the onboarding exam, and not the practice exam, is supported by the test backend? Or perhaps since the onboarding changes are temporary, it isn't worth the flip-flop. What do you think?

matthugs: Yeah, I'd stick closer to leaving all the changes we make sit behind some check of the backend's onboarding flag, so we can easily know what can be safely rolled back when we remove it

edx_proctoring/templates/practice_exam/error.html
matthugs: Added [here (link for internal edX folks only, apologies folks)](https://docs.google.com/a/edx.org/document/d/1xOB4v_YsGpcE9qL7sBi95OIhu5vTR0qhTYlh4_bCyO8/edit?disco=AAAACgydMvQ)

PR #524, opened 2019-01-30
Changed error response for retirement endpoint

davestgermain:
This changes the error response to a 500, so that the retirement pipeline will properly fail. 

It also checks the authenticated user against the retirement permission, which is defined in edx-platform [#19690](https://github.com/edx/edx-platform/pull/19690)

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -2677,7 +2677,7 @@ class TestBackendUserDeletion(LoggedInTestCase):
     def test_can_delete_user(self):
         deletion_url = reverse('edx_proctoring:backend_user_deletion_api', kwargs={'user_id': self.second_user.id})
 
-        response = self.client.delete(deletion_url)
+        response = self.client.post(deletion_url)
         assert response.status_code == 200
         data = response.json()
         # if there is no user data, then no deletion happens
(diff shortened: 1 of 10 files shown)


edx_proctoring/tests/test_views.py
```diff
@@ -2680,13 +2680,32 @@ def test_can_delete_user(self):
...
+
+        response = self.client.delete(deletion_url)
+        assert response.status_code == 200
+        data = response.json()
+        # if there is an attempt, we'll try to delete from the backend
         assert data == {'test': True}
         test_backend = get_backend_provider(name='test')
         assert test_backend.last_retire_user is not None
 
         # running a second time will return a false status
```
bmedx: Any idea if we will have a way to separate out the "no user to delete" case from a real error? We frequently have to re-run users and the way it's written currently will add a fair amount of work for us.

edx_proctoring/views.py
```diff
@@ -1032,24 +1032,28 @@ class BackendUserManagementAPI(AuthenticatedAPIView):
     """
     Manage user information stored on the backends
     """
-    @method_decorator(require_staff)
     def delete(self, request, user_id):  # pylint: disable=unused-argument
```
bmedx: Any chance we can change this to a POST since it's not actually deleting a resource from this service?

edx_proctoring/tests/test_views.py
davestgermain: "no user to delete" will be a 200 response with `"{}"` as content

"deleted user " will be a 200 response with `"{"backend": true}"`

PR #529, opened 2019-02-05
Added permission check for proctoring reviews

davestgermain:
[EDUCATOR-3918](https://openedx.atlassian.net/browse/EDUCATOR-3918)

This will require a new django user group for external reviews to work: "<backend_name>_review"

### DIFF
--- a/edx_proctoring/rules.py
+++ b/edx_proctoring/rules.py
@@ -1,0 +1,16 @@
+"Django Rules for edx_proctoring"
+from __future__ import absolute_import
+
+import rules
+
+
+@rules.predicate
+def is_in_reviewer_group(user, attempt):
+    """
+    Returns whether user is in a group allowed to review this attempt
+    """
+    backend_group = '%s_review' % attempt['proctored_exam']['backend']
+    return user.groups.filter(name=backend_group).exists()
+
+
+rules.add_perm('edx_proctoring.can_review_attempt', is_in_reviewer_group | rules.is_staff)
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -86,7 +86,7 @@ Then back in your host shell::
     pip install -e .[server]
     python -m mockprock.server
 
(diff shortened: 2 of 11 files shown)


edx_proctoring/rules.py
```diff
@@ -0,0 +1,16 @@
+"Django Rules for edx_proctoring"
+from __future__ import absolute_import
+
+import rules
+
+
+@rules.predicate
+def is_in_reviewer_group(user, attempt):
```
MichaelRoytman: Does the documentation for user-based permissions cover testing predicates? Is there a way to do that? I realize that you're indirectly testing the `is_in_reviewer_group` predicate via the tests in `test_reviews.py`, but I wonder whether it's useful to have some unit testing of predicates.

matthugs: I think all the documentation we have on this is https://github.com/edx/edx-platform/blob/9816cadc7168b3d5320b92f68a0097dce7e713e5/lms/djangoapps/courseware/local_decisions/permissions.rst#convert-track-membership-tests-to-permissions

davestgermain: There's more documentation in [OEP-9](https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0009-bp-permissions.rst#rule-based-authorization), but it's true that there is a gap in the documentation around permissions, and it's still unclear how best to override permissions in edx-platform in a way that isn't confusing to developers.

MichaelRoytman:
Could you comment on `This will require a new django user group for external reviews to work: "<backend_name>_review"`? What needs to be done before/after this PR is merged?

davestgermain:
> Could you comment on `This will require a new django user group for external reviews to work: "<backend_name>_review"`? What needs to be done before/after this PR is merged?

For edx, [this PR](https://github.com/edx/app-permissions/pull/832) will have to be merged, which automatically adds the django group. 

Developers will have to create a django user group in the django admin, with a name matching the proctoring backend -- for instance "mockprock_review", and add the user account associated with the backend to the group. I'll update the developer doc.

PR #532, opened 2019-02-11
add createWorkers function to handle Workers with urls across origins

MichaelRoytman:


edx_proctoring/backends/rest.py
```diff
@@ -122,7 +122,7 @@ def get_javascript(self):
                 u'Webpack stats file corresponding to WebWorkers not found: {}'
                 .format(str(err))
             )
-        return js_url
+        return js_url + u'?raw'
```
MichaelRoytman: from here: https://github.com/edx/edx-proctoring/compare/matthugs/we-may-need-this-to-integrate-other-proctoring-backends

matthugs: I'm not sure I want to ship this without @davestgermain's input on it; the goal of this is to hit https://github.com/edx/edx-platform/blob/master/common/djangoapps/static_replace/__init__.py#L176 and avoid static rewriting of this URL by automated static URL rewriting infrastructure.

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -49,11 +49,29 @@ var edx = edx || {};
...
   };
 
+  function createWorker(url) {
+    var blob, blobBuilder, blobUrl, url;
+    var importScript = "importScripts('" + location.origin + url + "');";
+
+    try {
+      blob = new Blob([importScript], { "type": 'application/javascript' });
+    } catch (e) {
+      //older browsers do not support the Blob constructor, so fallback to the deprecated BlobBuilder
```
matthugs: On closer review it looks like Blob has plenty of support for our needs: https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#Browser_compatibility c.f. https://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/front_matter/browsers.html

We should be able to do without any try/catches, as it turns out

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -49,11 +49,29 @@ var edx = edx || {};
...
+      //interface in those cases
+      blobBuilder = new (window.BlobBuilder || window.WebKitBlobBuilder || window.MozBlobBuilder)();
+      blobBuilder.append(importScript);
+      blob = blobBuilder.getBlob('application/javascript');
+    }
+
+    url = window.URL || window.webkitURL;
+    blobUrl = url.createObjectURL(blob);
```
matthugs: Similarly here I don't think we need the `webkitURL` fallback https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -49,11 +49,29 @@ var edx = edx || {};
...
+    return new Worker(blobUrl);
+  }
 
   function workerPromiseForEventNames(eventNames) {
     return function() {
-      var proctoringBackendWorker = new Worker(edx.courseware.proctored_exam.configuredWorkerURL);
-      return new Promise(function(resolve, reject) {
+      var proctoringBackendWorker = createWorker(edx.courseware.proctored_exam.configuredWorkerURL);
+      return new Promise(function (resolve, reject) {
```
matthugs: nit (but would fail eslint if I hadn't dropped the ball on getting it all the way set up):
```suggestion
      return new Promise(function(resolve, reject) {
```

PR #534, opened 2019-02-12
Moved the enrollment mode check to a user permission

davestgermain:
The permission will be [defined and tested in edx-platform](https://github.com/edx/edx-platform/pull/19802).

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1305,19 +1305,6 @@ def get_active_exams_for_user(user_id, course_id=None):
     return result
 
 
-def _check_eligibility_of_enrollment_mode(credit_state):
-    """
-    Inspects that the enrollment mode of the user
-    is valid for proctoring
-    """
-
-    # Allow only the verified students to take the exam as a proctored exam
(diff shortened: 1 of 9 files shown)


edx_proctoring/api.py
```diff
@@ -1571,13 +1558,15 @@ def get_attempt_status_summary(user_id, course_id, content_id):
...
+        user = User.objects.get(id=user_id)
+        if not user.has_perm('edx_proctoring.can_take_proctored_exam', exam):
             return None
 
     attempt = get_exam_attempt(exam['id'], user_id)
     if attempt:
         status = attempt['status']
-    elif not exam['is_practice_exam'] and has_due_date_passed(credit_state.get('course_end_date', None)):
+    elif not exam['is_practice_exam'] \
+            and credit_state and has_due_date_passed(credit_state.get('course_end_date', None)):
```
matthugs: nit: doesn't pep-8 prefer implicit line continuation with parentheses? I'm surprised this passes linting

davestgermain: I don't know. I rely on the tool to scold me.

PR #537, opened 2019-02-13
prepend LMS Root URL to relative Javascript URLs if LMS Root URL is d…

MichaelRoytman:
…efined

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -11,6 +11,8 @@ import jwt
 from webpack_loader.utils import get_files
 from webpack_loader.exceptions import BaseWebpackLoaderException, WebpackBundleLookupError
 
+from django.conf import settings
+
 from edx_proctoring.backends.backend import ProctoringBackendProvider
 from edx_proctoring.exceptions import (
     BackendProviderCannotRegisterAttempt,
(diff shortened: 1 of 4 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -8,9 +8,12 @@
 import uuid
 import jwt
 
+from six.moves.urllib.parse import urlparse
 from webpack_loader.utils import get_files
 from webpack_loader.exceptions import BaseWebpackLoaderException, WebpackBundleLookupError
 
+from django.conf import settings
```
MichaelRoytman: I think I got this import order right? 

https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports

edx_proctoring/backends/rest.py
```diff
@@ -122,6 +124,13 @@ def get_javascript(self):
                 u'Webpack stats file corresponding to WebWorkers not found: {}'
                 .format(str(err))
             )
+
+        # if the Javascript URL is not an absolute URL (i.e. doesn't have a scheme), prepend
+        # the LMS Root URL to it, if it is defined, to make it an absolute URL
+        if not js_url.startswith('http') or not js_url.startswith('https'):
```
davestgermain: you don't need the https -- `startswith('http')` is close enough.

MichaelRoytman: lol, oops

PR #541, opened 2019-02-19
log a warning if review_by user does not have permission to submit a …

MichaelRoytman:
…review

### DIFF
--- a/edx_proctoring/tests/test_reviews.py
+++ b/edx_proctoring/tests/test_reviews.py
@@ -7,7 +7,7 @@ import json
 
 from crum import set_current_request
 import ddt
-from mock import patch
+from mock import patch, call
 
 from django.contrib.auth.models import User
 from django.test import RequestFactory
@@ -25,7 +25,7 @@ from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus, ReviewSta
 from edx_proctoring.tests.test_services import (MockCertificateService, MockCreditService, MockGradesService,
                                                 MockInstructorService)
 from edx_proctoring.utils import locate_attempt_by_attempt_code
-from edx_proctoring.views import ProctoredExamReviewCallback
+from edx_proctoring.views import ProctoredExamReviewCallback, is_user_course_or_global_staff
 
 from .utils import LoggedInTestCase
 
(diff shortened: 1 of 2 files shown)


edx_proctoring/views.py
```diff
@@ -126,6 +126,16 @@ def wrapped(request, *args, **kwargs):  # pylint: disable=missing-docstring
...
 
 
+def is_user_course_or_global_staff(user, course_id):
+    """
+        Return whether a user is course staff for a given course, described the course_id,
+        or is global staff.
+    """
+    instructor_service = get_runtime_service('instructor')
+
+    return not user.is_staff or not instructor_service.is_course_staff(user, course_id)
```
davestgermain: Why `not`? This returns the opposite of what you want.

MichaelRoytman: I forgot to push up the fix; thanks.

edx_proctoring/views.py
```diff
@@ -852,6 +862,15 @@ def make_review(self, attempt, data, backend=None):
         except (User.DoesNotExist, KeyError):
             review.reviewed_by = None
 
+        # If the reviewing user is a user in the system (user may be None for automated reviews) and does
+        # not have permission to submit a review, log a warning.
+        course_id = attempt['proctored_exam']['course_id']
+        if review.reviewed_by is not None and not is_user_course_or_global_staff(review.reviewed_by, course_id):
+            LOG.warning(
```
davestgermain: How about a test?

And I think the preferred way of formatting log strings (according to pylint?) is to have the logger do the format. So, `LOG.warning(msg, {'user': user, 'attempt': attempt_code})`

MichaelRoytman: For future reference: https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation.

edx_proctoring/tests/test_reviews.py
```diff
@@ -448,3 +448,70 @@ def test_status_reviewed_by_field(self):
         ProctoredExamReviewCallback().make_review(self.attempt, test_payload)
         review = ProctoredExamSoftwareSecureReview.objects.get(attempt_code=self.attempt['attempt_code'])
         self.assertEqual(review.reviewed_by, user)
+
+    def test_is_user_course_or_global_staff(self):
```
MichaelRoytman: This may be overkill. I'm not sure that it's best to test a utility function instead of just testing the function that uses it directly. However, I do think this is clearer than combining the cases in this test with the ones in `test_reviewed_by_is_course_or_global_staff`.

PR #542, opened 2019-02-22
Prevent JSON errors from causing an exception

davestgermain:
(This also changes the license for backends, to Apache 2.0)

### DIFF
--- a/edx_proctoring/backends/tests/test_rest.py
+++ b/edx_proctoring/backends/tests/test_rest.py
@@ -243,6 +243,18 @@ class RESTBackendTests(TestCase):
         self.assertEqual(status, corresponding_status)
 
     @responses.activate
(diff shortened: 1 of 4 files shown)


edx_proctoring/backends/tests/test_rest.py
```diff
@@ -242,6 +242,18 @@ def test_update_exam_attempt_status(self, provider_method_name, corresponding_st
         status = getattr(self.provider, provider_method_name)(self.backend_exam['external_id'], attempt_id)
         self.assertEqual(status, corresponding_status)
 
+    @responses.activate
+    def test_failed_json(self):
```
matthugs: "failed" is a little vague; could we say "malformed_json_in_response"?

PR #543, opened 2019-02-25
Change URL scheme check in get_javascript to use urlparse library

MichaelRoytman:


edx_proctoring/backends/rest.py
```diff
@@ -8,6 +8,9 @@
 import uuid
 import jwt
 
+# this is a known PyLint false positive (https://github.com/PyCQA/pylint/issues/1640)
+# that is fixed in PyLint v1.8
```
matthugs: Would the tools team be the right group to ask about how difficult it would be to upgrade pylint?

MichaelRoytman: I can bug them. I asked Cale about it and he said it would be a pain, but I'll double check.

matthugs: (for posterity) I'm satisfied with Cale's vibe

PR #545, opened 2019-02-26
The retire user endpoint should only make a retire request at backends s that have been used by that user

davestgermain:


edx_proctoring/views.py
```diff
@@ -1065,21 +1065,25 @@ def post(self, request, user_id):  # pylint: disable=unused-argument
...
         if not request.user.has_perm('accounts.can_retire_user'):
             return Response(status=403)
-        from django.apps import apps
-        choices = apps.get_app_config('edx_proctoring').get_backend_choices()
         results = {}
         code = 200
-        if ProctoredExamStudentAttempt.objects.filter(user_id=user_id):
-            for backend_name, verbose_name in choices:
+        seen = set()
+        exams = ProctoredExam.objects.filter(is_proctored=True, proctoredexamstudentattempt__user__id=user_id)
```
bmedx: I'm not sure what query this will shake out to exactly, but `is_proctored` doesn't seem to be indexed and I'm a little worried about the performance implications of using it this way. If there are generally few `ProctoredExamStudentAttempt`'s per user it might be better to pull them all with a `select_related` on `ProctoredExam` and just filter `is_proctored` in the Python.

davestgermain: Fair enough. I've changed it to iterate the attempts instead of the exams.

PR #547, opened 2019-02-28
Gate onboarding exams with the same user permission as proctored exams

davestgermain:
[EDUCATOR-4137](https://openedx.atlassian.net/browse/EDUCATOR-4137)

### DIFF
--- a/edx_proctoring/tests/__init__.py
+++ b/edx_proctoring/tests/__init__.py
@@ -3,6 +3,7 @@ Monkeypatches the default backends
 """
 from __future__ import absolute_import
 
+import contextlib
 import rules
 
 
@@ -20,6 +21,18 @@ def setup_test_backends():
     config.backends['mock'] = MockProctoringBackendProvider()
 
 
+@contextlib.contextmanager
+def mock_perm(perm='edx_proctoring.can_take_proctored_exam'):
+    """
+    Context manager for mocking a specific permission to return False inside the block
+    """
+    try:
+        rules.set_perm(perm, rules.always_false)
+        yield
+    finally:
+        rules.set_perm(perm, rules.always_true)
+
+
 def setup_test_perms():
     """
     Create missing permissions that would be defined in edx-platform,
@@ -30,11 +43,11 @@ def setup_test_perms():
     """
     try:
         rules.add_perm('accounts.can_retire_user', rules.is_staff)
-    except KeyError:
+    except KeyError:  # pragma: no cover
         pass
     try:
-        rules.add_perm('edx_proctoring.can_take_proctored_exam', rules.is_authenticated)
-    except KeyError:
+        rules.add_perm('edx_proctoring.can_take_proctored_exam', rules.always_true)
+    except KeyError:  # pragma: no cover
         pass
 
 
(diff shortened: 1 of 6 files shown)


edx_proctoring/tests/test_student_view.py
```diff
@@ -764,6 +764,14 @@ def test_get_studentview_new_onboarding(self):
         rendered_response = self.render_onboarding_exam()
         self.assertIn('Proctoring onboarding exam', rendered_response)
 
+    def test_get_onboarding_no_perm(self):
+        """
+        Test that onboarding exams are gated by the same permission as proctored exams
```
matthugs: Comment's misleading; the permissions can easily diverge with this test staying green.

edx_proctoring/tests/test_student_view.py
```diff
@@ -764,6 +764,14 @@ def test_get_studentview_new_onboarding(self):
...
+        """
+        with patch('django.contrib.auth.models.User.is_authenticated', False):
```
MichaelRoytman: Nit: My only thought is that I would prefer to mock the thing that's being tested directly, which is `user.has_perm('edx_proctoring.can_take_proctored_exam', exam)`.

What about something like this?
```
 def test_get_onboarding_no_perm(self):
        """
        Test that onboarding exams are gated by the same permission as proctored exams
        """
        with patch('django.contrib.auth.models.User.has_perm', return_value=False):
            rendered_response = self.render_onboarding_exam()
        self.assertIsNotNone(rendered_response)
```
This has the effect of making all user permissions return false, which would be a problem if we were to test something with multiple user permissions. It works for this case.

Or, to be more specific about the actual permission, but more verbose:
```
        from six import get_method_function
        from django.contrib.auth import get_user_model
        original_has_perm = get_method_function(get_user_model().has_perm)

        def side_effect(*args, **kwargs):
            if args[0] == 'edx_proctoring.can_take_proctored_exam':
                    return False
            else:
                # call the unpatched version
                return original_has_perm(get_user_model(), *args, **kwargs)

        with patch('django.contrib.auth.models.User.has_perm', side_effect=side_effect):
            rendered_response = self.render_onboarding_exam()
```

I like this because it's explicit, but, as it's written, it's not very reusable. I'm not able to get Python 3 working in edx_proctoring to test whether this works in Python 3 or not.

I'll copy the suggestion you came up with in Slack here as well.

```
@rules.predicate
def mock_yes(user, obj):
    return mock_yes.return_val
mock_yes.return_val = True


def setup_test_perms():
    """
    Create missing permissions that would be defined in edx-platform,
    or elsewhere.

    edx-platform imports tests from edx-proctoring, which causes duplicate
    rules, so ignore the KeyError thrown by the rules package.
    """
    try:
        rules.add_perm('accounts.can_retire_user', rules.is_staff)
    except KeyError:
        pass
    try:
        rules.add_perm('edx_proctoring.can_take_proctored_exam', rules.is_authenticated & mock_yes)
    except KeyError:
        pass
```

What do you think?

edx_proctoring/tests/__init__.py
```diff
@@ -20,6 +21,18 @@ def setup_test_backends():
     config.backends['mock'] = MockProctoringBackendProvider()
 
 
+@contextlib.contextmanager
+def mock_perm(perm='edx_proctoring.can_take_proctored_exam'):
```
davestgermain: @MichaelRoytman I think this approach satisfies your concern in a relatively simple way.

PR #551, opened 2019-03-11
add accessibility improvements to proctored exam code

MichaelRoytman:


edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -39,13 +39,14 @@ <h4>
...
-      <p>
+      <label for="proctored-exam-code">
         {% blocktrans %}
           Copy this unique exam code. You will be prompted to paste this code later before you start the exam.
         {% endblocktrans %}
-      </p>
+      </label>
     <p>
-      <span class="proctored-exam-code">{{exam_code}}</span>
+      <input id ="proctored-exam-code" class="proctored-exam-code" type="text" value="{{exam_code}}" readonly>
```
MichaelRoytman: The input cuts off the code at 20 characters. I don't think it's a huge deal, since the code will be auto-selected on click/focus, and there's the copy button there, but just wanted to mention that. I could always add a large size to the input.

MichaelRoytman:
Here's a screenshot of the changes:

image

MichaelRoytman:
I didn't find any existing tests for the Javascript on this page, so I have not added any tests. I am happy to add them, but it'll take a little bit more time.

edx_proctoring/templates/proctored_exam/instructions.html
davestgermain: Why not set the size exactly to 36?

matthugs: I think we should add a styling diff to clear this up (loath as I am to increase scope). From the screencap it looks like there's a gross white border-top on the button you added; we really oughta' get rid of that.

I expect connecting the button with the input (i.e. having them share a common border) would be a good way to make sure it's clear that the button and input correspond. It'd also make the button look more UI-ish to have it resemble our workflow CTAs less.

MichaelRoytman:
I added styling changes in this PR: https://github.com/edx/edx-platform/pull/19979.

For ease, here's what it looks like now:

image

edx_proctoring/templates/proctored_exam/instructions.html
MichaelRoytman: Just to tie up any loose ends:

I set the `size` attribute of the input element to 40. The font size is `1.3em`, and I wanted to keep the font size the same, so I increased the size instead.

I made some styling changes to the button here: https://github.com/edx/edx-platform/pull/19979.

PR #552, opened 2019-03-12
Expands sandbox setup documentation

matthugs:
Won't require a release! Or anything! What great fun.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -5,7 +5,7 @@ edX Proctoring Developer Guide
 
 
 How do I use proctoring on devstack?
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+------------------------------------
 * Create a test course
     * Follow the steps here: Including Proctored Exams in Your Course
 * Read the `learner guide for using proctoring <http://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/completing_assignments/SFD_proctored_exams.html>`_
@@ -18,7 +18,7 @@ How do I use proctoring on devstack?
     * "Proctored exam software secure review" has log of responses from SoftwareSecure
 
 How do I develop on edx-proctoring?
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+-----------------------------------
 
 These are the steps to install edx-proctoring into your pre-existing devstack image:
 
@@ -56,7 +56,7 @@ In edx-platform/lms/envs/private.py and edx-platform/cms/envs/private.py:
 
 
 Using mockprock as a backend
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+----------------------------
 
 `Mockprock <https://github.com/edx/mockprock>`_ is a proctoring backend that runs as an HTTP server and a python module. It allows you to simulate the entire proctoring workflow.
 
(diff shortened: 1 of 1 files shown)


docs/developing.rst
```diff
@@ -5,7 +5,7 @@ edX Proctoring Developer Guide
 
 
 How do I use proctoring on devstack?
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+------------------------------------
```
matthugs: By way of explanation, we skipped a section heading level I found I wanted to use

docs/developing.rst
```diff
@@ -119,8 +123,79 @@ Start by following the steps here: https://github.com/edx/edx-proctoring
...
+Proctortrack
+""""""""""""
+
+As will be the case with all REST backends implementing `our spec`_, one
+doesn't need to configure much to get Proctortrack working on a
+sandbox, e.g.::
+  EDXAPP_PROCTORING_BACKENDS:
+    DEFAULT: 'proctortrack'
+    proctortrack:
+      client_id: "<you'll need to fill this these in wth credentials from Proctortrack>"
```
MichaelRoytman:
```suggestion
      client_id: "<you'll need to fill this in with credentials from Proctortrack>"
```

docs/developing.rst
```diff
@@ -119,8 +123,79 @@ Start by following the steps here: https://github.com/edx/edx-proctoring
...
+      client_secret: "<you'll need to fill this these in wth credentials from Proctortrack>"
```
MichaelRoytman:
```suggestion
      client_secret: "<you'll need to fill this in with credentials from Proctortrack>"
```

docs/developing.rst
```diff
@@ -119,8 +123,79 @@ Start by following the steps here: https://github.com/edx/edx-proctoring
...
 
+Backend-specific Information
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+One of the main motivations for setting up a sandbox to test
+proctoring is having an externally accessible system which can be
+accessed by our proctoring providers' systems. This enables more
+thorough end-to-end testing.
+
+To enable proctoring in a way that won't be overridden by ansible
```
MichaelRoytman: After one sshing to the running sandbox and updating the yaml file, is any action required after that change?  Do you need to rerun ansible on the sandbox (e.g. as describe in the `Updating code` section of [this](https://openedx.atlassian.net/wiki/spaces/EdxOps/pages/13960183/Sandboxes) document)? Or does that occur as part of restarting the lms and cms processes?

docs/developing.rst
```diff
@@ -119,8 +123,79 @@ Start by following the steps here: https://github.com/edx/edx-proctoring
...
+  EDXAPP_PROCTORING_BACKENDS:
+    ...
+
+Placing these configurations here (rather than the more generic
+locations mentioned in `the README`_) will allow us to leverage the
+power of the ansible playbooks used to construct and administer
+sandboxes.
+
+The contents of ``EDXAPP_PROCTORING_BACKENDS`` will depend on which
+backend you're interested in testing. It's nececssary to provide a
```
MichaelRoytman:
```suggestion
backends you're interested in testing. It's necessary to provide a
```
Re: `backend` versus `backends`, the way this is phrased implies, to me, that the sandbox can be configured with only one backend. I think making backends plural clarifies that you can configure your sandbox with more than one backend, which may be a helpful distinction for people new to proctoring.

docs/developing.rst
matthugs: @MichaelRoytman do you think the update I put for this is enough?

MichaelRoytman: I think so! The only thing I notice is that the bullet points under `In addition to adding these configurations, you'll also need to set up
a user which PT can authenticate as.` aren't displaying correctly when viewing the file in GitHub.

matthugs: Good eye, thanks!

PR #553, opened 2019-03-14
allow status transitions from completed statuses to error

matthugs:
adds a separate mechanism to prevent submitted -> error transition from the JS ping check (since it may well happen if the learner has multiple tabs open each running our pinging JS)

Verified these transitions aren't prevented for existing RPNow integration via this splunk query (splunk is dumb and I can't share it as a report unfortunately): `index="prod-edx" source="/edx/var/log/lms/edx.log" | search "ProctoredExamIllegalStatusTransition: A status transition from" | rex field=_raw "ProctoredExamIllegalStatusTransition: A status transition from (?P<from_status>[a-z_]+) to (?P<to_status>[a-z_]+) was attempted" | fields from_status to_status | stats count(_raw) by from_status to_status`

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -824,21 +824,6 @@ def update_attempt_status(exam_id, user_id, to_status,
         )
         raise ProctoredExamIllegalStatusTransition(err_msg)
 
-    # special case logic, if we are in a completed status we shouldn't allow
-    # for a transition to 'Error' state
-    if in_completed_status and to_status == ProctoredExamStudentAttemptStatus.error:
-        err_msg = (
-            'A status transition from {from_status} to {to_status} was attempted '
-            'on exam_id {exam_id} for user_id {user_id}. This is not '
-            'allowed!'.format(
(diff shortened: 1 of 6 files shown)


edx_proctoring/api.py
```diff
@@ -824,21 +824,6 @@ def update_attempt_status(exam_id, user_id, to_status,
         )
         raise ProctoredExamIllegalStatusTransition(err_msg)
 
-    # special case logic, if we are in a completed status we shouldn't allow
```
MichaelRoytman: I understand the code change here, but I think I'm missing some context around what the rationale is more generally. Could you let me know if I'm understanding this correctly?

It was the case before that a user would submit their exam and their exam attempt status would thus move to `submitted`. After that, it was possible that we would receive a `not_reviewed` status in a callback from Proctortrack, if, for example, data upload failed. Without this change, we wouldn't be able to move the exam attempt into the `error` state successfully.

Is that right?

matthugs: Correct

PR #557, opened 2019-03-21
Remove AWS.py, update version

coryleeio:
New version required since we are phasing out aws.py


davestgermain:
I think you'll have to change the references [here](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/apps.py#L77) too.

coryleeio:
@davestgermain  Good catch ty - updated.

PR #559, opened 2019-03-28
Reintegrate RPNow using browser functionality

matthugs:
... to make way for switching to configuration fo RPNow which will
auto-close when recordings are unable to stream

JIRA:EDUCATOR-4216


edx_proctoring/callbacks.py
```diff
@@ -43,11 +42,15 @@ def start_exam_callback(request, attempt_code):  # pylint: disable=unused-argume
...
+    course_id = attempt['proctored_exam']['course_id']
+    content_id = attempt['proctored_exam']['content_id']
+
+    try:
+        exam_url = reverse('jump_to', args=[course_id, content_id])
+    except NoReverseMatch:
+        log.exception("Can't find course info url for course %s", course_id)
+        exam_url = ''
+    response = HttpResponseRedirect(exam_url)
+    response.set_signed_cookie('exam', attempt['attempt_code'])
```
matthugs: Note: I did not set a max time on the cookie as Dave suggested, since that way lies the same madness I was trying to solve as in https://openedx.atlassian.net/browse/EDUCATOR-4219

edx_proctoring/templates/proctored_exam/error_wrong_browser.html
```diff
@@ -0,0 +1,32 @@
...
+    {% blocktrans %}
+      Error with proctored exam
+    {% endblocktrans %}
+  </h3>
+
+  <p>
+    {% blocktrans %}
+      The content of this exam can only be viewed through the RPNow
+      application. If you have yet to complete your exam, please
+      return to RPNow application to proceed.
```
MichaelRoytman:
```suggestion
      return to the RPNow application to proceed.
```
Nit.

edx_proctoring/callbacks.py
```diff
@@ -43,11 +42,15 @@ def start_exam_callback(request, attempt_code):  # pylint: disable=unused-argume
...
-    template = loader.get_template('proctored_exam/proctoring_launch_callback.html')
-
-    return HttpResponse(
-        template.render({
-            'platform_name': settings.PLATFORM_NAME,
-            'link_urls': settings.PROCTORING_SETTINGS.get('LINK_URLS', {})
-        })
-    )
+
+    course_id = attempt['proctored_exam']['course_id']
```
MichaelRoytman: I can't leave this comment on the actual line since it hasn't been modified, but on line 23, there's a comment: `NOTE: This returns HTML as it will be displayed in an embedded browser`. It is no longer correct, since we're now redirecting to the exam URL instead of returning HTML. Could you either correct the comment or remove it, please?

edx_proctoring/tests/test_student_view.py
```diff
@@ -479,6 +490,19 @@ def test_get_studentview_started_practice_exam(self):
         rendered_response = self.render_practice_exam()
         self.assertIsNone(rendered_response)
 
+    @patch('edx_proctoring.api.get_backend_provider')
+    def test_get_studentview_practice_from_wrong_browser(self, mocked_get_backend):
+        """
+        Test for get_student_view proctored exam which has started.
```
MichaelRoytman:
```suggestion
        Test for get_student_view practice proctored exam which has started.
```

Nit.

PR #560, opened 2019-03-28
differentiate edX support email used in emails depending on the backend

MichaelRoytman:


docs/developing.rst
```diff
@@ -205,6 +205,7 @@ Comparably more is required for our older support of PSI's RemoteProctor NOW sof
       secret_key_id: "<secret>"
       secret_key: "<secret>"
       software_download_url: "http://edxdev.remoteproctor.com"
+      integration_specific_email: "proctortrack-support@edx.org"
```
davestgermain: shouldn't say proctortrack

MichaelRoytman: Oops, I put it in the wrong spot.

MichaelRoytman: Fixed!

docs/developing.rst
```diff
@@ -174,6 +174,7 @@ sandbox, e.g.::
       client_id: "<you'll need to fill these in with credentials from Proctortrack>"
       client_secret: "<you'll need to fill these in with credentials from Proctortrack>"
       base_url: 'https://prestaging.verificient.com'
+      integration_specific_email: "proctortrack-support@edx.org"
```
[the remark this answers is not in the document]
davestgermain: Oops, my bad.

edx_proctoring/tests/test_email.py
```diff
@@ -183,3 +186,39 @@ def test_not_send_email_timed_exam(self, status):
...
+        [ProctoredExamStudentAttemptStatus.rejected, ''],
+        [ProctoredExamStudentAttemptStatus.rejected, None],
+    )
+    @ddt.unpack
+    def test_correct_edx_email(self, status, integration_specific_email,):
+        exam_attempt = self._create_started_exam_attempt()
+
+        test_backend = get_backend_provider(name='test')
+
+        if integration_specific_email is None:
```
MichaelRoytman: I don't love this. Any ideas for a better way to test the case where a provider doesn't have an integration_specific_email attribute?

davestgermain: There's no need to delete it if you're going to set it to None in the next line. Otherwise, I don't have any suggestions to improve the testing situation. It seems clear enough what's going on.

MichaelRoytman: I have addressed your comment. Thank you for being patient with the hiccups.

matthugs: Does the `null` provider have a non-`None` email set?

MichaelRoytman: The `null` provider does not have an email set, but I think I'd rather explicitly remove the attribute from the test backend. I think it's a little more clear what's going on in that particular test case and doesn't require me to change [_create_proctored_exam](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/tests/utils.py#L208) to accept a backend.

matthugs: kk, when I posted this I hadn't seen dave's response. If it's an available option to just set it directly to None in all cases then there's no asymmetry to be smoothed out. Where I was on on the miscommunication train was the "but he must've already considered that?(!!?!!??)" caboose

PR #561, opened 2019-04-11
Clarify learner expectations around proctored exam workflow

matthugs:
This workflow has changed for RPNow4

JIRA:EDUCATOR-4217

### DIFF
--- a/edx_proctoring/templates/proctored_exam/instructions.html
+++ b/edx_proctoring/templates/proctored_exam/instructions.html
@@ -80,10 +80,9 @@
       </h4>
     <p>
         {% blocktrans %}
-          When you've finished the system check and verified your identity, begin your exam.
+          For security and exam integrity reasons, we ask you to sign in to your edX account. Then we will direct you to the RPNow proctoring experience.
         {% endblocktrans %}
     </p>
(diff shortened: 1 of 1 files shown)


edx_proctoring/templates/proctored_exam/instructions.html
```diff
@@ -80,10 +80,9 @@ <h4>
       </h4>
     <p>
         {% blocktrans %}
-          When you've finished the system check and verified your identity, begin your exam.
+          For security and exam integrity reasons, we ask you to sign in to your edX account. Then we will direct you to the RPNow proctoring experience.
         {% endblocktrans %}
     </p>
-    <button href="#" class="exam-action-button js-start-proctored-exam btn btn-secondary">{% trans "Start Exam" %}</button>
```
davestgermain: If I'm using the `null` backend, which is the default on dev, would I not be able to continue to the exam?

matthugs: Yeah this would break the null backend

matthugs: Oh wait no it wouldn't; the learner can refresh the page

PR #564, opened 2019-04-25
Unpin djangorestframework so we can upgrade edx-platform

nedbat:


nedbat:
@davestgermain Not sure if you like this way of fixing it.

davestgermain:
@nedbat good enough! I don't think anybody is depending on the exact wording of error messages from the proctoring REST API...

feanil:
Closing this in favor of #568

PR #565, opened 2019-04-26
Put RPNow4 changes behind a waffle flag

matthugs:
JIRA:EDUCATOR-4250

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -12,6 +12,8 @@ import uuid
 import pytz
 import six
 
+from waffle import switch_is_active
+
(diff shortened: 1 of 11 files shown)


edx_proctoring/api.py
```diff
@@ -12,6 +12,8 @@
 import pytz
 import six
 
+from waffle import switch_is_active
```
davestgermain: Aren't waffle switches deprecated?

matthugs: That's not my understanding of https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0017-bp-feature-toggles.rst#decision-map; I think our situation lines up pretty closely with #5.

PR #566, opened 2019-04-29
Use edx-when to handle due date extensions

davestgermain:
For [EDUCATOR-4247](https://openedx.atlassian.net/browse/EDUCATOR-4247).  Instead of relying on the due date stored in the exam model, we'll lookup the due date with edx-when, which will handle due date extensions. If the date isn't in edx-when, it'll fall back on the exam due date. 

This depends on the due date data being in edx-when's tables, which happens on course publish. 

You can run this to republish all courses: `./manage.py cms --settings=devstack_docker simulate_publish --skip-ccx --receivers edx_when.signals.extract_dates`


edx_proctoring/api.py
```diff
@@ -2003,13 +2002,9 @@ def get_student_view(user_id, course_id, content_id,
             return None
 
         # Just in case the due date has been changed because of the
-        # self-paced courses, use the due date from the context and
-        # update the local exam object if necessary.
+        # self-paced courses, use the due date from the context
+        # import pdb;pdb.set_trace()
```
matthugs:
```suggestion
```

davestgermain: > What's the migration plan for due dates into edx-when? 

The migration story for prod is in [DEVOPS-8619](https://openedx.atlassian.net/browse/DEVOPS-8619). For dev, there's no risk. Everything already falls back on the date in the modulestore. If someone is intending on developing date extensions, they'll have to republish the course (which happens if you change the due date in studio, anyway).

> 
> Also, wanted to flag that we'll end up in conflict between this and #559

I'll reversion this as 1.7.1

PR #567, opened 2019-04-29
Add migration for missing backend data

davestgermain:
[EDUCATOR-4277](https://openedx.atlassian.net/browse/EDUCATOR-4277)


edx_proctoring/migrations/0010_update_backend.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('edx_proctoring', '0009_proctoredexamreviewpolicy_remove_rules'),
+    ]
+
+    operations = [
+        migrations.RunPython(update_backend)
```
matthugs: can we add a second parameter of `migrations.RunPython.noop()` so it's explicit that we consider this irreversible/unnecessary to reverse?

davestgermain:
> Checklist for data migrations
> 
> * [ ]  Is there a rollback migration?  Does it correctly rollback to the previous state?

No

> * [ ]  Is there a migration test? Django migration testing

No

> * [ ]  How long does a rollback take to run?

It won't rollback

> * [ ]  Is data being loaded into this table? Is it static or dynamic? How long will it take to load?

The data is static and should be very quick to load. My bet is under 2 seconds.

PR #586, opened 2019-07-08
Fixed saving exams to backend when there's a due date set

davestgermain:
If there was a due date set on an exam, the `ProctoredExamSerializer` was not serializing the date into a string, causing `on_exam_saved` to fail silently. This meant that exams were not being saved on the backend, which would lead to an error like [EDUCATOR-4458](https://openedx.atlassian.net/browse/EDUCATOR-4458).

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -266,8 +266,11 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
             response = self.session.post(url, json=exam)
             data = response.json()
         except Exception as exc:  # pylint: disable=broad-except
-            # pylint: disable=no-member
-            content = exc.response.content if hasattr(exc, 'response') else response.content
+            if response:
+                # pylint: disable=no-member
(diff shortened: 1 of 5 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -266,8 +266,11 @@ def on_exam_saved(self, exam):
             response = self.session.post(url, json=exam)
             data = response.json()
         except Exception as exc:  # pylint: disable=broad-except
```
matthugs: could we branching below by tracing down the specific exception types that would be thrown here? I'm not sure we ever had a compelling reason to relax this lint rule, and it seems like it's related to how this issue came to be

davestgermain: It's a broad except because it's not clear which exceptions could be thrown. In this case, it was a `TypeError` from json serialization. An exception already would've been logged in splunk.

matthugs: Okay, could we add the branching logic as a more-specific except listed above this one and eliminate the `if`/`else` that way?

PR #597, opened 2019-08-30
add grace period checks in edx-proctoring

DawoudSheraz:


DawoudSheraz:
@davestgermain Tagging you here for the code review of this PR.

edx_proctoring/utils.py
```diff
@@ -231,3 +231,30 @@ def obscured_user_id(user_id, *extra):
...
+
+
+def verify_and_add_wait_deadline(context, exam_due_date):
+    """
+    Verify if the wait deadline should be added to template context.
+
+    If the grace period is present and is valid after the exam
+    has passed due date, add the wait deadline to context.
+    """
+    wait_deadline = exam_due_date + context['grace_period']
```
DawoudSheraz: To verify if the due date is present(not None).

edx_proctoring/tests/test_student_view.py
```diff
@@ -589,6 +590,116 @@ def test_get_studentview_submitted_timed_exam_with_past_due_date(self, due_date,
...
+        context.pop('wait_deadline')
+
+        context['grace_period'] = timedelta(days=graceperiod)
+        rendered_response = get_student_view(
+            user_id=self.user_id,
+            course_id=self.course_id,
+            content_id=self.content_id_for_exam_with_due_date,
+            context=context
+        )
+        if is_timed:
```
awaisdar001: Its not encouraging to have `if`/`else` in tests. Is there any particular reason why we can't have separate tests for timed exams?

edx_proctoring/api.py
```diff
@@ -1623,7 +1613,12 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
         # check if the exam's due_date has passed. If so, return None
         # so that the user can see their exam answers in read only mode.
         if not exam['hide_after_due'] and is_exam_passed_due(exam, user=user_id):
-            return None
+            if context.get('grace_period', False):
+                is_updated = verify_and_add_wait_deadline(context, exam['due_date'])
```
awaisdar001: what is this variable? `is_updated`?  sounds like it should be `due_date_passed`?

edx_proctoring/utils.py
```diff
@@ -231,3 +231,30 @@ def obscured_user_id(user_id, *extra):
...
+    return True if due date is lesser than current datetime, otherwise False
+    and if due_datetime is None then we don't have to consider the due date for return False
+    """
+
+    if due_datetime:
+        return due_datetime <= datetime.now(pytz.UTC)
+    return False
+
+
+def verify_and_add_wait_deadline(context, exam_due_date):
```
awaisdar001: nit: `exam_has_due_date`?

edx_proctoring/api.py
awaisdar001: nit?
```
if context.get('grace_period', False) and not exam_has_due_date(context, exam['due_date']):
    return
```

edx_proctoring/api.py
```diff
@@ -1963,6 +1958,12 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
             attempt['is_status_acknowledged'],
             exam
         ) else 'proctored_exam/verified.html'
+        if context.get('grace_period', False):
+            is_updated = verify_and_add_wait_deadline(context, exam['due_date'])
+            # The edge case where student has already acknowledged the result
+            # but the course team changed the grace period
+            if is_updated and not student_view_template:
+                student_view_template = 'proctored_exam/verified.html'
```
awaisdar001: Can we define this variable at the start of this method? 
`student_view_template = 'proctored_exam/verified.html'`

edx_proctoring/api.py
DawoudSheraz: `is_updated` signifies if the context was updated with `wait_deadline`. I think it should be renamed to `is_context_updated` to show context has been updated with something.

edx_proctoring/api.py
DawoudSheraz: This variable will hold either None or `verified.html` before entering into the if condition. Now, as I have mentioned that this is the edge case where the user, who has already acknowledged the status, will be able to see exam content if the course team decides to increase grace period. is_updated tells context was updated, which means wait period is there. not student_view_template means it has been acknowledged, but as we don't want the content to be seen, we override the value of template to be `verified.html`

edx_proctoring/tests/test_student_view.py
DawoudSheraz: Well, the entire unit test is the same for both proctored/timed exam, except for the last condition.

edx_proctoring/utils.py
DawoudSheraz: exam_has_due_date doesn't signify the work being done in the function. This function is basically confirming if the grace period is active and the waiting banner should be shown or not.

edx_proctoring/api.py
awaisdar001: and the context would only be updated there exam due date (due+grace period) correct?

edx_proctoring/api.py
awaisdar001: thank you for explaining the behavior here. This looks good as is, but I was having second thoughts if we could move the `'proctored_exam/verified.html'` in some enum variable.

edx_proctoring/api.py
DawoudSheraz: Well, yes. But naming it exam due date is not correct. This condition is for the cases where the exam submission exists. If a user tries to attempt after the due date, but with an active grace period, they will see the expired time banner.

edx_proctoring/tests/test_student_view.py
awaisdar001: Right.

DawoudSheraz: I can separate them out(which I had done initially), but as I said that there was only a difference of one line, I combined them together.

edx_proctoring/utils.py
```diff
@@ -249,12 +249,15 @@ def verify_and_add_wait_deadline(context, exam_due_date):
...
+    date is not present, which happens for self-paced courses, no context
+    update will take place.
     """
-    wait_deadline = exam_due_date + context['grace_period']
-    if not has_due_date_passed(wait_deadline):
-        context.update(
-            {'wait_deadline': wait_deadline.isoformat()}
-        )
-        return True
+    if exam_due_date:
```
awaisdar001:
```
if not exam_due_date:
    return False
```

edx_proctoring/api.py
```diff
@@ -1623,7 +1613,12 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
         # check if the exam's due_date has passed. If so, return None
         # so that the user can see their exam answers in read only mode.
         if not exam['hide_after_due'] and is_exam_passed_due(exam, user=user_id):
-            return None
+            if context.get('grace_period', False):
+                has_context_updated = verify_and_add_wait_deadline(context, exam['due_date'])
```
davestgermain: `exam['due_date']` won't be accurate, since it doesn't take into account due date extensions. You will have to call `get_exam_due_date(exam, user)` to get the right date.

I recommend refactoring `verify_and_add_wait_deadline()` so that it takes the context, exam, and user_id, and move `get_exam_due_date()` to utils.py. 

Instead of checking for the existence of grace_period here, check it in verify_and_add_wait_deadline.

DawoudSheraz: I will utilize the method you mentioned to cater to the individual extensions. As for the grace period check, I wrapped it around the function call to avoid unnecessarily call if the grace period is not found. But I think doing so will help avoid duplication.

DawoudSheraz:
@davestgermain Can you take a second look? I have updated both the code and the sandbox.

DawoudSheraz:
configparser 4.0.1 has been pulled from PyPI, as indicated in the PR on edx-platform https://github.com/edx/edx-platform/pull/21634. The test failures and the subsequent merge will take place after the version is back on the PyPI or we decide to downgrade.

PR #633, opened 2019-12-13
PROD-1089 Bug fix for Instructor dashboard redirect URL get API call

schenedx:
The problem is that, if the last exam within a course is not a proctored exam, then the InstructorDashboardView get call would return no proctored exam for the whole course. This is a bug detailed in PROD-1089

@edx/masters-devs Please review

### DIFF
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -5,6 +5,6 @@ The exam proctoring subsystem for the Open edX platform.
 from __future__ import absolute_import
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '2.1.9'
+__version__ = '2.2.0'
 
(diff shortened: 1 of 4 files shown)


edx_proctoring/__init__.py
```diff
@@ -5,6 +5,6 @@
 from __future__ import absolute_import
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '2.1.9'
+__version__ = '2.2.0'
```
iloveagent57: nit: point releases can go higher than 9 - unless you really do want a new minor version for this release.

schenedx: I agree, but I think it's ok to do a minor version here to keep this clean.

iloveagent57: 👍

PR #649, opened 2020-02-03
Added constraint to jsonfield2 in base.in

iamsobanjaved:


mraarif:
@awais786 I think we no longer need this PR.

PR #654, opened 2020-02-21
Don't allow transitions from ready_to_submit -> error.

iloveagent57:
Many learners appear to be landing in error at the time when they go to attempt to submit their exam, because this takes a while (i.e. the app can take a bit to respond/close normally).  Learners can exacerbate this problem if they get impatient and try to close the app manually after the app has already been triggered to close itself by edX’s browser JS.  This PR attempts to make sure learners who’re going to end the exam don’t have a chance to move into the error state when they've already clicked "end" from the edX side.

### DIFF
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
@@ -159,9 +159,16 @@ edx = edx || {};
             var pingInterval = self.model.get('ping_interval');
             self.timerTick += 1;
             self.secondsLeft -= 1;
+
+            // AED 2020-02-21:
+            // If the learner is in a state where they've finished the exam
+            // and the attempt can be submitted (i.e. they are "ready_to_submit"),
+            // don't ping the proctoring app (which action could move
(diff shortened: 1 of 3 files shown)


edx_proctoring/static/proctoring/js/views/proctored_exam_view.js
```diff
@@ -159,9 +159,16 @@ edx = edx || {};
...
+            // AED 2020-02-21:
+            // If the learner is in a state where they've finished the exam
+            // and the attempt can be submitted (i.e. they are "ready_to_submit"),
+            // don't ping the proctoring app (which action could move
+            // the attempt into an error state).
             if (
                 self.timerTick % pingInterval === pingInterval / 2 &&
-                edx.courseware.proctored_exam.configuredWorkerURL
+                edx.courseware.proctored_exam.configuredWorkerURL &&
+                this.model.get('attempt_status') !== 'ready_to_submit'
```
matthugs: nitpick if you wanna', I would prefer that we refer to this consistently as `self` here just so we don't have a difference without a distinction

PR #657, opened 2020-03-02
Add PII Retirement Code

nsprenkle:
JIRA: [EDUCATOR-4776](https://openedx.atlassian.net/browse/EDUCATOR-4776)
Linked PR: https://github.com/edx/tubular/pull/411

## Changelog
- Exposes new endpoint at `edx_proctoring/v1/retire_user/<user_id>/`
- Add personally-identifiable information (PII) retirement code to do the following
   - Obfuscate/hash user names and IP addresses logged with proctored exam attempts
   - Delete student allowance records

FYI @edx/masters-devs

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -2857,19 +2857,47 @@ class TestUserRetirement(LoggedInTestCase):
         super(TestUserRetirement, self).setUp()
         self.user.is_staff = True
         self.user.save()
-        self.second_user = User(username='tester2', email='tester2@test.com')
-        self.second_user.save()
+        self.user_to_retire = User(username='tester2', email='tester2@test.com')
+        self.user_to_retire.save()
         self.client.login_user(self.user)
 
-    def test_can_delete_user(self):
-        deletion_url = reverse('edx_proctoring:user_retirement_api', kwargs={'user_id': self.second_user.id})
+    def test_can_delete_user_no_data(self):
+        """ 
+        Attempting to retire an unknown user or user with no proctored attempts
+        returns 204 but does not carry out a retirment
+        """
+        deletion_url = reverse('edx_proctoring:user_retirement_api', kwargs={'user_id': self.user_to_retire.id})
         response = self.client.post(deletion_url)
+
         assert response.status_code == 204
-        # if there is no user data, then no deletion happens
-        assert response.data == {}
+
+    def test_can_delete_user(self):
+        """ Retiring a user should obfuscate PII for all exam attempts and return a 204 status """
+        proctored_exam = ProctoredExam.objects.create(
+                course_id='a/b/c',
+                content_id='test_content',
+                exam_name='Test Exam',
+                external_id='123aXqe3',
+                is_proctored=True,
+                is_active=True,
+                time_limit_mins=90,
+                backend='test'
+            )
+        create_exam_attempt(proctored_exam.id, self.user_to_retire.id)
+        attempt = ProctoredExamStudentAttempt.objects.filter(user_id=self.user_to_retire.id).first()
+
+        deletion_url = reverse('edx_proctoring:user_retirement_api', kwargs={'user_id': self.user_to_retire.id})
+        response = self.client.post(deletion_url)
+        retired_attempt = ProctoredExamStudentAttempt.objects.filter(user_id=self.user_to_retire.id).first()
+
+        assert retired_attempt.student_name != attempt.student_name
+        assert retired_attempt.last_poll_ipaddr != attempt.last_poll_ipaddr
+
+        assert response.status_code == 204
 
     def test_no_access(self):
-        self.client.login_user(self.second_user)
+        """ A user without retirement permissions should not be able to retire other users """
+        self.client.login_user(self.user_to_retire)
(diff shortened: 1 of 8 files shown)


edx_proctoring/views.py
```diff
@@ -1104,3 +1107,40 @@ def post(self, request, user_id):  # pylint: disable=unused-argument
...
+                attempt.save()
+
+        attempts_history = ProctoredExamStudentAttemptHistory.objects.filter(user_id=user_id)
+        if attempts_history:
+            for attempt_history in attempts_history:
+                attempt_history.student_name = obscured_user_id(attempt_history.student_name)
+                attempt_history.last_poll_ipaddr = obscured_user_id(attempt_history.last_poll_ipaddr)
+                attempt_history.save()
+
+    def _retire_user_allowances(self, user_id):
```
mattcarter: As we're obfuscating exam attempts, remain consistent and obfuscate the allowances too.

edx_proctoring/tests/test_views.py
```diff
@@ -2847,3 +2854,106 @@ def test_no_access(self):
...
+        assert response.status_code == 204
+
+        assert len(ProctoredExamStudentAllowance.objects.filter(user=self.user_to_retire.id)) == 0
+
+    def test_retire_user_allowances_history(self):
+        """ Retiring a user should delete their allowances and return a 204 """
+        proctored_exam = self._create_proctored_exam()
+        add_allowance_for_user(proctored_exam.id, self.user_to_retire.id, 'a_key', 30)
+        add_allowance_for_user(proctored_exam.id, self.user_to_retire.id, 'a_key', 60)
+        assert len(ProctoredExamStudentAllowanceHistory.objects.filter(user=self.user_to_retire.id)) == 1
```
matthugs: You can also use `.count()` on the query_set, rather than `len`in' it

edx_proctoring/views.py
```diff
@@ -1104,3 +1107,40 @@ def post(self, request, user_id):  # pylint: disable=unused-argument
...
+    """
+    Retire user personally-identifiable information (PII) for a user
+    """
+    def _retire_exam_attempts_user_info(self, user_id):
+        """ Obfuscate PII for exam attempts and exam history """
+        attempts = ProctoredExamStudentAttempt.objects.filter(user_id=user_id)
+        if attempts:
+            for attempt in attempts:
+                attempt.student_name = obscured_user_id(attempt.student_name)
+                attempt.last_poll_ipaddr = obscured_user_id(attempt.last_poll_ipaddr)
```
bmedx: This seems like a poorly named function if we're using it for non-id values. I'm not familiar enough with proctoring to know if it's an option, but edx-platform has special methods for hashing some values (username, email at least) that take advantage of the ability to rotate a configurable seed instead of relying on the Django secret key, which is a bigger deal to rotate. We've actually had to do this a couple of times. If you can use those, I'd suggest it.

edx_proctoring/views.py
```diff
@@ -1104,3 +1107,40 @@ def post(self, request, user_id):  # pylint: disable=unused-argument
...
+                attempt_history.save()
+
+    def _retire_user_allowances(self, user_id):
+        """ Delete user allowances """
+        ProctoredExamStudentAllowance.objects.filter(user=user_id).delete()
+        ProctoredExamStudentAllowanceHistory.objects.filter(user=user_id).delete()
+
+    def post(self, request, user_id):  # pylint: disable=unused-argument
+        """ Obfuscates all PII for a given user_id """
+        if not request.user.has_perm('accounts.can_retire_user'):
```
bmedx: In most, if not all, other places we allow admin users to make the retirement calls as well. Might be nice to add just for consistency.

edx_proctoring/tests/test_views.py
nsprenkle: Thanks @matthugs!

edx_proctoring/views.py
nsprenkle: @bmedx, this is outdated since we're now just clearing values. Was reusing as a convenient hashing function but I agree, maybe not the best way to have done it 👍

edx_proctoring/views.py
nsprenkle: @bmedx, do you have the permission string handy? I'm happy to add it.

bmedx: This actually looks like it's been cleaned up to happen automatically, so you should be set! https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/user_api/rules.py

edx_proctoring/views.py
```diff
@@ -1104,3 +1107,50 @@ def post(self, request, user_id):  # pylint: disable=unused-argument
...
+            for attempt_history in attempts_history:
+                attempt_history.student_name = ''
+                attempt_history.last_poll_ipaddr = None
+                attempt_history.save()
+
+    def _retire_user_allowances(self, user_id):
+        """ Clear user allowance values """
+        allowances = ProctoredExamStudentAllowance.objects.filter(user=user_id)
+
+        if allowances:
```
iloveagent57: Nit: this check isn't necessary (ditto the one below) - every QuerySet is iterable, even if it's empty.

PR #661, opened 2020-03-18
add a log immediately after exam attempt registration on the backend …

MichaelRoytman:
…to log when an a null external id has been returned, with no exception

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -690,6 +690,20 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
             )
             log.error(log_msg)
 
+    if not external_id and exam['backend'] == 'proctortrack' and taking_as_proctored:
+        # temporary logging added to better understand cases where we end up
+        # with exam attempts with NULL external_ids - a situation, that, theoretically,
+        # shouldn't be occurring (EDUCATOR-4948)
+        log_msg = (
+            u'Registered exam attempt with an external ID {external_id}'
+            u'in {exam_id} for user {user_id}'.format(
+                external_id=external_id,
+                exam_id=exam_id,
+                user_id=user_id,
+            )
+        )
+        log.error(log_msg)
+
     attempt = ProctoredExamStudentAttempt.create_exam_attempt(
         exam_id,
         user_id,
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -5,6 +5,6 @@ The exam proctoring subsystem for the Open edX platform.
 from __future__ import absolute_import
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '2.3.2'
+__version__ = '2.3.3'
 
 default_app_config = 'edx_proctoring.apps.EdxProctoringConfig'  # pylint: disable=invalid-name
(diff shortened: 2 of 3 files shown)


edx_proctoring/api.py
```diff
@@ -690,6 +690,20 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
             )
             log.error(log_msg)
 
+    if not external_id:
```
MichaelRoytman: I realize that there's similar logging a few lines down. But I'd like to get this log as close to the operation that generates the `external_id` as possible just to be absolutely sure that we wind up in a case with a falsy `external_id` here, because, we should be hitting the except clause for handling `BackendProviderSentNoAttemptID` if that's the case.

edx_proctoring/api.py
```diff
@@ -690,6 +690,20 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
+        # shouldn't be occurring (EDUCATOR-4948)
+        log_msg = (
+            u'Registered exam attempt with an external ID {external_id}'
+            u'in {exam_id} for user {user_id}'.format(
+                external_id=external_id,
+                exam_id=exam_id,
+                user_id=user_id,
+            )
+        )
+        log.info(log_msg)
```
schenedx: If you believe this is a case that shouldn't happen, I would log with error. So use `log.error` function instead of `info`

edx_proctoring/api.py
```diff
@@ -690,6 +690,20 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
             )
             log.error(log_msg)
 
+    if not external_id:
+        # temporary logging added to better understand cases where we end up
+        # with exam attempts with NULL external_ids - a situation, that, theoretically,
+        # shouldn't be occurring (EDUCATOR-4948)
+        log_msg = (
+            u'Registered exam attempt with an external ID {external_id}'
```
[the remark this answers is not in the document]
MichaelRoytman: I don't feel that this log should be permanent. If you look up a few lines at line 655, you'll see that we're handling an exception `BackendProviderSentNoAttemptID`, which is raised when we don't get back an `external_id` from the provider. This is the what is supposed to handle the error case of a null `external_id`, but, for some unknown reason, that exception handler is not being hit, which is troublesome. For that reason, this log just serves to aid in debugging why that's happening. It could be that there's another code path we haven't considered that results in no `external_id` or it could be something else. In either case, we don't know what the error is, so I don't think this is a permanent error detection. Does that make sense?

edx_proctoring/api.py
MichaelRoytman: That's fair. I'll change that.

edx_proctoring/api.py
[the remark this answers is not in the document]
MichaelRoytman: I elected to just print out the value of the `external_id` just to be absolutely sure of its value. Our API contract with the provider specifies that they must return an `external_id`, so there's no guarantees on what this value could be if it's not being passed through in their response. I want to be sure there's not some strange case where the provider is sending back a value we aren't handling correctly.

edx_proctoring/api.py
MichaelRoytman: To clarify, do practice exams come through with null `external_id`s for the exam or the exam attempt? If Proctrotrack has been returning null `external_ids` for the attempts, I would expect that we would see logging described in the exception handler for the `BackendProviderSentNoAttemptID` exception in Splunk, since there is no exception in that `register_exam_attempt` code for practice exams. I thought that the `external_id` here was referring to the `external_id` of the exam attempt and not the exam, because it's returned by the backend's `register_exam_attempt` method.

PR #665, opened 2020-04-09
ran Tox modernizer

mraarif:
relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1508).
Adding python3.8 in travis and removing django version prior to django2.2

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,96 +1,94 @@
 [tox]
-envlist =
-        py{35}-django{111,20,21,22}-drf{39,310,latest},
-        docs,
-        quality,
-        version_check,
-        pii_check,
-        translations
+envlist = 
+	py{35,38}-django{22,30}-drf{39,310,latest},
+	docs,
+	quality,
+	version_check,
+	pii_check,
+	translations
 
 [testenv]
-deps =
-    django111: Django>=1.11,<2.0
-    django20: Django>=2.0,<2.1
-    django21: Django>=2.1,<2.2
-    django22: Django>=2.2,<2.3
-    drf39: djangorestframework<3.10.0
-    drf310: djangorestframework<3.11.0
-    drflatest: djangorestframework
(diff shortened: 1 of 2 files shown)


mraarif:
`.travis.yml` needs to be updated according to changes in tox config.

awais786:
@iamsobanjaved looking good please squash.

tox.ini
```diff
@@ -1,96 +1,94 @@
...
-    django21: Django>=2.1,<2.2
-    django22: Django>=2.2,<2.3
-    drf39: djangorestframework<3.10.0
-    drf310: djangorestframework<3.11.0
-    drflatest: djangorestframework
-    -rrequirements/test.txt
-commands =
-    python -Wd -m pytest {posargs:-n 3}
+deps = 
+	django22: Django>=2.2,<2.3
```
jmbowman, line 12: Why did the whitespace change here?  I thought we'd been using 4-space indentation pretty consistently in tox.ini files.  Making it hard to scan and see what actually changed.

iamsobanjaved: script replaced 4-spaces indentation with tab, should i revert whitespace changes?

jmbowman: Yes, we definitely don't want tabs in these files; please update the script to use 4 space indentation instead.

mraarif: Correction: we didn’t add a leading tab, we should be able to update the constants DJANGO_22_DEPEDENCY and
DJANGO_30_DEPENDENCY to add leading spaces without making any other changes.

UsamaSadiq: As discussed in the group, the script does not make any change here. You can fix this by converting tabs to spaces in the file. In PyCharm, you can do this by selecting **edit->ConvertIndents->ToSpaces**.

iamsobanjaved:
@mzulqarnain1 has opened a new #667

PR #678, opened 2020-05-24
MST-225 Prevent the learner from accessing exams after they have an incomplete attempt status

schenedx:
MST-225

This change would prevent a learner from entering the exam again when:
1. The exam is past due 
2. Student already entered the exam but didn't finish

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1669,6 +1669,12 @@ def _get_timed_exam_view(exam, context, exam_id, user_id, course_id):
             student_view_template = 'timed_exam/expired.html'
         else:
             student_view_template = 'timed_exam/entrance.html'
+    elif is_exam_passed_due(exam, user_id) and ProctoredExamStudentAttemptStatus.is_incomplete_status(attempt_status):
+        # When the exam is past due, we should prevent learners from accessing the exam even if
+        # they already accessed the exam before, but haven't completed.
+        student_view_template = 'timed_exam/expired.html'
+    elif attempt_status == ProctoredExamStudentAttemptStatus.timed_out:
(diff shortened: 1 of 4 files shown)


edx_proctoring/api.py
```diff
@@ -1987,6 +1987,8 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
             # emit an event that the user was presented with the option
             # to start timed exam
             emit_event(exam, 'option-presented')
+    elif ProctoredExamStudentAttemptStatus.is_incomplete_status(attempt_status) and is_exam_passed_due(exam, user_id):
+        student_view_template = 'proctored_exam/expired.html'
```
matthugs, line 1991: should we make this same change for onboarding & timed exams as well?

schenedx: I don't believe we have due dates for onboarding exams. If you look at the function `_get_onboarding_exam_view`, it doesn't care about due dates.
It's a good feedback about timed_exams. I will make that change as well.

schenedx: @matthugs I added the same logic to timed_exam as well. Please review again

matthugs: > I don't believe we have due dates for onboarding exams.
That sounds like a wish. What do we have in place that brings that about? Should we create a ticket to give instructors feedback if they try to set due dates for onboarding exams? IIRC it has happened.

schenedx: I don't know if this should be a product requirement or not. I have other discovery tickets in our team's back log to figure out onboarding exam issues. I'll update on that. If we have due dates for onboarding exam, I don't believe we have logic to have onboarding exam do anything differently because the exam's due date is past.

PR #683, opened 2020-06-11
refactor email templates to support overriding and template inheritance

MichaelRoytman:
[MST-223](https://openedx.atlassian.net/browse/MST-223)

This refactors our email templates to support overriding key pieces of the templates by proctoring backend. The Django project that installs edx-proctoring (and the proctoring provider plugins) can override the entire template or leverage template inheritance to override certain blocks within the templates.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,6 +98,16 @@ In your lms.auth.json file, please add the following *secure* information::
 You will need to restart services after these configuration changes for them to
 take effect.
 
+Email Templates
+---------------
+
+edx-proctoring provides generic base email templates that are rendered and sent to learners based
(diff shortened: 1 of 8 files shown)


MichaelRoytman:
@schenedx My rationale was the edx-proctoring is meant to be backend agnostic. We don't really ever refer to proctortrack specifically, for example. Backend specific logic is implemented within the proctoring plugin (with the exception of software_secure). However, I did not think that email templates for emails that are sent by edX belonged within a proctoring plugin. edx-platform is the project that installs both edx-proctoring as a Django app and the proctortrack plugin, so I thought that it made sense to keep template overrides that are specific to edX's installation of those libraries in edx-platform. What do you think?

PR #689, opened 2020-07-01
Improve accuracy of proctortrack error message

how015:
MST-297

### DIFF
--- a/edx_proctoring/templates/proctored_exam/error.html
+++ b/edx_proctoring/templates/proctored_exam/error.html
@@ -9,7 +9,7 @@
   <p>
     {% blocktrans %}
       A technical error has occurred with your proctored exam. Please
-      reach out to <a href="https://courses.edx.org/support/contact_us">
+      reach out to <a href="{{link_urls.contact_us}}" target="_blank">
       edX Support</a> and your course team immediately for assistance.
     {% endblocktrans %}
   </p>
--- a/edx_proctoring/templates/proctored_exam/error.html
+++ b/edx_proctoring/templates/proctored_exam/error.html
@@ -8,11 +8,9 @@
 
   <p>
     {% blocktrans %}
-      A technical error has occurred with your proctored exam. All
-      exam data, including completed answers, has been lost. To
-      resolve this error, please
-      contact <a href="mailto:{{tech_support_email}}">technical
-      support</a>.
+      A technical error has occurred with your proctored exam. Please
+      reach out to <a href="https://courses.edx.org/support/contact_us">
+      edX Support</a> and your course team immediately for assistance.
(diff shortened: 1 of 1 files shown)


edx_proctoring/templates/proctored_exam/error.html
```diff
@@ -8,11 +8,9 @@ <h3>
 
   <p>
     {% blocktrans %}
-      A technical error has occurred with your proctored exam. All
-      exam data, including completed answers, has been lost. To
-      resolve this error, please
-      contact <a href="mailto:{{tech_support_email}}">technical
-      support</a>.
+      A technical error has occurred with your proctored exam. Please
+      reach out to <a href="https://courses.edx.org/support/contact_us">
```
MichaelRoytman, line 12: Looks good!

I think we shouldn't hardcode the contact URL here, because the link is environment and installation dependent, though. For example, this template would link to the production Contact Us form from Stage or from a sandbox. 

It looks like there's a `link_urls` dictionary passed to this template via the [context](https://github.com/edx/edx-proctoring/blob/b809ba23a525b76b5bf5d9119fbc753b20ee4895/edx_proctoring/api.py#L1832), which should include a `contact_us` link. You can see the `LINK_URLS` dictionary in our configuration [here](https://github.com/edx/edx-internal/blob/e1ff6f160c135856f1d8728a966c88baa0e73ba7/edx-remote-config/prod/lms.yml#L1471). You could also have a default in the case that this setting isn't set - something like [this](https://github.com/edx/edx-proctoring/blob/b809ba23a525b76b5bf5d9119fbc753b20ee4895/edx_proctoring/api.py#L1163).

how015: Thank you @MichaelRoytman, I found a similar case https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/templates/proctored_exam/ready_to_start.html#L28 and changed the PR accordingly.

edx_proctoring/templates/proctored_exam/error.html
```diff
@@ -8,11 +8,9 @@ <h3>
 
   <p>
     {% blocktrans %}
-      A technical error has occurred with your proctored exam. All
-      exam data, including completed answers, has been lost. To
-      resolve this error, please
-      contact <a href="mailto:{{tech_support_email}}">technical
-      support</a>.
+      A technical error has occurred with your proctored exam. Please
+      reach out to <a href="{{link_urls.contact_us}}" target="_blank">
```
how015, line 12: https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/templates/proctored_exam/ready_to_start.html#L28  is the solution I copied to here.

PR #694, opened 2020-07-14
proctoring system doc

Zacharis278:
@edx/masters-devs-cosmonauts 

Document to aid developers in navigating our proctoring system

PR diff is kinda useless so here's the rendered document: https://github.com/edx/edx-proctoring/blob/zhancock/proc-doc/docs/system-overview.rst


docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+.. image:: images/components.png
+
+Proctored Exam Views
+^^^^^^^^^^^^^^^^^^^^
+
+Interstitial views within a section that are shown to in place of the actual
+exam content. They are used to walk a leaner through setup steps
+and display the state of the current attempt when the exam is not in progress.
+
+The LMS will call into edx-proctoring to load the relevent template when rendering the
```
schenedx, line 21: This line is too verbose. How about
`During a proctored exam, LMS renders different student views based on existing exam status. The LMS renders the views by calling into edx-proctoring and load the relevent html templates`

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+Example worker: `proctortrack_custom.js <https://github.com/joshivj/edx-proctoring-proctortrack/blob/master/edx_proctoring_proctortrack/static/proctortrack_custom.js>`_
+
+edx-proctoring
+^^^^^^^^^^^^^^
+Python plugin that handles the bulk of edX's proctoring logic. It hosts the models for proctored
+exam configuration and learner attempts.  It exposes a REST and Python interface to manage them.
+edx-proctoring is also responsible for calling out to the provider's backend (through a plugin) to keep
+exam configuration and learner attempts in-sync between the two systems.
+
+https://github.com/edx/edx-proctoring/
```
schenedx, line 51: I don't feel like this link is useful. Maybe link the code to https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/views.py#L288 is better?

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+System Components
+------------------
+
+.. image:: images/components.png
+
+Proctored Exam Views
+^^^^^^^^^^^^^^^^^^^^
+
+Interstitial views within a section that are shown to in place of the actual
+exam content. They are used to walk a leaner through setup steps
```
alangsto, line 18: typo: leaner

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+
+System Components
+------------------
+
+.. image:: images/components.png
+
+Proctored Exam Views
+^^^^^^^^^^^^^^^^^^^^
+
+Interstitial views within a section that are shown to in place of the actual
```
alangsto, line 17: typo: "shown to learners" or "shown in place" ?

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+exam content. They are used to walk a leaner through setup steps
+and display the state of the current attempt when the exam is not in progress.
```
matthugs, line 19: I think you're separating two things between this and the following paragraph: the goal of the code and its implementation. That's a good approach for documentation. I would expand on it, though, and be more explicit about the separation you're making, e.g. 
> The goal of these interstitials is to guide the learner through the many steps of attempting a proctored exam, where at many steps there are critical details about what the learner needs to do or understand. As such some of the copy on these interstitials is very important.

docs/system-overview.rst
matthugs: I usually like to say that the intestitials are "driven by" the status of an attempt; as the attempt's status gets updated, we choose which interstitials to show/whether to show an interstitial at all. I think Simon's idea of separating the ideas of 1. calling into the edx-proctoring plugin to do this, from 2. the status column driving the choice of interstitial is a good one.

With that said I think there are a lot of valid ways of phrasing this. Just trying to give perspective of ways I've explained it in the past that seemed helpful for folks (although maybe it was just helpful for me!)

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+
+The LMS will call into edx-proctoring to load the relevent template when rendering the
+student view for an exam section that is proctored.
+
+Notable Code:
+
+- `LMS render student_view() <https://github.com/edx/edx-platform/blob/a7dff8c21ee794e90bdc0f22876334a7843a032d/common/lib/xmodule/xmodule/seq_module.py#L274>`_
+- `edx-proctoring template logic <https://github.com/edx/edx-proctoring/blob/78976d93ab6ca5206f259dc420d2f45818fe636c/edx_proctoring/api.py#L1912>`_
+- `edx-proctoring interstitial templates <https://github.com/edx/edx-proctoring/tree/master/edx_proctoring/templates>`_
```
matthugs, line 28: I would tie this to a specific commit hash also, even if it seems unlikely that these would move.

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+exam states. These messages may be handled by a JS worker specific to the proctoring provider. 
+That worker is included as part of the the provider's python plugin and will 
+interface directly with the proctoring software running on the learner's machine. Exam
+state will transition forward only after these messages have been successfully handled.
+However, if a provider does not have a worker interface configured there will be no direct
+communication between edX courseware and the proctoring client.
+
+Message Interface: `exam_action_handler.js <https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/static/proctoring/js/exam_action_handler.js>`_
+
+Example worker: `proctortrack_custom.js <https://github.com/joshivj/edx-proctoring-proctortrack/blob/master/edx_proctoring_proctortrack/static/proctortrack_custom.js>`_
```
matthugs, line 42: I think this section should probably be lower in this document, certainly below the "states" section/diagram. The whole proctoring JS hullabaloo is only relevant to a small subset of attempt statuses (so we should reference what those statuses are in this section), and is specific to rest-based backends (i.e. doesn't apply to RPNow). These are important distinctions to try to reflect in the structure of this document.

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+Exam States
+-----------
+When a learner first enters a proctored exam subsection an exam attempt is created
+in the edX system. User actions and the proctoring provider will update the status of
+this attempt as the exam is completed and reviewed. The following diagram describes the 
+flow through those status updates.
+
+Note this figure does not include error states or display of unmet prerequite requirements
+
+.. image:: images/attempt_states.png
```
matthugs, line 76: this diagram also assumes RPNow, which should be explicitly called out/corrected

Adapting it to be more general might just be as simple as removing the reference to zendesk tickets. I just shared you the source of the chart in lucidchart: https://app.lucidchart.com/documents/edit/2a1cb6e9-f61f-4c0c-b0d0-7164424563f6/Xcy7oxO_ZdV~

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Example Action Sequence
+-------------------------
+
+The diagram below describes the happy-path of interactions between components to 
+sucessfully begin a proctored exam.
+
+.. image:: images/sequence.png
```
matthugs, line 84: This looks very similar to a sequence diagram I made when implementing this. Sent you a share to that also; it's less concise compared to this.

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+For testing backends see `mockprock <https://github.com/edx/edx-proctoring/blob/master/docs/developing.rst#using-mockprock-as-a-backend>`_
+
+Exam States
+-----------
+When a learner first enters a proctored exam subsection an exam attempt is created
+in the edX system. User actions and the proctoring provider will update the status of
+this attempt as the exam is completed and reviewed. The following diagram describes the 
+flow through those status updates.
+
+Note this figure does not include error states or display of unmet prerequite requirements
```
matthugs, line 74: nit: incomplete sentence

May also be a good opportunity to link to a more exhaustive list of states in other existing documentation: https://github.com/edx/edx-documentation/blob/master/en_us/course_authors/source/proctored_exams/rpnow_results.rst#values-in-the-status-column

docs/system-overview.rst
```diff
@@ -0,0 +1,84 @@
...
+This document outlines the components involved in the edX proctoring system. It should
+serve as a guide to orient developers on what components exist, how they interact, and 
+where to find the source code in our platform.
+
+.. contents::
+
+System Components
+------------------
+
+.. image:: images/components.png
```
matthugs, line 12: This seems like a great diagram to put on a slide & talk through, but I don't think it's very self-explanatory. Also several of the features pictured don't exist in RPNow; it's implicitly PT-centric. (So it may make sense to move below where you introduce the prospect of different backends and add some blurb about that.)

An additional comment about the image itself: I feel like "LMS/CMS" & "edx-platform" should probably be reversed in their positions. edx-platform & edx-proctoring are both chunks of code that both run on the LMS, which is an application. So I think the nesting is a little backwards there. Might also make sense to have a different type of box (e.g. a different color) for non-codebase things; the LMS/CMS is not a codebase, as the legend might suggest, nor is the learner's web browser (which we might consider labelling as such, rather than "(web) edX Courseware").

docs/system-overview.rst
Zacharis278: thanks, I've generalized the review part for either

docs/system-overview.rst
Zacharis278: I've updated this and attempted to make it a bit less PT centric.  I do like having it up front because I think the visual gives context to each of the components explained below.

Zacharis278:
@matthugs thanks for the feedback.  I've made a bunch of changes to try to make the PT/RPNow distinction a bit clearer but this still leans into the proctortrack implementation as 'default'.  I'll merge later today but feel free to tag me on any updates you think we need.

PR #705, opened 2020-09-04
MST-406 Do not create exam attempt on past due exams when taking as proctoring

schenedx:

This is the bug fix for [MST-406](https://openedx.atlassian.net/browse/MST-406).
We are not going to create an exam attempt if the exam is taking as proctored and past due.
We can return `None` attempt ID because everywhere we call `create_exam_attempt`, we would try to update its attempt status and then we would throw the proper exception case for frontend to handle.

Also refactored the function a bit for better readability

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -564,6 +564,104 @@ def _create_and_decline_attempt(exam_id, user_id):
     )
 
 
+def _register_proctored_exam_attempt(user_id, exam_id, exam, attempt_code, review_policy):
+    """
+    Call the proctoring backend to register the exam attempt. If there are exceptions
+    the external_id returned might be None. If the backend have onboarding status errors,
+    it will be returned with force_status
+    """
+    scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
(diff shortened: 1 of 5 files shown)


alangsto (approved):
LGTM

edx_proctoring/api.py
```diff
@@ -594,113 +692,26 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
-                external_id=external_id,
-                exam_id=exam_id,
-                user_id=user_id,
+            u'user_id {user_id} trying to create exam atttempt for past due proctored exam {exam_id} '
+            u'Do not register an exam attempt! Return None'.format(
+                exam_id=exam_id, user_id=user_id
             )
         )
         log.error(log_msg)
+        return None
```
matthugs, line 709: I would prefer that we threw an exception here, since we know we're in an error situation at this point. I think this is using a return value for some error cases, and exceptions for others. I would prefer if we were consistent (especially in the absence of a note in the docstring).

schenedx: @matthugs Got it. I updated the code according to your feedback and now we are raising exception.

PR #711, opened 2020-10-05
MST-436 Cleanup the RPNowV4 Waffle Flag

schenedx:

This is cleaning up the waffle flag `edx_proctoring.rpnowv4_flow` because this is rolled out completed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,7 @@
 We don't maintain a detailed changelog.  For details of changes, please see
 the `GitHub commit history`_.
 
+ - 2.4.7
+   - Removed the rpnowv4_flow waffle flag to cleanup code  
+
 .. _GitHub commit history: https://github.com/edx/edx-proctoring/commits/master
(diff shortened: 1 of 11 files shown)


edx_proctoring/tests/test_views.py
```diff
@@ -1773,8 +1773,7 @@ def _setup_for_test_exam_callback(self):
 
         return response_data['exam_attempt_id']
 
-    @patch('edx_proctoring.callbacks.switch_is_active')
-    def test_exam_callback(self, mock_switch_is_active):
+    def test_exam_callback_rpnowv4(self):
```
Zacharis278, line 1776: I think this should just stay 'test_exam_callback' since it isn't considered specific to rpnowv4 anymore.

CHANGELOG.rst
```diff
@@ -1,4 +1,7 @@
 We don't maintain a detailed changelog.  For details of changes, please see
 the `GitHub commit history`_.
 
+ - 2.4.6
```
Zacharis278, line 4: 🎉  Nice, we can come up with some instructions and a format for maintaining this file but good to at least get started.

Zacharis278 (commented):
Changes look good I think there's another location in the tests using this flag that might need to be cleaned up: https://github.com/edx/edx-proctoring/blob/5942cf2a9d315490ae9086fe4f49ed4fb95168bf/edx_proctoring/tests/test_views.py#L487

Zacharis278 (approved):

PR #714, opened 2020-10-16
Update proctoring error message for inactive account

alangsto:
## [MST-401](https://openedx.atlassian.net/browse/MST-401)

@edx/masters-devs-cosmonauts 

Updated error messaging for inactive accounts, and added new interstitial for users with inactive accounts. This change prevents users from entering a proctored exam if their account has not been activated. 

Example of what new interstitial looks like: 

(Your Platform Name Here will be replaced with whatever platform name is defined in Settings)
Screen Shot 2020-10-19 at 12 13 15 PM

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,11 +14,17 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.4.8] - 2020-10-19
+~~~~~~~~~~~~~~~~~~~~
+
+* Created a separate error message for inactive users. Refined the
+  existing error message to only show for network error or service disruption.
+
 
 [2.4.7] - 2020-10-06
 ~~~~~~~~~~~~~~~~~~~~
 
-* Removed the rpnowv4_flow waffle flag to cleanup code  
+* Removed the rpnowv4_flow waffle flag to cleanup code
 
 For details of changes prior to this release, please see
 the `GitHub commit history`_.
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1802,6 +1802,18 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
     """
     Common context variables for the Proctored and Practice exams' templates.
     """
+    password_url = ''
+    try:
+        password_assistance_url = reverse('password_assistance')
+        scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
+        password_url = '{scheme}://{site_name}{password_assistance_url}'.format(
+            scheme=scheme,
+            site_name=constants.SITE_NAME,
+            password_assistance_url=password_assistance_url
+        )
+    except NoReverseMatch:
+        log.exception(u"Can't find password reset link")
+
     has_due_date = exam['due_date'] is not None
     attempt_time = attempt.get('allowed_time_limit_mins', None) if attempt else None
 
@@ -1849,6 +1861,7 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
         'learner_notification_from_email': provider.learner_notification_from_email,
         'integration_specific_email': get_integration_specific_email(provider),
         'exam_display_name': exam['exam_name'],
+        'reset_link': password_url
     }
     if attempt:
         context['exam_code'] = attempt['attempt_code']
@@ -1870,13 +1883,17 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
     """
     Returns a rendered view for the practice Exams
     """
+    user = USER_MODEL.objects.get(id=user_id)
+
     student_view_template = None
 
     attempt = get_exam_attempt(exam_id, user_id)
 
     attempt_status = attempt['status'] if attempt else None
 
-    if not attempt_status:
+    if not user.is_active:
+        student_view_template = 'proctored_exam/inactive_account.html'
+    elif not attempt_status:
         student_view_template = 'practice_exam/entrance.html'
     elif attempt_status == ProctoredExamStudentAttemptStatus.started:
         provider = get_backend_provider(exam)
@@ -1919,7 +1936,9 @@ def _get_onboarding_exam_view(exam, context, exam_id, user_id, course_id):
 
     attempt_status = attempt['status'] if attempt else None
 
-    if not attempt_status:
+    if not user.is_active:
+        student_view_template = 'proctored_exam/inactive_account.html'
+    elif not attempt_status:
         student_view_template = 'onboarding_exam/entrance.html'
     elif attempt_status == ProctoredExamStudentAttemptStatus.started:
         # when we're taking the exam we should not override the view
@@ -1968,7 +1987,9 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
     if attempt_status == ProctoredExamStudentAttemptStatus.declined:
         return None
 
-    if not attempt_status:
+    if not user.is_active:
+        student_view_template = 'proctored_exam/inactive_account.html'
+    elif not attempt_status:
         # student has not started an attempt
         # so, show them:
         #       1) If there are failed prerequisites then block user and say why
--- a/edx_proctoring/templates/practice_exam/entrance.html
+++ b/edx_proctoring/templates/practice_exam/entrance.html
@@ -74,10 +74,11 @@
         ).fail(function(){
           enableClickEvent();
           var msg = gettext(
-            "There has been a problem starting your exam.\n\n" +
-            "Possible reasons are that your account has not been fully activated,\n" +
-            "you have are experiencing a network connection problem, or there has been\n" +
-            "a service disruption. Please check these and try again."
+            "There has been a problem starting your exam.\n\n " +
+            "Possible reasons are that you are experiencing a network connection problem, " +
+            "or there has been a service disruption. Please make sure that you are connected to a stable " +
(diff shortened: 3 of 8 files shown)


edx_proctoring/api.py
```diff
@@ -1870,32 +1874,39 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
...
-        student_view_template = 'proctored_exam/ready_to_start.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.error:
-        student_view_template = 'practice_exam/error.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.submitted:
-        student_view_template = 'practice_exam/submitted.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.ready_to_submit:
-        student_view_template = 'proctored_exam/ready_to_submit.html'
+    active_account = user.is_active
+
+    if not active_account:
```
alangsto, line 1887: I also added this interstitial for practice exams, because I thought that it could be applied here as well, although the ticket only specified onboarding and proctored exams

edx_proctoring/api.py
```diff
@@ -1802,6 +1802,9 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
     """
     Common context variables for the Proctored and Practice exams' templates.
     """
+    scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
+    reset_link = '{scheme}://{hostname}/reset'.format(scheme=scheme, hostname=settings.SITE_NAME)
```
matthugs, line 1806: can we use `reverse('password_assistance')` instead?

edx_proctoring/api.py
matthugs: That makes sense. Limited specification is that there's substantially less sensitivity about practice exams, so they're less of concern.

edx_proctoring/api.py
```diff
@@ -1870,32 +1874,39 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
...
-        student_view_template = 'proctored_exam/instructions.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.ready_to_start:
-        student_view_template = 'proctored_exam/ready_to_start.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.error:
-        student_view_template = 'practice_exam/error.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.submitted:
-        student_view_template = 'practice_exam/submitted.html'
-    elif attempt_status == ProctoredExamStudentAttemptStatus.ready_to_submit:
-        student_view_template = 'proctored_exam/ready_to_submit.html'
+    active_account = user.is_active
```
matthugs, line 1885: I prefer to name variables that contain boolean values with a verb, to help myself remember that they're boolean. I think about what yes/no question they answer: Does this user have an active account? Then `user_has_active_account` should be true.

edx_proctoring/api.py
```diff
@@ -1870,32 +1874,39 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
...
+                                ProctoredExamStudentAttemptStatus.download_software_clicked]:
+            student_view_template = 'proctored_exam/instructions.html'
+        elif attempt_status == ProctoredExamStudentAttemptStatus.ready_to_start:
+            student_view_template = 'proctored_exam/ready_to_start.html'
+        elif attempt_status == ProctoredExamStudentAttemptStatus.error:
+            student_view_template = 'practice_exam/error.html'
+        elif attempt_status == ProctoredExamStudentAttemptStatus.submitted:
+            student_view_template = 'practice_exam/submitted.html'
+        elif attempt_status == ProctoredExamStudentAttemptStatus.ready_to_submit:
+            student_view_template = 'proctored_exam/ready_to_submit.html'
```
matthugs, line 1909: This looks equivalent to a slightly longer chain of `elif`s, which would help to reflect the way in which the change you're making follows the existing pattern. The added indentation emphasizes the case you're adding unnecessarily.

matthugs (changes requested):

alangsto:
@matthugs Thank you so much for the feedback. I took in your suggestions

edx_proctoring/api.py
```diff
@@ -1802,6 +1802,19 @@ def _get_proctored_exam_context(exam, attempt, user_id, course_id, is_practice_e
...
     Common context variables for the Proctored and Practice exams' templates.
     """
+    password_assistance_url = ''
+    try:
+        password_assistance_url = reverse('password_assistance')
+    except NoReverseMatch:
+        log.exception(u"Can't find password reset link")
+
+    scheme = 'https' if getattr(settings, 'HTTPS', 'on') == 'on' else 'http'
+    password_url = '{scheme}://{site_name}{password_assistance_url}'.format(
```
schenedx, line 1812: Shall we update the `password_url` to be `''` when password_assistant_url is empty? Your logic would print out an invalid URL.

edx_proctoring/api.py
```diff
@@ -1870,13 +1884,19 @@ def _get_practice_exam_view(exam, context, exam_id, user_id, course_id):
...
+    user = USER_MODEL.objects.get(id=user_id)
+
     student_view_template = None
 
     attempt = get_exam_attempt(exam_id, user_id)
 
     attempt_status = attempt['status'] if attempt else None
 
-    if not attempt_status:
+    user_has_active_account = user.is_active
```
schenedx, line 1895: #NIT why create this variable when you can just do `not user.is_active` below? This line seems redundant.

edx_proctoring/api.py
```diff
@@ -1919,7 +1939,11 @@ def _get_onboarding_exam_view(exam, context, exam_id, user_id, course_id):
...
```
schenedx, line 1942: Seems redundant. Remove?

edx_proctoring/api.py
```diff
@@ -1963,12 +1987,16 @@ def _get_proctored_exam_view(exam, context, exam_id, user_id, course_id):
 
     attempt_status = attempt['status'] if attempt else None
 
+    user_has_active_account = user.is_active
```
schenedx, line 1990: Seems redundant. Remove?

edx_proctoring/templates/practice_exam/entrance.html
```diff
@@ -74,10 +74,10 @@ <h3>
         ).fail(function(){
           enableClickEvent();
           var msg = gettext(
-            "There has been a problem starting your exam.\n\n" +
-            "Possible reasons are that your account has not been fully activated,\n" +
-            "you have are experiencing a network connection problem, or there has been\n" +
-            "a service disruption. Please check these and try again."
+            "There is a network problem or service connection issue.\n\n " +
```
schenedx, line 77: The message update here removes the scenario where `service disruption` is a possible. Can we add that back in?
In the event of service disruption, like what we are experiencing with Proctortrack right now, showing this message makes our learners think there is something wrong with their environment.

CHANGELOG.rst
```diff
@@ -14,13 +14,18 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.4.8] - 2020-10-19
+~~~~~~~~~~~~~~~~~~~~
+
+* Updated error messaging for inactive accounts
```
schenedx, line 20: A more accurate changelog statement would be
` Create a separate error message for inactive users. Refined the existing error message to only for network error or service disruption.`

schenedx (changes requested):
See my feedback inline.

edx_proctoring/templates/practice_exam/entrance.html
alangsto: Thanks for catching this, I had copied what was in the ticket but I agree that this wording should be kept in.

alangsto:
@schenedx thank you for your feedback as well, I also incorporated your suggestions

edx_proctoring/templates/proctored_exam/inactive_account.html
```diff
@@ -0,0 +1,30 @@
...
+  <p>
+    {% blocktrans %}
+      Your {{platform_name}} account has not yet been activated. To take the proctored exam,
+      you are required to activate your account.
+    {% endblocktrans %}
+  </p>
+  <p>
+    {% blocktrans %}
+      Please check your registered email's Inbox and Spam folders for an activation email from
+      {{platform_name}}. If you cannot find this email, you can
```
schenedx, line 18: Can we do an if statement on `reset_link` not being empty?

schenedx (approved):
Just a NIT. Rest LGTM

edx_proctoring/templates/practice_exam/entrance.html
```diff
@@ -74,10 +74,11 @@ <h3>
...
           enableClickEvent();
           var msg = gettext(
-            "There has been a problem starting your exam.\n\n" +
-            "Possible reasons are that your account has not been fully activated,\n" +
-            "you have are experiencing a network connection problem, or there has been\n" +
-            "a service disruption. Please check these and try again."
+            "There has been a problem starting your exam.\n\n " +
+            "Possible reasons are that you are experiencing a network connection problem, " +
+            "or there has been a service disruption. Please make sure that you are connected to a stable " +
+            "connection, and try to find a location with a higher bandwidth if possible. " +
```
matthugs, line 80:
```suggestion
            "connection, and try to find a location with higher bandwidth if possible. " +
```

matthugs (approved):
🚀  this is so exciting!

PR #720, opened 2020-11-10
Allow practice exam reset

Zacharis278:
@edx/masters-devs-cosmonauts 

**Description:**

Adds a new action to exam attempt PUT to reset a practice exam back to the created state.

**JIRA:**

[MST-473](https://openedx.atlassian.net/browse/MST-473)

**Pre-Merge Checklist:**

- [x] ~Updated the version number in `edx_proctoring/__init__.py` and `package.json` if these changes are to be released.

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1224,7 +1224,7 @@ def reset_practice_exam(exam_id, user_id):
     Resets a completed practice exam attempt back to the created state.
     """
     log_msg = (
-        u'Resetting practice exam {exam_id} for user {user_id}'.format(
+        'Resetting practice exam {exam_id} for user {user_id}'.format(
             exam_id=exam_id,
             user_id=user_id,
         )
@@ -1232,9 +1232,10 @@ def reset_practice_exam(exam_id, user_id):
     log.info(log_msg)
 
     exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_exam_attempt(exam_id, user_id)
-    exam = get_exam_by_id(exam_id)
     if exam_attempt_obj is None:
         raise StudentExamAttemptDoesNotExistsException('Error. Trying to look up an exam that does not exist.')
+
+    exam = get_exam_by_id(exam_id)
     if not exam['is_practice_exam']:
         msg = (
             'Failed to reset attempt status on exam_id {exam_id} for user_id {user_id}. '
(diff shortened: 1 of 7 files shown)


edx_proctoring/api.py
```diff
@@ -1219,6 +1219,55 @@ def _get_email_template_paths(template_name, backend):
     return [base_template]
 
 
+def reset_practice_exam(exam_id, user_id):
+    """
+    Resets a completed practice exam attempt back to the created state.
+    """
+    log_msg = (
+        u'Resetting practice exam {exam_id} for user {user_id}'.format(
```
idegtiarov, line 1227: #nit Explicit Unicode prefix looks redundant since we on a py3. 
And the question, do we aim to support py3.5? I'd like to start using f-strings ))), it looks like koa release would be on py3.8

edx_proctoring/api.py
```diff
@@ -1219,6 +1219,55 @@ def _get_email_template_paths(template_name, backend):
...
+            exam_id=exam_id,
+            user_id=user_id,
+        )
+    )
+    log.info(log_msg)
+
+    exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_exam_attempt(exam_id, user_id)
+    exam = get_exam_by_id(exam_id)
```
idegtiarov, line 1235: #nit I would define `exam` variable after checking that `exam_attempt_obj` is not None to avoid database requests in the worse case when the object is not found.

edx_proctoring/api.py
```diff
@@ -1219,6 +1219,55 @@ def _get_email_template_paths(template_name, backend):
...
+            'Failed to reset attempt status on exam_id {exam_id} for user_id {user_id}. '
+            'Only practice exams may be reset!'.format(
+                exam_id=exam_id,
+                user_id=user_id,
+            )
+        )
+        raise ProctoredExamIllegalStatusTransition(msg)
+
+    # prevent a reset if the exam is currently in progress
+    in_incompleted_status = ProctoredExamStudentAttemptStatus.is_incomplete_status(exam_attempt_obj.status)
```
idegtiarov, line 1249: typo? `s/in_incompleted_status/is_incompleted_status/

idegtiarov (commented):
@Zacharis278 some nits, comments, and questions, if you don't mind.

edx_proctoring/api.py
Zacharis278: edx-platform still does support 3.5 at the moment so tentatively yes. I'll have to get back to you on when we can drop support in this library specifically.

schenedx (approved):
👍

PR #726, opened 2020-12-03
Add endpoint for onboarding status

bseverino:
**Description:**

Adds endpoint to expose onboarding status to learner.

**JIRA:**

[MST-516](https://openedx.atlassian.net/browse/MST-516)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.5.1] - 2020-12-10
+~~~~~~~~~~~~~~~~~~~~
+
+* Add endpoint to expose the learner's onboarding status
+
 [2.5.0] - 2020-12-09
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -3,6 +3,6 @@ The exam proctoring subsystem for the Open edX platform.
 """
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '2.5.0'
+__version__ = '2.5.1'
 
 default_app_config = 'edx_proctoring.apps.EdxProctoringConfig'  # pylint: disable=invalid-name
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -4,6 +4,7 @@ Proctored Exams HTTP-based API endpoints
 
 import json
 import logging
+from itertools import chain
 
 import waffle
 from crum import get_current_request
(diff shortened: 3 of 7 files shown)


edx_proctoring/views.py
```diff
@@ -286,6 +286,72 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+            return Response(
+                data='There is no onboarding exam related to this course id.',
+                status=404
+            )
+
+        # Get the most recent onboarding attempt + past attempts
+        recent_attempt = ProctoredExamStudentAttempt.objects.filter(
+            proctored_exam__course_id=course_id,
+            proctored_exam__is_practice_exam=True,
+            user__username=request.user.username,
```
schenedx, line 326: Instead of the using the `request.user`, I think the endpoint will be more useful if the username or the user_id is passed in from the URL. Make the URL like `/edx_proctoring/v1/onboarding_status/<user_id>` would work.

edx_proctoring/views.py
```diff
@@ -286,6 +286,72 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+    **Response Values**
+        * 'onboarding_status': string specifying the learner's onboarding status;
+            will return NULL if there are no onboarding attempts
+        * 'onboarding_link': link to the onboarding exam
+    """
+    def get(self, request):
+        """
+        HTTP GET handler. Returns the learner's onboarding status.
+        """
+        course_id = request.GET.get('course_id')
```
schenedx, line 308: If course_id is required on the view, you should incorporate it into the URL. But is course_id required? Can the API return a dictionary of all the onboarding status of all the courses the learner is with? Also, I thought onboarding status within Verificient should be course independent, meaning, if a user got onboarding approved for course A, they are also approved for course B, right?
If course_id is not required, I would agree you should keep that as the query string parameter, but you would need better code handling for no course_id case, and I want to see that covered by your unit tests.

edx_proctoring/views.py
```diff
@@ -286,6 +286,72 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+            proctored_exam__course_id=course_id,
+            proctored_exam__is_practice_exam=True,
+            user__username=request.user.username,
+            taking_as_proctored=True,
+        ).order_by('-started_at')
+
+        # Default to the most recent attempt if there are no verified attempts
+        relevant_attempt = recent_attempt
+        for attempt in past_attempts:
+            if attempt.status == ProctoredExamStudentAttemptStatus.verified:
```
schenedx, line 339: Can you double check in production that all onboarding exam with Software Secure have attempt status to be `verified`? If they don't have that status, which status do those have? I want to make sure this endpoint works the same in the case of Software Secure practice exams.

edx_proctoring/views.py
```diff
@@ -286,6 +286,72 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
         return Response(data)
 
 
+class StudentOnboardingStatusView(ProctoredAPIView):
+    """
+    Endpoint for the StudentOnboardingStatusView
+
+    Supports:
+        HTTP GET: returns the learner's onboarding status relative to the given course_id
```
schenedx, line 294: I can see we need a new endpoint to retrieve the onboarding status for a single user regardless of their course. I also believe the endpoint needs permission logic to ensure, either the request user is a course admin user, edX staff user, or the passed in user is the same user on the request. All other request should be denied.
Please make sure you setup your unit tests for these 4 permission cases.

schenedx (changes requested):
See inline comments. Let me know if any of these doesn't make sense.

edx_proctoring/views.py
Zacharis278: Yes onboarding status with Verificient doesn't have to be course dependent but I think we should still treat it that way on our end right now. The user experience for onboarding is still tied to a single course and I think allowing that state to carry over between courses may lead to some confusion. (ex. a rejected onboarding in a different course) However, I like this endpoint design because if we go down that path in a next iteration we don't have to create a new endpoint, we can just drop the query parameter.

Simplest solution IMO is just have `course_id` be a mandatory query parameter right now. (and a test covering this)

edx_proctoring/views.py
```diff
@@ -286,6 +286,72 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+                status=404
+            )
+
+        # Get the most recent onboarding attempt + past attempts
+        recent_attempt = ProctoredExamStudentAttempt.objects.filter(
+            proctored_exam__course_id=course_id,
+            proctored_exam__is_practice_exam=True,
+            user__username=request.user.username,
+            taking_as_proctored=True,
+        ).first()
```
Zacharis278, line 328: We should be using .latest('modified') here. This is just going to grab the attempt with the lowest primary key.

edx_proctoring/views.py
bseverino: I added logic to pass in a username, and to deny access if the username doesn't match the logged in user + the user is also not course/edX staff. If a username isn't given, the endpoint just returns the logged in user's own status.

edx_proctoring/views.py
bseverino: I left the course_id as a required query parameter, but also added logic so that it can be easily removed as a requirement in the future (IE if a course_id isn't given, the onboarding link will just be returned as None)

edx_proctoring/views.py
bseverino: It should be the same: https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/proctored_exams/rpnow_results.html#values-in-the-status-column

edx_proctoring/views.py
```diff
@@ -286,6 +286,98 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+
+        attempt_filters = {'proctored_exam__is_practice_exam': True, 'taking_as_proctored': True}
+
+        username = request.GET.get('username')
+        course_id = request.GET.get('course_id')
+
+        if username:
+            # Check that the user is staff if trying to view another user's status
+            if username != request.user.username:
+                if course_id:
```
idegtiarov, line 326: I would prefer to avoid code duplication and have one exit from the condition statement as it is totally equal.

```
if username and username != request.user.username:
    if course_id and not is_user_course_or_global_staff(request.user, course_id) or not request.user.is_staff:
         return Response(
             status=status.HTTP_403_FORBIDDEN,
             data={'detail': _('Must be a Staff User to Perform this request.')
        )
```
Please ignore if a comment doesn't make sense to you.

edx_proctoring/views.py
```diff
@@ -286,6 +286,98 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+                            data={'detail': _('Must be a Staff User to Perform this request.')}
+                        )
+                elif not request.user.is_staff:
+                    return Response(
+                        status=status.HTTP_403_FORBIDDEN,
+                        data={'detail': _('Must be a Staff User to Perform this request.')}
+                    )
+            attempt_filters['user__username'] = username
+
+        if not course_id:
```
idegtiarov, line 339: I would check mandatory parameters first so we do not need to do redundant work. Could that block be moved to the L323?

bseverino: Thanks for the review! Your suggestions make sense, I will implement them.

edx_proctoring/__init__.py
```diff
@@ -3,6 +3,6 @@
 """
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '2.4.9'
+__version__ = '2.5.0'
```
schenedx, line 6: I think you need this to be 2.5.1 now

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.5.0] - 2020-12-03
```
schenedx, line 17: This will be 2.5.1

schenedx (approved):
👍 LGTM. You need to address the merge conflicts.

Zacharis278 (approved):

Zacharis278:
@bseverino can you also bump the version of this in edx-platform so this gets deployed? (example https://github.com/edx/edx-platform/pull/25817)

PR #725, opened 2020-11-30
Add reset button to onboarding exams

Zacharis278:
@edx/masters-devs-cosmonauts 

**Description:**

Adds a button allowing learners to restart an onboarding exam from the rejected, verified, and submitted states. This feature is enabled by a context flag passed down by the LMS. See https://github.com/edx/edx-platform/pull/25704.

Leaving this as draft to gather some feedback on the copy/UX while I look into some potential backend issues with resetting a 'submitted' but not reviewed attempt.

**JIRA:**

[MST-474](https://openedx.atlassian.net/browse/MST-474)

**Pre-Merge Checklist:**

- [x] Updated the version number in `edx_proctoring/__init__.py` and `package.json` if these changes are to be released.
- [x] Described your changes in `CHANGELOG.rst`
- [x] Confirmed Github reports all automated tests/checks are passing.
- [x] Approved by at least one additional reviewer.

**Post-Merge:**

- [ ] Create a tag matching the new version number.

PR #728, opened 2020-12-07
Practice Exam Reset Updates

Zacharis278:
**Description:**

This updates the practice exam reset action to create a new attempt instead of rolling back the existing attempt's state.  In the special case that a submitted (but not reviewed) attempt is reset we will remove that attempt to prevent it from getting reviewed and creating a disjoint state between verificient's system and the lms.

This PR removes the constraint that a user may only have one exam attempt. However, validation will only allow this to occur for practice exams at this time.

@edx/masters-devs-cosmonauts 

**JIRA:**

[MST-473](https://openedx.atlassian.net/browse/MST-473)

**Pre-Merge Checklist:**

- [ ] Updated the version number in `edx_proctoring/__init__.py` and `package.json` if these changes are to be released. (to be released with #725)

**Post-Merge:**


edx_proctoring/tests/test_views.py
```diff
@@ -1213,21 +1214,25 @@ def test_reset_attempt_action(self):
...
-                    'action': action,
-                }),
-                content_type='application/json'
-            )
+        # reject exam, then attempt to reset progress
+        set_runtime_service('grades', MockGradesService(rejected_exam_overrides_grade=False))
+        update_attempt_status(
+            proctored_exam.id, self.student_taking_exam.id, ProctoredExamStudentAttemptStatus.rejected
+        )
+        response = self.client.put(
```
schenedx, line 1222: Does this mean it's frontend's job to reset the existing attempt into `reset_attempt` action?

edx_proctoring/api.py
```diff
@@ -1246,28 +1242,29 @@ def reset_practice_exam(exam_id, user_id):
...
         raise ProctoredExamIllegalStatusTransition(msg)
 
-    exam_attempt_obj.status = ProctoredExamStudentAttemptStatus.created
-    exam_attempt_obj.started_at = None
-    exam_attempt_obj.completed_at = None
-    exam_attempt_obj.allowed_time_limit_mins = None
-    exam_attempt_obj.save()
+    # resetting a submitted attempt that has not been reviewed will entirely remove that submission.
+    if exam_attempt_obj.status == ProctoredExamStudentAttemptStatus.submitted:
+        remove_exam_attempt(exam_attempt_obj.id, requesting_user)
```
schenedx, line 1260: When this attempt is removed, how do we remove it from Verificient? Is that implemented? If so, is this properly implemented for RPNowV4 exams?

schenedx (commented):
Several questions:
1. What is the effect of this change on RPNowV4 backend proctored exams? Would that also results in multiple attempts on the practice exams?
2. I didn't see the logic in the python code to set the existing verified exam attempt to another status. Does this mean the new attempt will also become verified once created? For example, the learner already have a verified onboarding profile with the previous exam attempt. Then the user practice on the exam again. We create a new attempt. Does this new attempt become verified right away? If not, if learner practice the third time, it would remove the second attempt, right? Would that be desired?

edx_proctoring/api.py
Zacharis278: there's a signal that calls out to the backend interface to delete. In the case of RPNowV4 this is a no-op (we don't delete attempts with them)

https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/signals.py#L94

edx_proctoring/tests/test_views.py
Zacharis278: yup

Zacharis278:
@schenedx to answer your questions

1. On the front end we're only calling this from onboarding specifically. However if we wanted to call it for a RPNow exam I don't see any issue here since afaik they only track by distinct attempts and not by user+exam. To their system this workflow would actually look the same as the current flow of deleting the attempt and starting a new one.
2. For this PR you cannot reset a verified exam because it creates a bunch of complications as you mentioned. I plan to log another story to track this. I don't think it's a necessary requirement for the problem on reducing blockers tho.

edx_proctoring/api.py
```diff
@@ -679,18 +679,15 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
-        else:
-            err_msg = (
-                u'Cannot create new exam attempt for exam_id = {exam_id} and '
-                u'user_id = {user_id} because it already exists!'
-            ).format(exam_id=exam_id, user_id=user_id)
+    existing_attempt = ProctoredExamStudentAttempt.objects.get_current_exam_attempt(exam_id, user_id)
+    # only practice exams may have multiple attempts
+    if existing_attempt and not existing_attempt.is_sample_attempt:
+        err_msg = (
+            u'Cannot create new exam attempt for exam_id = {exam_id} and '
```
idegtiarov, line 686: #nit `u` prefix could be omitted

idegtiarov (commented):
One minor nit other looks good. Thanks.

schenedx (approved):
👍

idegtiarov (approved):
👍

PR #735, opened 2020-12-18
Save remaining time after attempt enters error state

bseverino:
**Description:**

Updated the attempt model to include a `time_remaining_seconds` field, which saves the remaining time after the attempt enters an `error` state.

**JIRA:**

[MST-528](https://openedx.atlassian.net/browse/MST-528)

**Pre-Merge Checklist:**

- [x] ~~Updated the version number in `edx_proctoring/__init__.py` and `package.json` if these changes are to be released.~~

**Post-Merge:**

- [x] ~~Create a tag matching the new version number.~~


edx_proctoring/models.py
```diff
@@ -349,6 +349,10 @@ class ProctoredExamStudentAttempt(TimeStampedModel):
     # else always false
     is_status_acknowledged = models.BooleanField(default=False)
 
+    # if the attempt enters an error state, the remaining time should
+    # be saved in order to allow the learner to resume
+    time_remaining_seconds = models.FloatField(null=True)
```
matthugs, line 354: why a float? This seems like a value that would be very well modeled by an integer

bseverino: It was to keep it consistent with the value returned here: https://github.com/edx/edx-proctoring/blob/85a51a3ea05f9555556a1b43bc9a4c7057f36172/edx_proctoring/utils.py#L37

Although I see how it could be confusing. I have no problem changing it.

schenedx: I think keep this as an integer sounds better. Float just complicates code everywhere and we don't need that level of detail.

schenedx (changes requested):
Could you please specify where the newly added `time_remaining_seconds` gets used? Is it on the frontend JS? Is it on any of the backend logic?

bseverino:
@schenedx It isn't currently being used anywhere, but I assume it will be used as part of [MST-530](https://openedx.atlassian.net/browse/MST-530) and [MST-531](https://openedx.atlassian.net/browse/MST-531) to allow the learner to resume an exam after it errors out (so probably just backend logic).

schenedx (approved):
👍 LGTM

PR #745, opened 2021-01-06
[MST-531] Allow creation of multiple active exam attempts

bseverino:
@edx/masters-devs-cosmonauts 

**Description:**

The `create_exam_attempt` function has been modified to allow multiple active attempts, as long as all other active attempts are in the `ready_to_resume` or `resumed` state. This saves the remaining time left in the new attempt, which is used to calculate the remaining time instead of the initial time limit.

**JIRA:**

[MST-531](https://openedx.atlassian.net/browse/MST-531)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.5.7] - 2021-01-08
+~~~~~~~~~~~~~~~~~~~~
+* Allow the creation of multiple exam attempts for a single user in a single exam, as long
+  as the most recent attempt is `ready_to_resume` or `resumed`. When an exam is resumed, the
+  time remaining is saved to the new attempt and is used to calculate the expiration time.
+
 [2.5.6] - 2021-01-06
 ~~~~~~~~~~~~~~~~~~~~
 * Updated the StudentProctoredExamAttempt view's PUT handler to allow for a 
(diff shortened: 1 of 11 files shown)


schenedx (commented):
A way to make your change here easier to read is to make your branch PR against Michael's branch at `mroytman/MST-530-modify-attempt-endpoint`. This way, when Michael's PR is merged, you can just rebase your branch against master and create a new PR there.
That said, upon review your WIP logic here, I cannot see anything needs suggestion.

schenedx (changes requested):
@bseverino Can we have more test cases?
in the test_api.py, can we add test case to create multiple resumed attempts for the same exam and user, then do a get API call to get all attempts successfully? And see if we are able to get all the resumed attempts?
Can we have similar test cases in test_views for a user who created an exam attempt, then update the exam attempt to error, then resumed the attempt, and getting the two attempts successfully?

Zacharis278 (approved):
lgtm

alangsto (changes requested):
I think that this function https://github.com/edx/edx-proctoring/blob/0cf3076c4ed9cd9e6e37ea30936bf0ff89a4e2d4/edx_proctoring/api.py#L1879 may also need to be updated, so that the remaining time is accounted for in the exam length (so that the remaining time is actually enforced on the attempt)

bseverino:
@alangsto Thank you for catching that!

alangsto (approved):

schenedx (approved):
👍

PR #748, opened 2021-01-11
Added is_attempt_active field to review models

alangsto:
**Description:**

## [MST-586](https://openedx.atlassian.net/browse/MST-586)

Added an `is_attempt_active` field to ProctoredExamSoftwareSecureReview and ProctoredExamSoftwareSecureReviewHistory model to note if the attempt for that review is no longer active

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/migrations/0013_auto_20210111_1728.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 2.2.17 on 2021-01-11 17:28
```
Zacharis278, line 1: can you add a descriptive filename to this migration?

alangsto: Yes, thank you for catching that!

schenedx:
@alangsto Could you please check with Support team to hash out the implication of this change? This change means, when an attempt is deleted, their review results are also deleted, which means we won't have a tracking history. Would that be OK? Or should we use this opportunity to create some history that are not deleted? We can also add into the histories to specify reset happened?

Zacharis278:
Similar feedback to @schenedx, can you clarify why do we need to have cascade on history table rows as well to solve this?

CHANGELOG.rst
```diff
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.5.8] - 2021-01-12
+~~~~~~~~~~~~~~~~~~~~
+* Added a foreign key constraint to the ProctoredExamSoftwareSecureReview
+  and ProctoredExamSoftwareSecureReviewHistory models so that reviews are
```
schenedx, line 20: Does history table also gets deleted?

edx_proctoring/migrations/0013_proctoredexamsoftwaresecurereview_attempt_code_constraint.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('edx_proctoring', '0012_proctoredexamstudentattempt_time_remaining_seconds'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
schenedx, line 14: I noticed this just now. This is going to cause production issue during deployment. See https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtorenameacolumn

schenedx (changes requested):
I think that rename column migration might cause issues. Could you please double check?

edx_proctoring/admin.py
```diff
@@ -271,6 +271,7 @@ def student_username_for_review(self, obj):
         'exam_name_for_review',
         'student_username_for_review',
         'attempt_code',
+        'is_attempt_active',
```
alangsto, line 274: I'm also realizing that if we don't want to add a new field, we could also just display the info on the LMS Admin page. Not sure if it is better to actually add it to the model so it's also visible in snowflake or just add it here so that support can see it.

schenedx: I rather have this info recorded in the database because this way, we keep our business logic simple, and then make data query easier

edx_proctoring/migrations/0014_update_is_attempt_active.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            ProctoredExamAttempt.objects.get(attempt_code=review_object.attempt_code)
+        except ProctoredExamAttempt.DoesNotExist:
+            review_object.is_attempt_active = False
+            review_object.save()
+
+    ProctoredExamReview = apps.get_model('edx_proctoring', 'ProctoredExamSoftwareSecureReview')
+    ProctoredExamReviewHistory = apps.get_model('edx_proctoring', 'ProctoredExamSoftwareSecureReviewHistory')
+
+    for review in ProctoredExamReview.objects.filter(is_attempt_active=True):
```
schenedx, line 23: This migration might run pretty long time. Could you first get to know how many review we will be iterating over? Also, please print out each Review record changed so we know that the migration is moving along successfully?

schenedx (approved):
Just 1 simple suggestion. Also, could you please update the PR title? That is now incorrect.
Could you put in PR description how many Review records will the migration be iterating through in production?

PR #758, opened 2021-01-20
Allow blank fields for ProctoredExam in Django admin

bseverino:
@edx/masters-devs-cosmonauts 

**Description:**

This fixes an issue where `ProctoredExam` allows the fields `external_id`, `due_date`, and `backend` to be null in the database, but they cannot be left blank in Django admin.

**JIRA:**

[MST-609](https://openedx.atlassian.net/browse/MST-609)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.5.12] - 2021-01-20
+~~~~~~~~~~~~~~~~~~~~~
+* Allow blank fields in Django admin for `external_id`, `due_date`, and `backend`
+  in proctored exams.
+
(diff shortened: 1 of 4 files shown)


matthugs (changes requested):
I am hoping that we can address this without a migration by customizing the form instead & setting https://docs.djangoproject.com/en/2.2/ref/forms/fields/#django.forms.Field.required to False

schenedx (dismissed):
I agree with @matthugs

matthugs (approved):
I think this may do one of two things on submitting the django admin form:
- it might save these fields as the empty string, even though we didn't change the contents of the fields and they were null before
- it might leave them as null

I think we're fine to go forward with this even without clarity on that point because I don't think it will matter in these use cases.

PR #759, opened 2021-01-21
Added logic to account for enrollment when returning onboarding status

alangsto:
**Description:**

@edx/masters-devs-cosmonauts 
The onboarding status endpoint does not currently take into account if a user is in an enrollment mode that is eligible for proctored exams. Now this endpoint should return a 404 when a user does not have the proper permissions to take proctored exams.

**JIRA:**

[MST-579](https://openedx.atlassian.net/browse/MST-579)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[2.6.2] - 2021-01-25
+~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


schenedx:
@alangsto The work here is awesome. There is a concern to me tho.
Can you test, maybe easier to do in production, to see if audit learner can access Proctortrack onboarding exams? I'd like to know because I suspect this is a bug we have where audit learner can create onboarding exam attempts right now?

PR #762, opened 2021-01-26
Add exam attempts removal endpoint

alangsto:
**Description:**

Created new endpoint to remove all attempts for a given exam id and user id. The endpoint makes use of the `remove_exam_attempt` function which updates the credit requirement status and grade overrides as needed.

This endpoint will be invoked on the instructor dashboard in place of the current reset attempt endpoint, as we now have multiple attempts. In the current code, only course staff or edX staff can use this endpoint.

**JIRA:**

[MST-582](https://openedx.atlassian.net/browse/MST-582)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -85,8 +85,12 @@ class ProctoredExamsApiTests(LoggedInTestCase):
                         try:
                             response = self.client.get(reverse(name, args=["0/0/0"]))
                         except NoReverseMatch:
-                            # some require 2 args.
-                            response = self.client.get(reverse(name, args=["0/0/0", 0]))
+                            try:
+                                # some require 2 args. Try first with course id regex match
+                                response = self.client.get(reverse(name, args=["0/0/0", 0]))
+                            except NoReverseMatch:
+                                # next try with a user id or exam id regex match
+                                response = self.client.get(reverse(name, args=["000", 0]))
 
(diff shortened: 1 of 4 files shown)


edx_proctoring/tests/test_views.py
```diff
@@ -85,8 +85,11 @@ def test_no_anonymous_access(self):
                         try:
                             response = self.client.get(reverse(name, args=["0/0/0"]))
                         except NoReverseMatch:
-                            # some require 2 args.
-                            response = self.client.get(reverse(name, args=["0/0/0", 0]))
+                            try:
```
schenedx, line 88: Can you explain why we need to make this change? Very weird code and difficult to read. Add some comments?

schenedx (commented):
Can you add to the PR description the context behind this change? Where do we plan to invoke this endpoint? Could learner delete their own exam attempts?

edx_proctoring/tests/test_views.py
alangsto: This was the one test that was failing. I think it failed because the arguments used for this test did not match the regex on the new endpoint I added, so I added another case to fit the regex for an exam id and a user id. I'll add a comment to explain that

alangsto:
@schenedx I updated the PR comment to add more context

schenedx (approved):
👍

PR #766, opened 2021-01-27
Bug fix for exam reset

alangsto:
**Description:**

The current code was preventing an exam reset in self paced courses, where the exam would cause a change in certificate/grades. I believe this is caused by the fact that this line here https://github.com/edx/edx-proctoring/blob/a7ca6841fb09804986acf6c6f3156c4fcac72ba3/edx_proctoring/signals.py#L128 triggers this line https://github.com/edx/edx-proctoring/blob/a7ca6841fb09804986acf6c6f3156c4fcac72ba3/edx_proctoring/signals.py#L178 which expects an active attempt. To resolve this, I've updated the code to archive the attempt before saving the update to the review model to ensure that any post save workflow for the review model mirrors if an attempt is active or not. 

**JIRA:**

[MST-616](https://openedx.atlassian.net/browse/MST-616)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,10 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+
+[2.6.4] - 2021-01-26
+~~~~~~~~~~~~~~~~~~~~~
+* Fix bug that was preventing exams from being reset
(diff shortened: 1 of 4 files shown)


mattcarter (approved):
LGTM

PR #768, opened 2021-01-28
Update onboarding exam error interstitial

alangsto:
**Description:**

This PR updates the api endpoint used to reset an onboarding attempt that is in an error state. This change aligns all of the onboarding interstitials that allow a reset to use the same endpoint. 

The retry button is not put behind a flag right now, as the option to retry an onboarding exam has already been available. 

**JIRA:**

[MST-616](https://openedx.atlassian.net/browse/MST-616)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/templates/practice_exam/error.html
+++ b/edx_proctoring/templates/practice_exam/error.html
@@ -25,39 +25,18 @@
     {% endblock %}
   </p>
   <hr>
-  <button class="gated-sequence start-timed-exam action-primary" data-ajax-url="{{enter_exam_endpoint}}" data-exam-id="{{exam_id}}" data-attempt-proctored=true data-start-immediately=false>
-    <span class="icon fa fa-lock" aria-hidden="true"></span>
-    <a>
-      {% block retry_cta %}
-        {% trans "Try this practice exam again" %}
-      {% endblock %}
-    </a>
-  </button>
+  <div>
+    {% trans "Retry my exam" as retry_exam %}
+    <button type="button" class="exam-action-button proctored-reset-onboarding btn btn-pl-primary btn-base" data-action="reset_attempt" data-exam-id="{{exam_id}}" data-change-state-url="{{change_state_url}}" data-loading-text="<span class='fa fa-circle-o-notch fa-spin'></span> {% trans 'Resetting Onboarding Exam' %}" data-cta-text="{{ retry_exam }}">
+      {{ retry_exam }}
+    </button>
+  </div>
+
 </div>
 {% include 'proctored_exam/footer.html' %}
 <script type="text/javascript">
-  $('.start-timed-exam').click(
-    function(event) {
-      var action_url = $(this).data('ajax-url');
-      var exam_id = $(this).data('exam-id');
-      var attempt_proctored = $(this).data('attempt-proctored');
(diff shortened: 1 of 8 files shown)


schenedx (commented):
Is there a way to add unit tests to this change?

edx_proctoring/templates/practice_exam/error.html
```diff
@@ -25,17 +25,20 @@ <h4>
     {% endblock %}
   </p>
   <hr>
-  <button class="gated-sequence start-timed-exam action-primary" data-ajax-url="{{enter_exam_endpoint}}" data-exam-id="{{exam_id}}" data-attempt-proctored=true data-start-immediately=false>
-    <span class="icon fa fa-lock" aria-hidden="true"></span>
-    <a>
-      {% block retry_cta %}
+  {% block retry_cta %}
+    <button class="gated-sequence start-timed-exam action-primary" data-ajax-url="{{enter_exam_endpoint}}" data-exam-id="{{exam_id}}" data-attempt-proctored=true data-start-immediately=false>
```
Zacharis278, line 29: Is there a reason we can't go through the same "reset_attempt" code path for these non-onboarding exams? I'm thinking the bug could still exist when called from this view unless we make backend changes to delete when this button is clicked.

alangsto: I think technically they can, and it may be better long term so that practice + onboarding exams have an even more similar flow

edx_proctoring/tests/test_student_view.py
```diff
@@ -1309,3 +1311,44 @@ def test_inactive_account_interstitial(self, render_exam):
...
+            exam_attempt = self._create_started_practice_exam_attempt()
+        exam_attempt.status = original_status
+        exam_attempt.save()
+
+        # check that only one attempt exists in history table
+        active_attempts = ProctoredExamStudentAttempt.objects.filter(status=original_status)
+        self.assertEqual(len(active_attempts), 1)
+
+        rendered_response = render_exam(self)
+        self.assertIn('exam-action-button', rendered_response)
```
alangsto, line 1339: I couldn't see a clean way to mock clicking on the button in this test set up, so instead I checked to see that the button is there, and then called the endpoint that would be hit if the button is clicked

edx_proctoring/tests/utils.py
```diff
@@ -306,7 +306,7 @@ def _create_onboarding_attempt(self):
         """
         Create onboarding attempt
         """
-        return self._create_exam_attempt(self.onboarding_exam_id)
+        return self._create_exam_attempt(self.onboarding_exam_id, is_practice_exam=True)
```
alangsto, line 309: I think this was an oversight, so I marked that onboarding exams should also be considered practice exams

edx_proctoring/tests/test_student_view.py
schenedx: That's fine. I think it would be better to put the explanation you have here into the comments of the code, so next engineer will benefit from your explanation here. Just a #NIT.

schenedx (approved):
👍

Zacharis278 (approved):
👍

PR #779, opened 2021-02-09
Fix Django admin bugs in ProctoredExamStudentAttempt admin form.

MichaelRoytman:
**Description:**

* Add time_remaining_seconds field of ProctoredExamStudentAttempt model to readonly_fields in 
  Django admin page so it is not required when editing the model.

### Old:

image
image

### New:

image

* Update reference to Exception.message to use string representation of the exception, as message
  is no longer an attribute of the Exception class.

### Old:

image

### New:

image

**JIRA:**

None.

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/admin.py
+++ b/edx_proctoring/admin.py
@@ -437,7 +437,8 @@ class ProctoredExamStudentAttemptAdmin(admin.ModelAdmin):
         'is_sample_attempt',
         'student_name',
         'review_policy_id',
-        'is_status_acknowledged'
+        'is_status_acknowledged',
+        'time_remaining_seconds'
     ]
 
     list_display = [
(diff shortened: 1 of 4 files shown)


schenedx (approved):
👍

alangsto (approved):

edx_proctoring/admin.py
```diff
@@ -485,7 +486,9 @@ def save_model(self, request, obj, form, change):
             if change:
                 update_attempt_status(obj.proctored_exam.id, obj.user.id, form.cleaned_data['status'])
         except (ProctoredExamIllegalStatusTransition, StudentExamAttemptDoesNotExistsException) as ex:
-            messages.error(request, ex.message)
+            # prevent showing success message inappropriately
+            messages.set_level(request, messages.ERROR)
```
idegtiarov, line 490: Hi, @MichaelRoytman could you please clarify what success messages could be shown inappropriately? 
Maybe you have an example. I'd like to leave as much info to the user as we can, or I missed something?

MichaelRoytman: Hi, @idegtiarov. Thanks for the feedback! I'll attach a photo below of what is happening if I don't set the level of the messages. The model isn't saved due to an illegal status transition (the red banner), but the admin page is incorrectly informing the user that the model was saved successfully (the green banner). What do you think?

image

idegtiarov: Thanks for the clarification. You are right two opposite messages looks odd

idegtiarov (approved):

PR #792, opened 2021-02-23
Onboard panel visibility

Zacharis278:
@edx/masters-devs-cosmonauts 

**Description:**

Determine visibility of the onboarding panel based on a user's access to the exam's sequence and if that content has been released.

~Note! The learning sequences code this is based on is potentially a few weeks out from production ready. I'm gathering some info on the implications of turning this on now. This would likely be an improvement over the current experience even if a few courses (not outline backfilled) would lose access to the panel.~
This backfill has been completed

Related edx-platform PR: https://github.com/edx/edx-platform/pull/26692

PR for learning MFE: https://github.com/edx/frontend-app-learning/pull/375

**JIRA:**

[MST-635](https://openedx.atlassian.net/browse/MST-635)

**Pre-Merge Checklist:**

**Post-Merge:**


Zacharis278:
Screen Shot 2021-02-24 at 11 05 24 AM

schenedx (commented):
What happens if the onboarding exam is past due? Do we show a date in the past? I don't feel like that's a good learner experience?

edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
```diff
@@ -112,16 +116,21 @@
...
+            var now = new Date();
             var data = this.model.toJSON();
             if (this.template) {
                 this.status = data.onboarding_status;
                 statusText = this.getExamAttemptText(data.onboarding_status);
+                releaseDate = new Date(data.onboarding_release_date);
                 data = {
                     onboardingStatus: statusText.status,
                     onboardingMessage: statusText.message,
+                    onboardingClosed: releaseDate > now,
```
alangsto, line 129: I feel like onboardingClosed could have two meanings - the onboarding is in the future or the onboarding window has already passed. Do we want to differentiate between these two?

Zacharis278:
> What happens if the onboarding exam is past due? Do we show a date in the past? I don't feel like that's a good learner experience?

The experience for past due is unchanged. You can click the onboarding button and you will just end up at the interstitial that the due date has passed.

edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
Zacharis278: Yeah this could probably use a better variable name

edx_proctoring/views.py
```diff
@@ -344,25 +348,36 @@ def get(self, request):
...
                 status=404,
                 data={'detail': _('There is no onboarding exam related to this course id.')}
             )
 
+        learning_sequences_service = get_runtime_service('learning_sequences')
+        course_key = CourseKey.from_string(course_id)
         user = get_user_model().objects.get(username=(username or request.user.username))
-        serialized_onboarding_exam = ProctoredExamSerializer(onboarding_exam).data
+        details = learning_sequences_service.get_user_course_outline_details(
+            course_key, user, pytz.utc.localize(datetime.max)
```
ormsbee, line 363: For user-research purposes (as the maintainers of the `learning_sequences` API), why `datetime.max`? Is there a concept the API is missing?

Zacharis278: Yeah, this seems to be the only way to know if a sequence will become accessible to the user later in the course. I'd use the course end date here if I had it before making this call.

ormsbee: Ah, got it. So it's not enough to know whether they can know it exists now, we have to know whether they will someday have access to it. Thank you!

Minor nit: now that we're running in Python 3 only, you don't have to use pytz–you can get the max datetime in UTC with just the stdlib:

```python
from datetime import datetime, timezone

datetime.max.replace(tzinfo=timezone.utc)
```

ormsbee: Actually, do we have to worry that it will close before `datetime.max`? Just trying to figure out if there's another API call we're looking for that will specifically get anything that will ever become accessible between a certain range of dates...?

Zacharis278: Actually yes, that's something I hadn't fully considered. The due date does not prevent a section from returning as an accessible sequence from what I've seen. Is there another mechanism which would cause sections not to be returned for `datetime.max`?

edx_proctoring/tests/test_views.py
```diff
@@ -736,6 +760,79 @@ def test_ineligible_for_onboarding_exam(self):
         message = 'There is no onboarding exam accessible to this user.'
         self.assertEqual(response_data['detail'], message)
 
+    def test_multiple_hidden_onboarding_exams(self):
```
schenedx, line 763: Can we have a test case where none of the exams within the course content is visible to the learner? The test case is to mimic the Masters learner in GT course case. The Masters learner should not see any onboarding exams.

edx_proctoring/views.py
```diff
@@ -349,25 +352,36 @@ def get(self, request):
...
+        )
+
+        for onboarding_exam in onboarding_exams:
+            usage_key = BlockUsageLocator.from_string(onboarding_exam.content_id)
+            if usage_key not in details.outline.accessible_sequences:
+                onboarding_exams.remove(onboarding_exam)
 
-        if not user.has_perm('edx_proctoring.can_take_proctored_exam', serialized_onboarding_exam):
+        if not onboarding_exams:
```
schenedx, line 375: Can we do a log.info in this case to inform splunk with the user_id who cannot see any onboarding exams in which course?

schenedx (approved):
I got two minor suggestions. Let me know if the make sense and got adopted.

edx_proctoring/tests/test_views.py
Zacharis278: Sounds like a good idea to test for no exams. This won't entirely handle the GT case however since the exam sections are still accessible to the learner.

PR #794, opened 2021-02-24
[MST-666] Update onboarding status view to consider attempts from other courses

bseverino:
**Description:**

If the learner has no onboarding attempts in a course, an additional check is added for verified attempts in other courses. If at least one verified attempt exists, a message will display on the onboarding status panel that their onboarding profile is approved in another course. If the attempt is close to expiring (within a month), the message will be updated to inform the user.

---
Screen Shot 2021-02-25 at 5 35 53 PM
---
Screen Shot 2021-02-26 at 11 35 57 AM
---

**JIRA:**

[MST-666](https://openedx.atlassian.net/browse/MST-666)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/models.py
+++ b/edx_proctoring/models.py
@@ -6,6 +6,9 @@ Data models for the proctoring subsystem
 # pylint: disable=model-missing-unicode
 
 
+from datetime import datetime, timedelta
+
+import pytz
 from model_utils.models import TimeStampedModel
 
 from django.contrib.auth import get_user_model
@@ -272,9 +275,10 @@ class ProctoredExamStudentAttemptManager(models.Manager):
         """
         return self.filter(proctored_exam_id=exam_id)
 
-    def get_onboarding_attempts_by_course_id(self, course_id, users=None):
+    # pylint: disable=invalid-name
+    def get_proctored_practice_attempts_by_course_id(self, course_id, users=None):
         """
-        Returns all onboarding attempts for a course, ordered by descending modified field.
+        Returns all proctored practice attempts for a course, ordered by descending modified field.
 
         Parameters:
         * course_id: ID of the course
@@ -288,6 +292,24 @@ class ProctoredExamStudentAttemptManager(models.Manager):
         if users:
             queryset = queryset.filter(user__in=users)
         return queryset
+
+    # pylint: disable=invalid-name
+    def get_last_verified_proctored_practice_attempt(self, user_id, proctoring_backend):
+        """
+        Returns the user's last verified proctored practice attempt for a specific backend,
+        if it exists. This only considers attempts within the last two years, as attempts
+        before this point are considered expired.
+
+        Parameters:
+            * user_id: ID of the user
+            * proctoring_backend: The name of the proctoring backend
+        """
+        earliest_allowed_date = datetime.now(pytz.UTC) - timedelta(days=730)
+        return self.filter(
+            user_id=user_id, taking_as_proctored=True, proctored_exam__is_practice_exam=True,
+            proctored_exam__backend=proctoring_backend, modified__gt=earliest_allowed_date,
+            status=ProctoredExamStudentAttemptStatus.verified
+        ).order_by('-modified').first()
 
     def get_filtered_exam_attempts(self, course_id, search_by):
         """
--- a/edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
+++ b/edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
@@ -38,6 +38,26 @@
         error: {
             status: gettext('Error'),
             message: gettext('An error has occurred during your onboarding exam. Please retry onboarding.')
+        },
+        other_course_approved: {
+            status: gettext('Approved in Another Course'),
+            message: gettext(
+                'Your onboarding profile has been approved in another course, ' +
+                'so you are eligible to take proctored exams in this course. ' +
+                'However, it is highly recommended that you complete this ' +
+                'course\'s onboarding exam in order to ensure that your device ' +
+                'still meets the requirements for proctoring.'
+            )
+        },
+        expiring_soon: {
+            status: gettext('Expiring Soon'),
+            message: gettext(
+                'Your onboarding profile has been approved in another course, ' +
+                'so you are eligible to take proctored exams in this course. ' +
+                'However, your onboarding status is expiring soon. Please ' +
+                'complete onboarding again to ensure that you will be ' +
+                'able to continue taking proctored exams.'
+            )
         }
     };
 
@@ -54,9 +74,9 @@
         updateCss: function() {
             var $el = $(this.el);
             var color = '#b20610';
-            if (this.status === 'verified') {
+            if (['verified', 'other_course_approved'].includes(this.status)) {
                 color = '#008100';
-            } else if (['submitted', 'second_review_required'].includes(this.status)) {
+            } else if (['submitted', 'second_review_required', 'expiring_soon'].includes(this.status)) {
                 color = '#0d4e6c';
             }
 
@@ -104,6 +124,13 @@
             }
         },
 
+        isExpiringSoon: function(expirationDate) {
+            var today = new Date();
+            var expirationDateObject = new Date(expirationDate);
+            // Return true if the expiration date is within 28 days
+            return today.getTime() > expirationDateObject.getTime() - 2419200000;
+        },
+
         shouldShowExamLink: function(status) {
             // show the exam link if the user should retry onboarding, or if they haven't submitted the exam
             var NO_SHOW_STATES = ['submitted', 'second_review_required', 'verified'];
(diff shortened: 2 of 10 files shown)


edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
```diff
@@ -38,6 +38,26 @@
         error: {
             status: gettext('Error'),
             message: gettext('An error has occurred during your onboarding exam. Please retry onboarding.')
+        },
+        other_course_approved: {
+            status: gettext('Approved in Another Course'),
+            message: gettext(
+                'You have completed an onboarding exam in another course, ' +
```
MichaelRoytman, line 45: I'm not sure whether you saw my feedback in the thread you posted, so I wanted to repost it here.

Nit: I wonder if "you have completed an onboarding exam in another course" may incorrectly suggest that a learner just needs to  complete an onboarding exam in order to be verified. Maybe "Your onboarding profile has been approved in another course" or something along those lines. Or "Your submitted onboarding attempt in another course has been approved"?

edx_proctoring/models.py
```diff
@@ -321,6 +322,53 @@ def get_active_student_attempts(self, user_id, course_id=None):
 
         return self.filter(filtered_query).order_by('-created')  # pylint: disable=no-member
 
+    def get_user_onboarding_attempt(self, user_id, course_id=None):
```
MichaelRoytman, line 325: Can you make this name more descriptive? There's a lot of logic here that is not really clear from the name.

MichaelRoytman, line 325: I'm a little wary of referring to this as an onboarding attempt. An attempt is an onboarding attempt by virtue of the backend used in the course supporting onboarding. It's not really a property on the attempt itself. The attempt is a proctored practice exam. There's nothing about the attempt itself to suggest it's onboarding. I think it creates some confusion to refer to practice proctored exam attempts as being onboarding attempts when we don't know whether the attempts are in an exam in a course that uses a vendor that supports onboarding.

edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
bseverino: Meant to update this piece based on your feedback, thank you for reminding me!

edx_proctoring/models.py
bseverino: Yeah, I was having trouble coming up with something that encompassed all the logic. Also considering your comment below: Maybe something like`get_last_relevant_proctored_practice_attempt`?

bseverino: That makes sense. I can update the language to make it more generic.

MichaelRoytman: I think that's better!

edx_proctoring/models.py
```diff
@@ -321,6 +322,53 @@ def get_active_student_attempts(self, user_id, course_id=None):
...
+            1. The last verified onboarding attempt, if it exists
+            2. The most recent attempt, if there is no verified attempt
+            3. None, if there are no onboarding attempts
+        
+        Parameters:
+            * user_id: ID of the user
+            * course_id: Optional, used if we want to prioritize attempts from a certain course
+                         (I.E. for the instructor dashboard)
+        """
+        earliest_allowed_date = timezone.now() - timedelta(years=2)
```
MichaelRoytman, line 338: I'm not really sure of the best solution to this potential problem (maybe it's not even a problem), but I wanted to get my thoughts out there.

This method will run code that will hit the database multiple times in each call to it, say _m_. When you call bool on a queryset, the queryset is evaluated. If we're calling this function in a loop _n_  times for _n_ users we'd be making _m*n_ database calls. Could you take a look at the Django debug toolbar and see if you can see the database queries? Perhaps Django is smart enough to optimize to some degree. If not, it may make sense to think about how we can do this in fewer calls to the database if it seems like it might be problematic. But hopefully not!

edx_proctoring/statuses.py
```diff
@@ -252,6 +252,12 @@ class InstructorDashboardOnboardingAttemptStatus:
     verified = 'verified'
     error = 'error'
 
+    # The following statuses are not true attempt statuses, but are used when the
+    # user's onboarding profile is approved in a different course.
+    other_course_approved = 'other_course_approved'
+    # Onboarding attempts expire after two years.
+    expiring_soon = 'expiring_soon'
```
Zacharis278, line 259: I am wondering if this should really be a status and instead some attribute we return on the API call for the learner panel specifically. Maybe a `notify_of_pending_expiration` bool or something? I'm worried about overloading the use of status with things that aren't actually a real value for status.  For example I don't think we wanna show this on the instructor side since it's not reflected in Verificient's dashboard that way.

edx_proctoring/models.py
bseverino: Thank you for pointing this out. I think taking advantage of [exists()](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#exists) will allow better performance (but by how much, I'm not sure yet)

edx_proctoring/statuses.py
bseverino: I think that makes sense. I meant to ask if we wanted to include the expiring status on the instructor dashboard as well; I'll leave it out in this case.

Zacharis278: another solution is maybe we just want to have the UI do the date logic here and just return `onboarding_expiration_date`

edx_proctoring/tests/utils.py
```diff
@@ -373,32 +373,3 @@ def _normalize_whitespace(string):
         Replaces newlines and multiple spaces with a single space.
         """
         return ' '.join(string.replace('\n', '').split())
-
-
-def create_onboarding_exam(
```
bseverino, line 378: Removed this util function as it duplicated another and was causing discrepancies in tests.

edx_proctoring/models.py
```diff
@@ -321,6 +324,58 @@ def get_active_student_attempts(self, user_id, course_id=None):
...
+            * course_id: Optional, used if we want to prioritize attempts from a certain course
+                         (I.E. for the instructor dashboard)
+        """
+        earliest_allowed_date = datetime.now(pytz.UTC) - timedelta(days=730)
+        proctored_practice_attempts = self.filter(
+            user_id=user_id, taking_as_proctored=True, proctored_exam__is_practice_exam=True,
+            proctored_exam__backend=proctoring_backend, modified__gt=earliest_allowed_date
+        ).order_by('-modified')
+
+        if proctored_practice_attempts.exists():
```
schenedx, line 347: I don't believe this if statement is needed. This is because, we use the records in `proctored_practice_attempts` below to determine behavior anyways. Filtering on top of this result, if this result is already empty, will return empty anyways.

edx_proctoring/api.py
```diff
@@ -1488,6 +1488,25 @@ def get_filtered_exam_attempts(course_id, search_by):
     return [ProctoredExamStudentAttemptSerializer(active_exam).data for active_exam in exam_attempts]
 
 
+def get_last_relevant_proctored_practice_attempt(user_id, proctoring_backend, course_id=None):
```
schenedx, line 1491: The current place this function is used, the course_id is always passed in. Why should we consider course_id is specified case at all in your change? Shouldn't we consider course_id <> the current course_id of the request?

edx_proctoring/models.py
```diff
@@ -321,6 +324,58 @@ def get_active_student_attempts(self, user_id, course_id=None):
 
         return self.filter(filtered_query).order_by('-created')  # pylint: disable=no-member
 
+    def get_last_relevant_proctored_practice_attempt(self, user_id, proctoring_backend, course_id=None):
+        """
+        Filter user's proctored practice attempts within the last two years, and return one of the
+        following:
+            1. The last proctored practice attempt, if it exists
+            2. The most recent attempt, if there is no verified attempt
```
schenedx, line 332: Why bother return the most recent not verified attempt? Why have this function be generic to return last relevant attempt, instead of last verified attempt? This makes the function specificity mixed.

schenedx (commented):
There are some code design choices I think can use a bit more discussion. If you are available, we can chat today.

edx_proctoring/models.py
```diff
@@ -289,6 +293,24 @@ def get_onboarding_attempts_by_course_id(self, course_id, users=None):
...
+        """
+        Returns the user's last verified proctored practice attempt for a specific backend,
+        if it exists. This only considers attempts within the last two years, as attempts
+        before this point are considered expired.
+
+        Parameters:
+            * user_id: ID of the user
+            * proctoring_backend: The name of the proctoring backend
+        """
+        earliest_allowed_date = datetime.now(pytz.UTC) - timedelta(days=730)
```
schenedx, line 307: As a follow up PR, I'd prefer to have this number `730` to be in a edx-platform setting specific to Proctoring.

schenedx (approved):
👍 You can address my comment in a later PR

edx_proctoring/statuses.py
```diff
@@ -261,7 +265,7 @@ class InstructorDashboardOnboardingAttemptStatus:
         ProctoredExamStudentAttemptStatus.submitted: submitted,
         ProctoredExamStudentAttemptStatus.rejected: rejected,
         ProctoredExamStudentAttemptStatus.verified: verified,
-        ProctoredExamStudentAttemptStatus.error: error,
```
idegtiarov, line 264: An odd change, it seems that it should be reverted.

idegtiarov (changes requested):
Some minor comments are inline, please take a look at them.

edx_proctoring/static/proctoring/js/views/proctored_exam_info.js
```diff
@@ -104,6 +124,13 @@
             }
         },
 
+        isExpiringSoon: function(expirationDate) {
+            var today = new Date();
+            var expirationDateObject = new Date(expirationDate);
+            // Return true if the expiration date is within a month
+            return today > expirationDateObject.getMonth() - 1;
```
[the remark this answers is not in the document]
bseverino: Thank you for pointing this out - I've changed the comparison to use 28 days in milliseconds, to make it more consistent across the board. This may be changed to a setting in the future.

idegtiarov (approved):
Thanks for the quick changes! Now LGTM

PR #801, opened 2021-03-01
MST-675: QuerySet Performance for  StudentOnboardingStatusByCourse Endpoint

MichaelRoytman:
**Description:**

See the corresponding edx-platform PR [here](https://github.com/edx/edx-platform/pull/26789).

The original use of `user.has_perm('can_take_proctored_exam')` had very poor performance. The `can_take_proctored_exam` permission, defined in the LMS, was not designed for use in bulk operations. It was being called for each user in a loop, resulting in many queries to the database. This change uses a new service endpoint that performs this permission checking in the database, resulting in one single query to the necessary LMS SQL tables and many fewer queries to the modulestore.

**JIRA:**

[MST-675](https://openedx.atlassian.net/browse/MST-675)

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/api.py
```diff
@@ -2490,12 +2490,13 @@ def get_integration_specific_email(provider):
     return getattr(provider, 'integration_specific_email', None) or constants.DEFAULT_CONTACT_EMAIL
 
 
-def get_enrollments_for_course(course_id):
+def get_enrollments_for_course_can_take_proctored_exams(course_id, text_search=None):
```
schenedx, line 2493: Why do we need the API function? Why not just use the `enrollment_service` directly in the view? I see patterns either way in edx_proctoring, so there is not standard on this, right?

schenedx (approved):
I think this looks much better.

edx_proctoring/api.py
MichaelRoytman: I assume it's to provide an abstraction over how data is being fetched, allowing us to keep that logic out of the view.

edx_proctoring/tests/test_views.py
```diff
@@ -855,182 +855,6 @@ def test_pagination_maintains_query_params(self):
         self.assertIn(text_search_string, next_url)
         self.assertIn(statuses_filter_string, next_url)
 
-    def test_partial_text_search_username(self):
```
schenedx, line 858: Why were all these test cases removed?

MichaelRoytman: edx_proctoring is no longer doing the username and email searching - it has been moved to the LMS as part of the `get_enrollments_can_take_proctored_exams ` method. I wrote tests in the LMS to capture this functionality.

schenedx: Got it. But I feel like these tests also cover the restful API interface on the endpoint here. Therefore, keeping them here won't hurt. We should have double coverage here.

MichaelRoytman: We're mocking out the enrollments service, so I would need to eitherreplicate the user searching there, which I don't think we want to do, or I'd need to import the function from the LMS here. Are you suggesting the second?

schenedx: I see. So, if we keep mocking here, the tests are no longer relevant. 
That said, I believe you can update one of these tests to check the mock is called with text search passed into the mock. That way, you leave maybe 1 test to ensure the text search interface is implemented. I am not suggesting you need to import from LMS

schenedx (approved):
👍

PR #803, opened 2021-03-02
MST-669 Show "Approved in other course" status on Student onboarding status panel

schenedx:
**Description:**

Make sure if the user do not have a verified onboarding attempt on the current course, we check the verified onboarding attempt on other courses within 2 years

**JIRA:**

[MST-669](https://openedx.atlassian.net/browse/MST-669)

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/views.py
```diff
@@ -469,18 +469,29 @@ def get(self, request, course_id):
...
 
-            data = {}
-            data['username'] = user.username
-            data['status'] = (InstructorDashboardOnboardingAttemptStatus
-                              .get_onboarding_status_from_attempt_status(user_attempt.get('status')))
-            data['modified'] = user_attempt.get('modified')
+            data = {'username': user.username}
+
+            if not user_attempt and other_verified_attempt:
+                data['status'] = InstructorDashboardOnboardingAttemptStatus.other_course_approved
```
bseverino, line 489: We'll also have to add this status to [proctored_exam_onboarding_view.js](https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/static/proctoring/js/views/proctored_exam_onboarding_view.js) - I believe it will just render as `other_course_approved` without changing it, but it won't be included in the filters.

edx_proctoring/views.py
```diff
@@ -469,18 +469,29 @@ def get(self, request, course_id):
...
+            other_verified_attempt = last_verified_attempt_dict.get(user.id)
 
-            data = {}
-            data['username'] = user.username
-            data['status'] = (InstructorDashboardOnboardingAttemptStatus
-                              .get_onboarding_status_from_attempt_status(user_attempt.get('status')))
-            data['modified'] = user_attempt.get('modified')
+            data = {'username': user.username}
+
+            if not user_attempt and other_verified_attempt:
```
Zacharis278, line 488: I think this differs from the logic of the user view where we have `verified in another course` taking precedence over non-verified states in the current course.

My gut is that we want to replicate that here as well.

edx_proctoring/views.py
schenedx: Got it. I will update that file.

edx_proctoring/views.py
schenedx: I understand that logic differently than what you said here. See https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/views.py#L374
That logic says to me, only if the learner have no attempt in the current course, we would check the verified attempt in other courses. I think that makes more sense because we want to show learner's status within this course.
Do you understand this differently?

Zacharis278: ahh you're right, this is the same in both places

Zacharis278 (approved):
LGTM pending the JS changes.

idegtiarov (commented):
LGTM so far!

bseverino (approved):
LGTM 👍

PR #804, opened 2021-03-03
[MST-626] Add more detail and consistency to log messages

bseverino:
**Description:**

Added more useful attributes to log messages, in a `key=value` format that is easier to extract, and reduced duplicate exception logs.

**JIRA:**

[MST-626](https://openedx.atlassian.net/browse/MST-626)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/backends/rest.py
+++ b/edx_proctoring/backends/rest.py
@@ -184,7 +184,11 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
         payload['status'] = 'created'
         # attempt code isn't needed in this API
         payload.pop('attempt_code', False)
-        log.debug(u'Creating exam attempt for %r at %r', exam['external_id'], url)
+        log.debug(
+            'Creating exam attempt for exam_id={exam_id} (external_id={external_id}) at {url}'.format(
+                exam_id=exam['id'], external_id=exam['external_id'], url=url
+            )
+        )
         response = self.session.post(url, json=payload)
         if response.status_code != 200:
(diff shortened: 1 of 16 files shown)


edx_proctoring/backends/rest.py
```diff
@@ -184,7 +184,11 @@ def register_exam_attempt(self, exam, context):
         payload['status'] = 'created'
         # attempt code isn't needed in this API
         payload.pop('attempt_code', False)
-        log.debug(u'Creating exam attempt for %r at %r', exam['external_id'], url)
+        log.debug(
+            'Creating exam attempt for exam_id = {exam_id} (external_id = {external_id}) at {url}'.format(
```
Zacharis278, line 188: Something worth confirming is if spunk can extract the field with spaces on either side of equals. You may have to reformat this as `key=val`

If you already know the spaces work, disregard this.

bseverino: Good call, I think you're right

edx_proctoring/backends/software_secure.py
```diff
@@ -80,7 +80,7 @@ def register_exam_attempt(self, exam, context):
 
         if status not in [200, 201]:
             err_msg = (
-                u'Could not register attempt_code = {attempt_code}. '
+                u'Could not register attempt_code={attempt_code}. '
```
idegtiarov, line 83: One nit question: why `u` prefix was left? It seems that it could be removed here as well.

idegtiarov (commented):
Other LGTM! Thanks.

PR #809, opened 2021-03-09
onboard verified state precedence

Zacharis278:
@edx/masters-devs-cosmonauts 

**Description:**

Allows the 'verified in another course' state to take precedence over a non-verified onboarding state in the current course.

Example: if a user is verified in course A and submitted in course B, requesting the onboarding state of that user for course B will return "verified in another course". 

**JIRA:**

[MST-688](https://openedx.atlassian.net/browse/MST-688)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -1267,7 +1267,6 @@ class TestStudentOnboardingStatusByCourseView(ProctoredExamTestCase):
             'next': None,
             'num_pages': 1,
         }
-        self.maxDiff = None
         self.assertEqual(response_data, expected_data)
 
     def test_not_staff_or_course_staff(self):
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -604,7 +604,7 @@ class TestStudentOnboardingStatusView(ProctoredExamTestCase):
 
     def test_verified_in_another_course(self):
         """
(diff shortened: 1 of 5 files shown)


edx_proctoring/tests/test_views.py
```diff
@@ -1203,6 +1267,7 @@ def test_other_course_verified(self):
             'next': None,
             'num_pages': 1,
         }
+        self.maxDiff = None
```
schenedx, line 1270: Please remove.

edx_proctoring/views.py
```diff
@@ -371,23 +371,26 @@ def get(self, request):
 
         attempts = ProctoredExamStudentAttempt.objects.get_proctored_practice_attempts_by_course_id(course_id, [user])
 
-        if len(attempts) == 0:
-            # If there are no attempts in the current course, check for a verified attempt in another course
+        # Default to the most recent attempt in the course if there are no verified attempts
+        relevant_attempt = attempts[0] if attempts else None
```
schenedx, line 375: #NIT I feel like this set of logic to select the correct attempt to use can be refactored, so the logic at line 487 can share the same logic. Basically, a function that takes in the attempt within the course, and a verified_attempt outside of the current course, and return the attempt to return.

Zacharis278: Making a note to take a look at this while re-factoring in #792

PR #817, opened 2021-03-24
Improve onboarding copy

Zacharis278:
@edx/masters-devs-cosmonauts 

**Description:**

Improves a few messages to learners to reduce confusion.

1. De-emphasize the need to retake onboarding if the user already has a profile. I have attempted to reduce the amount of text and the size of the panel in this case.
2. Remove reference to identity verification as it may imply a relation IDV
3. Updates this submitted interstitial to explicitly state practice attempts will not be reviewed. The first sentence implies this but it's easy to read past. Do we think this change is too verbose?

**JIRA:**

[MST-687](https://openedx.atlassian.net/browse/MST-687?atlOrigin=eyJpIjoiMzE5NDU4ZDQzYmRmNDE5Y2I3OGMxZWRkYTAwYjUyOGMiLCJwIjoiaiJ9)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/templates/onboarding_exam/submitted.html
+++ b/edx_proctoring/templates/onboarding_exam/submitted.html
@@ -47,6 +47,13 @@
   {% endif %}
   <p>
     {% blocktrans %}
+      If you already have an onboarding profile approved through another course,
+      the this submission will not be reviewed. You may retry this exam at any time
+      to validate the proctoring software is set up properly on your device.
+    {% endblocktrans %}
+  </p>
+  <p>
+    {% blocktrans %}
       Please contact
       <a href="mailto:{{ integration_specific_email }}">
         {{ integration_specific_email }}
--- a/edx_proctoring/templates/onboarding_exam/submitted.html
+++ b/edx_proctoring/templates/onboarding_exam/submitted.html
@@ -11,9 +11,8 @@
     {% blocktrans %}
       If you do not have an onboarding profile with the system,
       Verificient will review your submission and create an onboarding
-      profile, to grant you access to proctored exams.  Onboarding
-      profile review, including identity verification, can take 2+
-      business days.
+      profile to grant you access to proctored exams. Onboarding
+      profile review can take 2+ business days.
     {% endblocktrans %}
   </p>
   <p>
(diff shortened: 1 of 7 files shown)


Zacharis278:
screenshots 👀 
Screen Shot 2021-03-24 at 2 45 39 PM

Screen Shot 2021-03-24 at 2 46 02 PM

edx_proctoring/templates/onboarding_exam/submitted.html
```diff
@@ -45,6 +45,13 @@
       </button>
     </p>
   {% endif %}
+  <p>
+    {% blocktrans %}
+      If you already have an onboarding profile approved through another course,
```
MichaelRoytman, line 50: Nit.

We could echo the language used in the onboarding panel to establish consistency.

"You may retry this exam at any time to validate that your setup still meets the requirements for proctoring."

edx_proctoring/templates/onboarding_exam/submitted.html
```diff
@@ -45,6 +45,13 @@
       </button>
     </p>
   {% endif %}
+  <p>
```
MichaelRoytman, line 48: I can't comment on the line, but line 15 also references identity verification with respect to onboarding. I think we should also remove it there.

If you choose to do that, there's also an errant comma.

```
 If you do not have an onboarding profile with the system, then Verificient will review your submission and create an onboarding profile to grant you access to proctored exams.  Onboarding profile review can take 2+ business days.
```

edx_proctoring/templates/onboarding_exam/submitted.html
MichaelRoytman, line 50: There's something about the way this looks that is a little heavy on the eyes. It's hard for me to see what the call to action is, because the button is now sandwiched between text. What do you think about moving this new copy above "If you have made an error..." but below "Once your profile has been reviewed"?

schenedx (approved):
LGTM
I think the new message makes sense.

MichaelRoytman (approved):
This looks good to me. I appreciate the increased clarify that this adds.

I left a few comments, but they're not blockers.

edx_proctoring/templates/onboarding_exam/submitted.html
Zacharis278: yeah that would make sense, I think this panel is a bit verbose overall but I'd rather over-inform I suppose

alangsto: Alternatively, could we only present certain information if the learner is in another course? This is something that I had outlined in my [discovery work](https://openedx.atlassian.net/wiki/spaces/~579530305/pages/2523103313/MST-673+Discovery+Work) for MST-673. I think it may be easier to read this panel if we're only presenting the relevant info that a learner will need

Zacharis278:
I've updated the screens based on feedback

PR #826, opened 2021-04-02
MST-734 Add due_date to the learning_sequence accessible_sequences check

schenedx:
**Description:**

We want to ensure due dates do not hide learner's onboarding status panel

**JIRA:**

[MST-734](https://openedx.atlassian.net/browse/MST-734)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -80,6 +80,7 @@ from edx_proctoring.statuses import (
 from edx_proctoring.utils import (
     AuthenticatedAPIView,
     get_time_remaining_for_attempt,
+    get_visibility_check_date,
     humanized_time,
     locate_attempt_by_attempt_code,
     obscured_user_id
@@ -369,7 +370,12 @@ class StudentOnboardingStatusView(ProctoredAPIView):
 
         for onboarding_exam in onboarding_exams:
             usage_key = BlockUsageLocator.from_string(onboarding_exam.content_id)
-            if usage_key not in details.outline.accessible_sequences:
+            visibility_check_date = get_visibility_check_date(details.schedule, usage_key)
+            user_outline = learning_sequences_service.get_user_course_outline(
+                course_key, user, visibility_check_date
+            )
+
+            if usage_key not in user_outline.accessible_sequences:
                 onboarding_exams.remove(onboarding_exam)
 
         if not onboarding_exams:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+[3.8.3] - 2021-04-05
+~~~~~~~~~~~~~~~~~~~~~
+* Use exam due_date or course end date to evaluate the visibility of the onboarding status panel
 
 [3.8.2] - 2021-04-02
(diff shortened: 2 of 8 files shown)


edx_proctoring/views.py
```diff
@@ -366,10 +366,18 @@ def get(self, request):
...
-
+        due_date = details.schedule.course_end or pytz.utc.localize(datetime.max)
         for onboarding_exam in onboarding_exams:
             usage_key = BlockUsageLocator.from_string(onboarding_exam.content_id)
-            if usage_key not in details.outline.accessible_sequences:
+            exam_schedule_item = details.schedule.sequences.get(usage_key)
+            if exam_schedule_item and exam_schedule_item.due:
+                due_date = exam_schedule_item.due
+
+            user_outline_detail = learning_sequences_service.get_user_course_outline_details(
```
schenedx, line 376: @ormsbee and @Zacharis278 I am calling the `get_user_course_outline_details` a second time here. This is because I needed the scheduledItem from the first call. Please advice if you have a better way to get course_end_date or exam due dates

ormsbee: I think it's fine to do this. Minor note that this second call, you probably don't need the `get_user_course_outline_details`, since you're only looking at whether the sequence is accessible or not and `get_user_course_outline` is sufficient to give you that.

ormsbee: In theory, you could get those specific dates more quickly from edx-when's API directly, but in practice, edx-when always pre-loads scheduling for the whole course anyway, so you don't actually save any time.

MichaelRoytman:
Is there a reason we are not hiding the onboarding panel after the course due date? I thought MST-734 was about the onboarding panel not appearing for learners because the onboarding exam was not visible to the learner. Can you explain how this pull request fixes that?

The reason I ask is that I think it will be confusing for learners if they see an onboarding panel after the exam due date and then are unable to take onboarding exams after the due date, which seems to be what Jennifer is recommending as a future change.

schenedx:
@MichaelRoytman You are correct in your feedback and helped me with a case I did not think of. I made the change here to address the onboarding panel after due date. I selected the current time if it is larger than the due date. This way, the exam with a due date will no longer be accessible.

edx_proctoring/views.py
schenedx: @ormsbee Took your advice on using `get_user_course_outline` for the second call.

Zacharis278:
The initial change grabbing the exam due date from sequences looks good to me. I'm going to suggest we think more about removing the panel after the onboarding exam due date however. In most cases, after the onboarding date the panel would be affirming the learner's ability to take the real exam ('verified') it may cause concern when that status is suddenly removed. ( a learner inferring something changed or went wrong with their status )

I do agree there is a potentially unaccounted for gap here in regards to the `hide_after_due` flag. Without that feature selected, the call to action after the due date will just sent the learner to the past due interstitial. I haven't tested it yet, but with the due date passed and `hide_after_due=True` we probably do have a bug.

schenedx:
@MichaelRoytman and @Zacharis278 Updated this PR according to your suggestions. https://openedx.atlassian.net/browse/MST-746 created for the onboarding panel display after exam due dates. Please review again.

Zacharis278 (approved):

PR #827, opened 2021-04-05
fix: Display Useful Status in InstructorDashboard StudentOnboardingPa…

MichaelRoytman:
fix: Display Useful Status in InstructorDashboard StudentOnboardingPanel for "onboarding reset" Attempt Status

**Description:**
Due to inconsistencies in the way we handle attempts in past due practice proctored/onboarding exams, learners can end up in an unintended liminal state after attempting to reset their onboarding attempt. If a learner attempts to reset their rejected onboarding attempt after the exam's due date, we process the reset request and move their attempt into the "onboarding_reset" state. Theoretically, a new exam attempt should be created immediately thereafter. However, we have code that prevents the creation of an exam attempt after the exam's due date, so the call to create a subsequent exam attempt fails, leaving the learner with an exam attempt with the "onboarding_reset" status. Theoretically, this situation should never occur, and the fact that it does is a bug. Because of this, we did not handle the "onboarding_reset" status in the StudentOnboardingStatus panel, and this status appears as "null". As an intermediate step, while we think about our due date logic, this pull request adds a new onboarding status "onboarding_status_past_due". This status is displayed as "Onboarding Reset Failed Due to Past Due Exam" in the StudentOnboardingPanel in the InstructorDashboard, which should provide course staff with a clearer explanation.

JIRA: [MST-745](https://openedx.atlassian.net/browse/MST-736) tracks the removal of this intermediate code from the code base once we fix the underlying cause of this bug.
JIRA: [MST-749](https://openedx.atlassian.net/browse/MST-749) tracks the fix for the behavior that allowed for this state to occur.

**JIRA:**

[MST-736](https://openedx.atlassian.net/browse/MST-736)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -523,8 +523,24 @@ class StudentOnboardingStatusByCourseView(ProctoredAPIView):
                 data['status'] = InstructorDashboardOnboardingAttemptStatus.other_course_approved
                 data['modified'] = other_verified_attempt.modified
             else:
-                data['status'] = (InstructorDashboardOnboardingAttemptStatus
-                                  .get_onboarding_status_from_attempt_status(user_attempt.get('status')))
+                attempt_status = user_attempt.get('status')
+
+                # If the learner's most recent attempt is in the "onboarding_reset" state,
+                # return the onboarding_reset_past_due state.
+                # This is a consequence of a bug in our software that allows a learner to end up
+                # with their only or their most recent exam attempt being in the "onboarding_reset" state.
+                # The learner should not end up in this state, but while we work on a fix, we should not
+                # display "null" in the Instructor Dashboard Student Onboarding Panel.
+                # TODO: remove as part of MST-745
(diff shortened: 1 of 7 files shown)


edx_proctoring/views.py
```diff
@@ -523,8 +523,23 @@ def get(self, request, course_id):
...
+                    )
+
+                # If we have no corresponding status for this attempt, and the learner's most recent attempt
+                # is in the "onboarding_reset" state, return the onboarding_reset_past_due state.
+                # This is a consequence of a bug in our software that allows a learner to end up
+                # with their only or their most recent exam attempt being in the "onboarding_reset" state.
+                # The learner should not end up in this state, but while we work on a fix, we should not
+                # display "null" in the Instructor Dashboard Student Onboarding Panel.
+                # TODO: remove as part of MST-745
+                if (onboarding_status is None and
```
Zacharis278, line 538: isn't the first half of this conditional redundant?

Zacharis278 (approved):

PR #832, opened 2021-04-16
fix: Add Explicit Dependency on pyjwt Library with Constraint

MichaelRoytman:
**Description:**

This pull request adds pyjwt as an explicit dependency of the library. Previously, it was an implicit dependency installed as a dependency of explicit dependencies of the library. The pyjwt library was being referenced directly in the codebase, so it should be an explicitly defined dependency of the library.

The pull request constrains the version of pyjwt to less than 2.0.0. The edx-platform, which edx-proctoring is installed into, constrains the version of the pyjwt to less than 2.0.0, which causes a version conflict when a version greater to or equal to 2.0.0 is installed. This pull request constrains the version so that subsequent Python upgrade pull requests, created via edX automation, do not reupgrade the pyjwt library.

**Pre-Merge Checklist:**

**Post-Merge:**


schenedx (approved):
👍

PR #841, opened 2021-05-06
[feat]: Make exam attempt resumable when the is_resumable field is set

schenedx:
**Description:**
Allow user to resume exam attempts which had error status before and got the is_resumable field set. This includes:

Implement the logic to set or reset the is_resumable field of the attempt model. The is_resumable field can only be set when the attempt is moved to "error" status. It will be reset when the attempt is transitioned to "ready_to_resumed", "resumed" or "submitted". The submitted status is not really possible for learners have no way to submit an errored attempt.

Update the UI to allow attempt to be resumable if the attempt has the is_resumable set. The UI no longer looks at only those attempts with "error" status

**JIRA:**

[MST-817](https://openedx.atlassian.net/browse/MST-817)

**Pre-Merge Checklist:**

- [ ] Updated the version number in `edx_proctoring/__init__.py` and `package.json` if these changes are to be released.
- [ ] Described your changes in `CHANGELOG.rst`
- [ ] Confirmed Github reports all automated tests/checks are passing.
- [ ] Approved by at least one additional reviewer.

**Post-Merge:**

- [ ] Create a tag matching the new version number.

PR #842, opened 2021-05-12
Revert "[feat]: Make exam attempt resumable when the is_resumable field is set"

MichaelRoytman:
Reverts edx/edx-proctoring#841


schenedx:
No longer needed.

PR #845, opened 2021-05-17
[Fix]: Add the proper lockfileVersion property to the npm-shrinkwrap.json

schenedx:
**Description:**

This is to fix the NPM install on github action while trying to publish a new version

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,7 +1,10478 @@
 {
+    "name": "@edx/edx-proctoring",
+    "version": "3.9.1",
+    "lockfileVersion": 1,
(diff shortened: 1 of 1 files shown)


alangsto (approved):

Zacharis278 (approved):
out of curiosity why did we need to add this?

schenedx:
@Zacharis278 npm-shrinkwrap.json is always there to fix the problem associated with `gulp test`. See https://stackoverflow.com/questions/55921442/how-to-fix-referenceerror-primordials-is-not-defined-in-node-js
I just updated this file in my last PR to try to make it work on my machine. That caused a failure with `npm ci` command.

PR #849, opened 2021-05-19
fix: process onboarding exams correctly in onboarding status API used by onboarding panel

MichaelRoytman:
**Description:**

When returning information about a learner's onboarding status in a course, we exclude onboarding exams that the Learning Sequences API marks as inaccessible. While processing the list of onboarding exams in the course, the list was being modified during iteration, leading to incorrect behavior. This code change fixes this behavior. This fix also exposed a bug in the corresponding tests, which was also fixed in this code change.

**JIRA:**

None.

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -438,6 +438,7 @@ class StudentOnboardingStatusView(ProctoredAPIView):
             course_key, user, pytz.utc.localize(datetime.max)
         )
 
+        inaccessible_onboarding_exams = []
         for onboarding_exam in onboarding_exams:
             usage_key = BlockUsageLocator.from_string(onboarding_exam.content_id)
             visibility_check_date = get_visibility_check_date(details.schedule, usage_key)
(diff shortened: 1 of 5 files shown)


edx_proctoring/views.py
```diff
@@ -446,7 +447,10 @@ def get(self, request):
             )
 
             if usage_key not in user_outline.accessible_sequences:
-                onboarding_exams.remove(onboarding_exam)
+                inaccessible_onboarding_exams.append(onboarding_exam)
```
schenedx, line 450: Can you explain why this change fixes the bug?

MichaelRoytman: The original code is modifying a list that it is concurrently iterating over, resulting in an incorrect calculation for `onboarding_exams`.

For example, if you have a list `onboarding_exams` with a length of two, and you hit line 449 and remove an exam, the iteration stops short of the second onboarding exam.

schenedx: Got it. Good catch.

schenedx (approved):
👍

PR #851, opened 2021-05-20
Use Verificient onboarding status API endpoint for course home proctoring panel

alangsto:
**Description:**

Verificient's onboarding status API should be used for the student onboarding panel endpoint `/edx_proctoring/v1/user_onboarding/status`. The use of the endpoint is behind a waffle switch. If that switch is activated, both the current logic and the new onboarding status API will be used so that we can log if there are differences between the two. If the onboarding status API responds with a non 200 HTTP status, we will default to use the data gathered by the current (non onboarding status api) logic. 

As part of this status, functions dealing with the onboarding status api endpoint that previously were named with `attempt` have been renamed as to prevent confusion around attempt vs onboarding profile information.

**JIRA:**

[MST-760](https://openedx.atlassian.net/browse/MST-760)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -55,7 +55,7 @@ from edx_proctoring.serializers import (
     ProctoredExamStudentAllowanceSerializer,
     ProctoredExamStudentAttemptSerializer
 )
-from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus
+from edx_proctoring.statuses import InstructorDashboardOnboardingAttemptStatus, ProctoredExamStudentAttemptStatus
 from edx_proctoring.utils import (
     emit_event,
     get_exam_due_date,
@@ -2723,3 +2723,42 @@ def get_enrollments_can_take_proctored_exams(course_id, text_search=None):
     """
     enrollments_service = get_runtime_service('enrollments')
     return enrollments_service.get_enrollments_can_take_proctored_exams(course_id, text_search)
+
+
+def get_onboarding_attempt_data_for_learner(course_id, user, backend):
+    """
+    Return the onboarding status and expiration date for a learner in a specific course. If no
+    onboarding status is associated with the course, check to see if the
+    learner is approved for onboarding in another course
+
+    Parameters:
+        * course_id: course ID for the course
+        * user: User object
+        * backend: proctoring provider to filter exams on
+    """
+    data = {'onboarding_status': None, 'expiration_date': None}
+    attempts = ProctoredExamStudentAttempt.objects.get_proctored_practice_attempts_by_course_id(course_id, [user])
+
+    # Default to the most recent attempt in the course if there are no verified attempts
+    relevant_attempt = attempts[0] if attempts else None
+    for attempt in attempts:
+        if attempt.status == ProctoredExamStudentAttemptStatus.verified:
+            relevant_attempt = attempt
+            break
+
+    # If there is no verified attempt in the current course, check for a verified attempt in another course
+    verified_attempt = None
+    if not relevant_attempt or relevant_attempt.status != ProctoredExamStudentAttemptStatus.verified:
+        attempt_dict = get_last_verified_onboarding_attempts_per_user(
+            [user],
+            backend,
+        )
+        verified_attempt = attempt_dict.get(user.id)
+
+    if verified_attempt:
+        data['onboarding_status'] = InstructorDashboardOnboardingAttemptStatus.other_course_approved
+        data['expiration_date'] = verified_attempt.modified + timedelta(days=constants.VERIFICATION_DAYS_VALID)
+    elif relevant_attempt:
+        data['onboarding_status'] = relevant_attempt.status
+
+    return data
--- a/edx_proctoring/statuses.py
+++ b/edx_proctoring/statuses.py
@@ -288,3 +288,35 @@ class InstructorDashboardOnboardingAttemptStatus:
         if status:
             return cls.onboarding_statuses.get(status)
         return cls.not_started
+
+
+class VerificientOnboardingProfileStatus:
+    """
+    A class for mapping onboarding statuses from Verificient's onboarding status endpoint to
+    internal database values.
+    """
+    no_profile = 'no-profile'
+    approved = 'approved-in-course'
+    other_course_approved = 'approved-in-different-course'
+    rejected = 'rejected'
+    expired = 'expired'
+    pending = 'pending'
+
+    profile_status_mapping = {
+        no_profile: None,
+        approved: ProctoredExamStudentAttemptStatus.verified,
+        other_course_approved: InstructorDashboardOnboardingAttemptStatus.other_course_approved,
+        rejected: ProctoredExamStudentAttemptStatus.rejected,
+        expired: ProctoredExamStudentAttemptStatus.expired,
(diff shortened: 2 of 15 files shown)


edx_proctoring/views.py
```diff
@@ -489,6 +494,65 @@ def get(self, request):
         elif relevant_attempt:
             data['onboarding_status'] = relevant_attempt.status
```
alangsto, line 495: Why are we not updating the expiration date here? Should that be added in?

edx_proctoring/views.py
```diff
@@ -489,6 +494,65 @@ def get(self, request):
...
+                    status = 404,
+                )
+            else:
+                readable_status = \
+                    VerificientOnboardingAttemptStatus.get_attempt_status_from_api_onboarding_status(
+                        attempt_data['status']
+                    )
+                expiration_date = attempt_data['expiration_date']
+
+                if readable_status != data['onboarding_status'] or expiration_date != data['expiration_date']:
```
alangsto, line 529: If the data from the endpoint differs from our logic, which data should we choose to use?

edx_proctoring/statuses.py
```diff
@@ -288,3 +288,34 @@ def get_onboarding_status_from_attempt_status(cls, status):
...
+    """
+    no_profile = 'no-profile'
+    approved = 'approved-in-course'
+    other_course_approved = 'approved-in-different-course'
+    rejected = 'rejected'
+    expired = 'expired'
+    pending = 'pending'
+
+    attempt_statuses = {
+        no_profile: None,
```
alangsto, line 305: Are we concerned about more granular statuses such as `created`, `ready_to_start` etc.? If the plan is to only use this API endpoint, we will lose that granularity, so is this simple mapping okay right now?

edx_proctoring/views.py
```diff
@@ -489,6 +494,65 @@ def get(self, request):
...
+                        course_id=course_id,
+                        response=str(exc),
+                        status=exc.http_status,
+                    )
+                )
+                LOG.warning(log_message)
+                return Response(data)
+
+            if attempt_data is None:
+                return Response(
```
alangsto, line 516: Instead of returning a 404, would it be better to return the data from our logic (if it exists)?

Zacharis278: while it pains me to remove code we've all put effort into I'm of the opinion that we don't need to maintain two code paths for this once the waffle flag is removed.

edx_proctoring/statuses.py
Zacharis278: I think it's fine and we should actually go further to make it clear that attempt status and onboarding status are two different things. I'd actually suggest renaming this model, and the associated `get_attempt_status_from_api_status` to make that more clear. (removing the word attempt)

edx_proctoring/views.py
```diff
@@ -489,6 +493,65 @@ def get(self, request):
         elif relevant_attempt:
             data['onboarding_status'] = relevant_attempt.status
 
+        if waffle.switch_is_active(ONBOARDING_STATUS_API):
+            attempt_data = None
+            try:
+                attempt_data = backend.get_onboarding_attempts(course_id=course_id, user_id=user.id)
```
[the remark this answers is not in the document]
alangsto: I agree, I can rename it as part of this PR

edx_proctoring/views.py
```diff
@@ -496,6 +499,61 @@ def get(self, request):
         elif relevant_attempt:
             data['onboarding_status'] = relevant_attempt.status
 
+        if waffle.switch_is_active(ONBOARDING_PROFILE_API):
+            attempt_data = None
+            try:
+                attempt_data = backend.get_onboarding_profile_info(course_id=course_id, user_id=user.id)
+            except BackendProviderOnboardingProfilesException as exc:
+                log_message = (
+                    'Failed to use backend onboarding status API endpoint for user_id={user_id}'
```
schenedx, line 508: #NIT Please add to the warning message saying, the `onboarding status will be determined by user's onboarding exam attempts` or something like that?

edx_proctoring/views.py
```diff
@@ -496,6 +499,61 @@ def get(self, request):
...
+                    'in course_id={course_id} because backend failed to respond. '
+                    'Status: {status}, Response: {response}'.format(
+                        user_id=user.id,
+                        course_id=course_id,
+                        response=str(exc),
+                        status=exc.http_status,
+                    )
+                )
+                LOG.warning(log_message)
+                return Response(data)
```
schenedx, line 518: I understand this data is coming from the logic before. Can we package that logic and the logic into a new function so we can understand the boundary of the onboarding status from attempts logic better?

edx_proctoring/views.py
schenedx: Make sense. No onboarding status API means no onboarding status from this point on.

edx_proctoring/views.py
```diff
@@ -496,6 +499,61 @@ def get(self, request):
...
+                    'Backend onboarding status API endpoint info for user_id={user_id}'
+                    'in course_id={course_id} differs from system info. Endpoint onboarding attempt'
+                    'has status={endpoint_status}. System onboarding attempt has status={system_status}.'.format(
+                        user_id=user.id,
+                        course_id=course_id,
+                        endpoint_status=readable_status,
+                        system_status=data['onboarding_status']
+                    )
+                )
+                LOG.warning(log_message)
```
schenedx, line 545: Sounds like an error log message to me?

schenedx (commented):
See my inline comments. I think I don't have strong opinion on this work after my comments. Therefore you can feel merge or not regardless if you address my comments.

edx_proctoring/views.py
alangsto: I just moved the logic that gets attempt data from our models to a function in api.py, this will make cleanup work for the waffle flag much easier

Zacharis278 (approved):

edx_proctoring/api.py
```diff
@@ -2723,3 +2723,42 @@ def get_enrollments_can_take_proctored_exams(course_id, text_search=None):
     """
     enrollments_service = get_runtime_service('enrollments')
     return enrollments_service.get_enrollments_can_take_proctored_exams(course_id, text_search)
+
+
+def get_onboarding_attempt_data_for_learner(course_id, user, backend):
```
schenedx, line 2728: Nice. This makes the code easier to read.

edx_proctoring/api.py
```diff
@@ -2723,3 +2723,42 @@ def get_enrollments_can_take_proctored_exams(course_id, text_search=None):
...
+    if not relevant_attempt or relevant_attempt.status != ProctoredExamStudentAttemptStatus.verified:
+        attempt_dict = get_last_verified_onboarding_attempts_per_user(
+            [user],
+            backend,
+        )
+        verified_attempt = attempt_dict.get(user.id)
+
+    if verified_attempt:
+        data['onboarding_status'] = InstructorDashboardOnboardingAttemptStatus.other_course_approved
+        data['expiration_date'] = verified_attempt.modified + timedelta(days=730)
```
schenedx, line 2760: #NIT can we update the number `730` to be a `constant` so it's not so generic embeded in code?

schenedx (approved):
👍

PR #852, opened 2021-05-20
[Chore]: MST-823 Improve unit tests on is_status_acknowledged functionality

schenedx:
**Description:**

I found the unit tests for is_status_acknowledged functionality has issues and do not cover all test cases. 
This functionality would exposes exam content after
exam past due, which on edx xblocks, the past due date would prevent
submission of the problems again.

**JIRA:**

[MST-823](https://openedx.atlassian.net/browse/MST-823)


Zacharis278 (approved):
NIT: These tests would be easier to read if the doc string stated the expected behavior we are testing for IMO

PR #855, opened 2021-05-21
feat: Modify Learner Onboarding API to determine whether onboarding is past due in a course and display this status in the onboarding panel

MichaelRoytman:
**Description:**

In [MST-734](https://openedx.atlassian.net/browse/MST-734), code was introduced to fix a bug where the learner onboarding panel was not showing up in learners' course outlines when the onboarding exam(s) was/were past due. However, this did not address the fact that there remained a link to a past due onboarding exam in the panel, generating confusion for learners.

This code change addresses this issue by extending the StudentOnboardingStatusView to add a new field "onboarding_past_due". "onboarding_past_due" is true if every onboarding exam in the course is past due. If there is at least one currently available or to be released onboarding exam in the course, we favor that onboarding exam and return false. If this flag is true, and the learner's onboarding status is in a state where the onboarding link is to be displayed, we display "Onboarding Past Due" instead. This code also changes the current behavior to not pass "onboarding_link" unless the onboarding exam is accessible by the learner.

Screen Shot 2021-05-21 at 3 46 57 PM

**JIRA:**

[MST-746](https://openedx.atlassian.net/browse/MST-746)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,15 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+[3.11.0] - 2021-05-24
+~~~~~~~~~~~~~~~~~~~~~
+* Add ability to get onboarding statuses from a proctoring provider API endpoint
+* Extend the learner onboarding status API to determine whether the only onboarding exam or all
+  onboarding exams are past due and past an "onboarding_past_due" flag in the response. modify
+  the API to not return a link to the onboarding exam if the onboarding exam should not be
+  accessible by the learner (i.e. it is to be released or is past due).
+* Modify the display behavior of the learner onboarding panel to display "Onboarding Past Due"
+  if the only onboarding or all onboarding exams are past due.
 
 [3.10.2] - 2021-05-24
 ~~~~~~~~~~~~~~~~~~~~~
@@ -24,10 +33,6 @@ Unreleased
 * Add API endpoint which provides proctoring generic and backend specific
   instructions for the proctoring exam. Usage case is to provide required data
   for the learning app MFE.
-
-[3.10.0] - 2021-05-19
-~~~~~~~~~~~~~~~~~~~~~
-* Add by-backend configurability of the link which shows on the onboarding panel
 
 [3.10.0] - 2021-05-19
 ~~~~~~~~~~~~~~~~~~~~~
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -766,9 +766,10 @@ class TestStudentOnboardingStatusView(ProctoredExamTestCase):
         Test that the request returns a 404 error if the user cannot view the onboarding exam
         """
         course_sections = self.course_scheduled_sections
-        del course_sections[BlockUsageLocator.from_string(self.onboarding_exam.content_id)]
+        accessible_sequences = list(self.course_scheduled_sections.keys())
+        accessible_sequences.remove(BlockUsageLocator.from_string(self.onboarding_exam.content_id))
         set_runtime_service('learning_sequences', MockLearningSequencesService(
-            list(self.course_scheduled_sections.keys()),
+            accessible_sequences,
             course_sections,
         ))
         with mock_perm('edx_proctoring.can_take_proctored_exam'):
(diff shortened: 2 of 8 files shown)


CHANGELOG.rst
```diff
@@ -25,10 +36,6 @@ Unreleased
 ~~~~~~~~~~~~~~~~~~~~~
 * Add by-backend configurability of the link which shows on the onboarding panel
 
-[3.10.0] - 2021-05-19
```
MichaelRoytman, line 28: This seemed to be duplicated above, so I removed it.

edx_proctoring/tests/test_views.py
```diff
@@ -760,9 +760,10 @@ def test_ineligible_for_onboarding_exam(self):
         Test that the request returns a 404 error if the user cannot view the onboarding exam
         """
         course_sections = self.course_scheduled_sections
-        del course_sections[BlockUsageLocator.from_string(self.onboarding_exam.content_id)]
+        accessible_sequences = list(self.course_scheduled_sections.keys())
```
MichaelRoytman, line 763: My understanding is that, for this scenario, the course_sections should contain exams that the user does not have access to. accessible_sequences should not contain exams the learner does not have access to. I changed the test to reflect that.

schenedx (approved):
👍

edx_proctoring/views.py
```diff
@@ -464,22 +468,18 @@ def get(self, request):
         course_key = CourseKey.from_string(course_id)
         user = get_user_model().objects.get(username=(username or request.user.username))
         details = learning_sequences_service.get_user_course_outline_details(
-            course_key, user, pytz.utc.localize(datetime.max)
+            course_key, user, pytz.utc.localize(datetime.now())
```
Zacharis278, line 471: originally I was thinking this date pruned the results of `details.schedule`. Probably wrong on that. But for my sanity can you clarify how the date impacts that `details.schedule` object?

Zacharis278 (approved):
🚀

edx_proctoring/views.py
MichaelRoytman: Sure thing.

The `get_user_course_outline_details` method calls the [_get_user_course_outline_and_processors](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/outlines.py#L301) method. This runs the various processors on the course outline. The [ScheduleOutlineProcessor](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/processors/schedule.py#L19), which processes the start dates and due dates for exams, only extends the [inaccessible_sequences](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/processors/schedule.py#L58) method. It also includes a [schedule_data](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/processors/schedule.py#L109) method. It doesn't extend the `usage_keys_to_remove` method, which defines which usage keys to remove from the course outline. Note that [trimmed_course_outline](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/outlines.py#L354) in the [_get_user_course_outline_and_processors](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/outlines.py#L301) method is the result of removing the result of `usage_keys_to_remove` from the full course outline, and since the `ScheduleOutlineProcessor` doesn't extend this method, there are no keys to remove with respect to start dates and due dates. The `sections` attribute of the `UserCourseOutlineData` is the `sections` of the `trimmed_course_outline`, so I expect that the sections attribute will include exams that are yet to be released or that are past due.

The `schedule` attribute of `details` is determined by [calling](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/content/learning_sequences/api/outlines.py#L288) the `ScheduleOutlineProcessor`'s [schedule_data](https://github.com/edx/edx-platform/blob/351228cee40b5b0d51111d9a8c90a5432911838b/openedx/core/djangoapps/content/learning_sequences/api/processors/schedule.py#L109) method. The `schedule_data` method accepts as an argument the course outline returned by `_get_user_course_outline_and_processors`. It then iterates through the `sections` on `pruned_course_outline.sections` and creates the section classes from there. Because `sections` reflects the sections in the course outline with the `usage_keys_to_remove` removed, yet to be released and past due exams should still be in the `schedule`, since the `ScheduleOutlineProcessor` does not extend that method of the base class..

Does that help?

PR #858, opened 2021-05-25
fix: Verificient onboarding API requires obscured user id

alangsto:
**Description:**

The onboarding status API requires an obscured user id as a parameter. Currently we are receiving 404 responses from this endpoint because we are sending non-obscured user ids, which Verificient's system does not recognize.


bseverino (approved):

schenedx (approved):
👍

PR #861, opened 2021-05-27
fix: Fix bug in StudentOnboardingStatusView for no sequence schedule for a sequence

MichaelRoytman:
**Description:**

This code change fixes a bug introduced by https://github.com/edx/edx-proctoring/pull/855 ([MST-746](https://openedx.atlassian.net/browse/MST-746)). The Learning Sequences API returns details about a given user's course outline, including sequence scheduling information. In some cases, a sequence that is inaccessible to a learner is still returned in the schedule, because the learner is allowed to know of the sequence's existence. A future onboarding exam is an example of this. In other cases, inaccessible sequences should be truly inaccessible to the learner, and, as such, they are not returned in the schedule. An onboarding exam that is content gated by enrollment track to an enrollment track in which the learner is not enrolled is an example of this. The bug was that the `StudentOnbaordingStatusView` was not taking into account that an inaccessible sequence might not have a schedule. This code change fixes that bug by adding None checking.

**JIRA:**

None.

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[3.11.3] - 2021-05-27
+~~~~~~~~~~~~~~~~~~~~~
+* Fix a bug where the Learning Sequences API does not have a schedule for a sequence, which can occur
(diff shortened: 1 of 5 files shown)


schenedx (approved):
👍

matthugs (approved):

PR #871, opened 2021-06-04
feat: AA-773: Add in logic to mark special exams as complete

Dillon-Dumesnil:
**Description:**

We have had numerous support tickets/bugs related to special exams
not properly displaying completeness (see https://openedx.atlassian.net/browse/AA-773 for
details). Along with a corresponding edx-platform PR, this will now
call the instructor service to complete all children within a special
exam upon the exam being completed for the first time. Actual completion
happens in the edx-platform PR

**JIRA:**

[AA-773](https://openedx.atlassian.net/browse/AA-773)

**Pre-Merge Checklist:**

**Post-Merge:**


MatthewPiatetsky (commented):
it looks like you still need to update some files mentioned in the pre-merge checklist

edx_proctoring/signals.py
```diff
@@ -105,6 +106,16 @@ def on_attempt_changed(sender, instance, signal, **kwargs):  # pylint: disable=u
             # to archive it
             original = sender.objects.get(id=instance.id)
 
+            # if the exam was just completed for the first time, we want to mark it as complete
+            # in the completion service
```
mikix, line 110: A) you use completed on both sides of this sentence - is `completed` also a term of art in the proctoring djangoapp? Maybe `attempted` would work better?
B) Might be worth mentioning _why_ we have custom completion logic - that we are worried about timing out and such?

mikix (approved):

Dillon-Dumesnil:
> it looks like you still need to update some files mentioned in the pre-merge checklist
> the code itself looks fine to me though i haven't worked in proctoring before so a review from someone with experience in proctoring might be useful

Yeah, I will have someone on the squad review prior to merging. Need to add tests too, just wanted to check approach with y'all.

edx_proctoring/signals.py
Dillon-Dumesnil: A) I'll look into rephrasing. I think I used `completed` because of the field we are looking at being updated.
B) Sure thing

edx_proctoring/signals.py
```diff
@@ -105,6 +106,18 @@ def on_attempt_changed(sender, instance, signal, **kwargs):  # pylint: disable=u
             # to archive it
             original = sender.objects.get(id=instance.id)
 
+            # if the exam was finished for the first time, we want to mark it as
+            # complete in the Completion Service.
+            # This functionality was added because regardless of submission status
+            # on individual problems, we want to mark the entire exam as complete
+            # when the exam is finished since there are no more actions a learner can take.
+            if not original.completed_at and instance.completed_at:
```
schenedx, line 114: Is it difficult to have unit test covering the code change?

Dillon-Dumesnil: So I'm honestly not super sure. I didn't see any tests for the existing signals that are calling out so I wasn't sure how this repo wanted to handle that. I'm happy to try and figure it out though.

schenedx: I would feel more comfortable with a unit test. Also, it would improve the codecov check too.

Dillon-Dumesnil: Okay, happy to figure that out

alangsto, line 114: This approach looks good to me, but I wanted to point out that a learner could have multiple attempts for a single exam (this is due to exams being "resumable" now). 

However, if a learner does have multiple attempts only one should have a `completed_at` field - this is because we only mark the time for `completed_at` if an attempt transitions to `submitted`, and a learner cannot resume an exam once they have a `submitted` status on an attempt. 

Just wanted to point this out in case this would need to be triggered by anything other than a status change to `submitted` (i.e. if instead an attempt should be considered completed once it is reviewed)

Dillon-Dumesnil: Thank you for checking in on this! I believe I do want just the scenario where they have submitted so I think we are all good.
The intent here is that as soon as a learner has no more actions they can do, we mark the whole exam as complete. This is to help account for cases where a learner leaves a question blank and then the entire exam is forever incomplete, but the learner has no ability to correct for it either.

Dillon-Dumesnil:
Simon and Alie, I have added a test if you don't mind re-reviewing

schenedx (approved):
👍

PR #877, opened 2021-06-08
fix: only request backend status update to started once

alangsto:
**Description:**

If a user is on the ready to submit panel, but chooses to continue in the exam, their attempt status will be updated to started. Each time an attempt status is updated to started, we previously would make a call to the proctoring provider backend to let them know that the attempt has started. Verificient uses the most recent status update to started to calculate how long a learner’s attempt is, resulting in different attempt durations shown to instructors on the edX attempt panel versus on the Proctortrack review dashboard. 

With these changes, we will only make a call to the proctoring provider backend when an exam attempt is updated to `started` if it is the first time that attempt is transitioning to `started`.

**JIRA:**

[MST-862](https://openedx.atlassian.net/browse/MST-862)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,17 +13,22 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+
+[3.13.1] - 2021-06-08
+~~~~~~~~~~~~~~~~~~~~~
+* If an attempt transitions from `ready_to_submit` back to `started`, the proctoring provider
+  backend function `start_exam_attempt` will not be called.
+
+[3.13.0] - 2021-06-07
+~~~~~~~~~~~~~~~~~~~~~
+* If the Django setting `PROCTORED_EXAM_VIEWABLE_PAST_DUE` is false, exam content will not be viewable past
+  an exam's due date, even if a learner has acknowledged their status.
 * Extend exam attempt API to return exam type and to check if
   user has satisfied prerequisites before taking proctored exam.
 * Extend proctoring settings API to return additional data about proctoring
   provider.
 * Add API endpoint which provides exam review policy for specific exam.
   Usage case is to provide required data for the learning app MFE.
-
-[3.13.0] - 2021-06-07
-~~~~~~~~~~~~~~~~~~~~~
-* If the Django setting `PROCTORED_EXAM_VIEWABLE_PAST_DUE` is false, exam content will not be viewable past
-  an exam's due date, even if a learner has acknowledged their status.
 
 [3.12.0] - 2021-06-04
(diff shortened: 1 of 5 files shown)


edx_proctoring/api.py
```diff
@@ -1404,7 +1404,7 @@ def update_attempt_status(attempt_id, to_status,
         # only proctored/practice exams have a backend
         # timed exams have no backend
         backend_method = None
-        if to_status == ProctoredExamStudentAttemptStatus.started:
+        if (to_status == ProctoredExamStudentAttemptStatus.started) and add_start_time:
```
alangsto, line 1407: add_start_time is defined earlier in the function as
`add_start_time = (to_status == ProctoredExamStudentAttemptStatus.started  and not exam_attempt_obj.started_at   )`

It should have a value of true only if the attempt has never transitioned to `started` before.

edx_proctoring/api.py
```diff
@@ -1404,7 +1404,7 @@ def update_attempt_status(attempt_id, to_status,
         # only proctored/practice exams have a backend
         # timed exams have no backend
         backend_method = None
-        if to_status == ProctoredExamStudentAttemptStatus.started:
+        if to_status == ProctoredExamStudentAttemptStatus.started and add_start_time:
```
alangsto, line 1407: add_start_time is defined earlier in the function as
add_start_time = (to_status == ProctoredExamStudentAttemptStatus.started and not exam_attempt_obj.started_at )

It should have a value of true only if the attempt has never transitioned to started before.

CHANGELOG.rst
```diff
@@ -13,18 +13,23 @@ Change Log
...
+[3.13.1] - 2021-06-08
+~~~~~~~~~~~~~~~~~~~~~
+* If an attempt transitions from `ready_to_submit` back to `started`, the proctoring provider
+  backend function `start_exam_attempt` will not be called.
+
+[3.13.0] - 2021-06-07
+~~~~~~~~~~~~~~~~~~~~~
+* If the Django setting `PROCTORED_EXAM_VIEWABLE_PAST_DUE` is false, exam content will not be viewable past
+  an exam's due date, even if a learner has acknowledged their status.
 * Extend exam attempt API to return exam type and to check if
```
alangsto, line 26: These bullet points should have been listed under 3.13.0 when that version was merged/released (they were missed due to PRs being merged around the same time)

edx_proctoring/api.py
schenedx: #NIT please add a comment to the code here like how you added this explanation to the PR.

schenedx (approved):
👍

PR #878, opened 2021-06-09
Add Post Bulk Allowance Endpoint

mohtamba:
Implement the bulk allowance post endpoint. Had some questions regarding the implementation. I was wondering what the expected behavior should be if one of the exams or users passed were to be invalid? Should we add the allowances for all the other users and return a 200 response? Or should we not add any of the allowances and return a 400 error?

**Description:**

Adds a post endpoint for bulk allowances, adds functions to the views.py and api.py files.

**JIRA:**

[MST-809](https://openedx.atlassian.net/browse/MST-809)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -7,7 +7,7 @@ API which is in the views.py file, per edX coding standards
 
 import logging
 import uuid
-from datetime import datetime, time, timedelta
+from datetime import datetime, timedelta
 
 import pytz
 from opaque_keys import InvalidKeyError
@@ -519,8 +519,8 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
         ).format(value=value)
         try:
             multiplier = float(value) - 1
-        except ValueError:
-            raise AllowanceValueNotAllowedException(err_msg)
+        except ValueError as error:
+            raise AllowanceValueNotAllowedException(err_msg) from error
         if multiplier >= 0:
             for exam_id in exam_ids:
                 try:
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -503,7 +503,8 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
         for exam_id in exam_ids:
             time_allowed = str(round(get_exam_by_id(exam_id)["time_limit_mins"] * (float(value) - 1)))
             for user_id in user_ids:
-                add_allowance_for_user(exam_id, user_id, ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, time_allowed)
+                add_allowance_for_user(exam_id, user_id, 
+                ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, time_allowed)
     else:
         for exam_id in exam_ids:
             for user_id in user_ids:
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -531,7 +531,7 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
                     )
                     log.error(log_message)
                     continue
-                
+
                 exam_time = target_exam["time_limit_mins"]
                 added_time = round(exam_time * multiplier)
                 time_allowed = str(added_time)
@@ -555,8 +555,7 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
         err_msg = (
             u'allowance_value "{value}" should be non-negative integer value.'
         ).format(value=value)
-        if ProctoredExamStudentAllowance.is_allowance_value_valid(ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED,
-                                                                  value):
+        if value.isdigit():
             for exam_id in exam_ids:
                 try:
                     target_exam = get_exam_by_id(exam_id)
@@ -1664,7 +1663,7 @@ def reset_practice_exam(exam_id, user_id, requesting_user):
             user_id=user_id,
         )
     )
-    (log_msg)
+    log.info(log_msg)
 
     exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_current_exam_attempt(exam_id, user_id)
     if exam_attempt_obj is None:
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -7,7 +7,7 @@ API which is in the views.py file, per edX coding standards
 
 import logging
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime, time, timedelta
 
 import pytz
 from opaque_keys import InvalidKeyError
@@ -25,10 +25,12 @@ from django.utils.translation import ugettext_noop
 from edx_proctoring import constants
 from edx_proctoring.backends import get_backend_provider
 from edx_proctoring.exceptions import (
+    AllowanceValueNotAllowedException,
     BackendProviderCannotRegisterAttempt,
     BackendProviderNotConfigured,
     BackendProviderOnboardingException,
     BackendProviderSentNoAttemptID,
+    ProctoredBaseException,
     ProctoredExamAlreadyExists,
     ProctoredExamIllegalStatusTransition,
     ProctoredExamNotActiveException,
@@ -499,18 +501,86 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
     """
     exam_ids = set(exam_ids)
     user_ids = set(user_ids)
-    if allowance_type == 'time_multiplier':
-        for exam_id in exam_ids:
-            time_allowed = str(round(get_exam_by_id(exam_id)["time_limit_mins"] * (float(value) - 1)))
-            for user_id in user_ids:
-                add_allowance_for_user(exam_id, user_id, 
-                ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, time_allowed)
+
+    log_message = (
+        'Adding allowances of type allowance_type={allowance_type} with value={value} '
+        'for exams exam_ids={exam_ids} and users user_ids={user_ids}'.format(
+            allowance_type=allowance_type,
+            value=value,
+            exam_ids=exam_ids,
+            user_ids=user_ids
+        )
+    )
+    log.info(log_message)
+    if allowance_type == constants.TIME_MULTIPLIER:
+        err_msg = (
+            u'allowance_value "{value}" should be a float value greater than 1.'
+        ).format(value=value)
+        try:
+            multiplier = float(value) - 1
+        except ValueError:
+            raise AllowanceValueNotAllowedException(err_msg)
+        if multiplier >= 0:
+            for exam_id in exam_ids:
+                try:
+                    target_exam = get_exam_by_id(exam_id)
+                except ProctoredExamNotFoundException:
+                    log_message = (
+                        'Attempted to get exam_id={exam_id}, but this exam does not exist.'.format(exam_id=exam_id)
+                    )
+                    log.error(log_message)
+                    continue
+                
+                exam_time = target_exam["time_limit_mins"]
+                added_time = round(exam_time * multiplier)
+                time_allowed = str(added_time)
+                for user_id in user_ids:
+                    try:
+                        add_allowance_for_user(exam_id, user_id,
+                                               ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED,
+                                               time_allowed)
+                    except ProctoredBaseException:
+                        log_message = (
+                            'Failed to add allowance for user_id={user_id} '
+                            'for exam_id={exam_id}'.format(
+                                user_id=user_id,
+                                exam_id=exam_id
+                            )
+                        )
+                        log.error(log_message)
+        else:
+            raise AllowanceValueNotAllowedException(err_msg)
     else:
-        for exam_id in exam_ids:
-            for user_id in user_ids:
-                add_allowance_for_user(exam_id, user_id, ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, value)
-
-
+        err_msg = (
+            u'allowance_value "{value}" should be non-negative integer value.'
+        ).format(value=value)
+        if ProctoredExamStudentAllowance.is_allowance_value_valid(ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED,
+                                                                  value):
+            for exam_id in exam_ids:
+                try:
+                    target_exam = get_exam_by_id(exam_id)
+                except ProctoredExamNotFoundException:
+                    log_message = (
+                        'Attempted to get exam_id={exam_id}, but this exam does not exist.'.format(exam_id=exam_id)
+                    )
+                    log.error(log_message)
+                    continue
+                for user_id in user_ids:
+                    try:
+                        add_allowance_for_user(exam_id, user_id,
+                                               ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED,
+                                               value)
+                    except ProctoredBaseException:
+                        log_message = (
+                            'Failed to add allowance for user_id={user_id} '
+                            'for exam_id={exam_id}'.format(
+                                user_id=user_id,
+                                exam_id=exam_id
+                            )
+                        )
+                        log.error(log_message)
+        else:
+            raise AllowanceValueNotAllowedException(err_msg)
 
 
 def _check_for_attempt_timeout(attempt):
@@ -1584,7 +1654,7 @@ def reset_practice_exam(exam_id, user_id, requesting_user):
             user_id=user_id,
         )
     )
-    log.info(log_msg)
+    (log_msg)
 
     exam_attempt_obj = ProctoredExamStudentAttempt.objects.get_current_exam_attempt(exam_id, user_id)
     if exam_attempt_obj is None:
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -493,6 +493,25 @@ def remove_allowance_for_user(exam_id, user_id, key):
         emit_event(exam, 'allowance.deleted', override_data=data)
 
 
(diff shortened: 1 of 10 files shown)


edx_proctoring/api.py
```diff
@@ -493,6 +493,26 @@ def remove_allowance_for_user(exam_id, user_id, key):
         emit_event(exam, 'allowance.deleted', override_data=data)
 
 
+def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
+    """
+    Adds (or updates) an allowance for multiple users and exams
+    """
+    exam_ids = set(exam_ids)
+    user_ids = set(user_ids)
+    if allowance_type == 'time_multiplier':
```
schenedx, line 502: Instead of string, we should define these options as python ENUMs. See the `constants.py` for examples.

edx_proctoring/api.py
```diff
@@ -493,6 +493,26 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+        for exam_id in exam_ids:
+            time_allowed = str(round(get_exam_by_id(exam_id)["time_limit_mins"] * (float(value) - 1)))
+            for user_id in user_ids:
+                add_allowance_for_user(exam_id, user_id, 
+                ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, time_allowed)
```
schenedx, line 507: indentation is incorrect on this line.

edx_proctoring/api.py
```diff
@@ -493,6 +493,26 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
 
+def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
+    """
+    Adds (or updates) an allowance for multiple users and exams
+    """
+    exam_ids = set(exam_ids)
+    user_ids = set(user_ids)
+    if allowance_type == 'time_multiplier':
+        for exam_id in exam_ids:
+            time_allowed = str(round(get_exam_by_id(exam_id)["time_limit_mins"] * (float(value) - 1)))
```
schenedx, line 504: Instead of chaining so much functions together, could you please use temporary variables? For example:
```
target_exam = get_exam_by_id(exam_id)
if target_exam:
    exam_time_limit = target_exam.get('time_limit_mins')
    if exam_time_limit:
         time_allowed = ....
```
This way, you can handle the error conditions and edge cases properly.

edx_proctoring/api.py
```diff
@@ -493,6 +493,26 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+            for user_id in user_ids:
+                add_allowance_for_user(exam_id, user_id, 
+                ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, time_allowed)
+    else:
+        for exam_id in exam_ids:
+            for user_id in user_ids:
+                add_allowance_for_user(exam_id, user_id, ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED, value)
+
+
```
schenedx, line 513: Please add logging statements. Do `log.info()` for happy path, and `log.warning()` for edge cases. Use `log.error()` for error cases.

edx_proctoring/tests/test_api.py
```diff
@@ -498,6 +499,150 @@ def test_remove_allowance_for_user(self):
...
         self.assertEqual(len(ProctoredExamStudentAllowance.objects.filter()), 0)
 
+    def test_add_time_bulk_allowance(self):
+        """
+        Test to add time bulk allowances.
+        """
+        user_list = self.create_batch_users(3)
+        exam_list = []
+        exam_list.extend(
+            (create_exam(
```
schenedx, line 509: You can move the function call to the next line and indent one more `tab`

edx_proctoring/tests/test_api.py
```diff
@@ -498,6 +499,150 @@ def test_remove_allowance_for_user(self):
...
+        student_allowance_exam2 = ProctoredExamStudentAllowance.get_allowance_for_user(
+            exam_list[1], user_list[1], self.key
+        )
+        all_course_allowances = get_allowances_for_course(self.course_id)
+        self.assertEqual(len(all_course_allowances), 6)
+        self.assertIsNotNone(student_allowance)
+        self.assertEqual(student_allowance.value, '30')
+        self.assertEqual(student_allowance_exam2.value, '30')
+
+    def test_add_multiplier_bulk_allowance(self):
```
schenedx, line 539: Please learn about `ddt` library. Search this repo for `@ddt.data` decorators. This help you combine this test case with the test case above.

bseverino:
Still need to take a closer look at the code, but a suggestion for your implementation question: we could probably include the list of allowances that failed to process in the response, and still allow the others to be created. I think a 200 response is still acceptable for something like that.

edx_proctoring/tests/test_api.py
```diff
@@ -498,6 +499,150 @@ def test_remove_allowance_for_user(self):
...
+        exam_list.append(exam_list[0])
+        add_bulk_allowances(exam_list, user_list, 'time_multiplier', '1.5')
+        student_allowance = ProctoredExamStudentAllowance.get_allowance_for_user(
+            exam_list[0], user_list[1], self.key
+        )
+        all_course_allowances = get_allowances_for_course(self.course_id)
+        self.assertEqual(len(all_course_allowances), 6)
+        self.assertIsNotNone(student_allowance)
+        self.assertEqual(len(exam_list), 3)
+
```
schenedx, line 645: Please add unit tests to handle edge cases. For example:
- Courses without exams
- Exams without time_limit specified
- Multiplier somehow is negative.

edx_proctoring/views.py
```diff
@@ -1369,6 +1370,44 @@ def delete(self, request):
...
+    **Response Values**
+        * returns Nothing. Add or update the allowances for the users and exams.
+    """
+
+    @method_decorator(require_course_or_global_staff)
+    def put(self, request):
+        """
+        HTTP PUT handler. Adds or updates Allowances for many exams and students
+        """
+        return Response(add_bulk_allowances(
```
schenedx, line 1405: Please assign the result of `add_bulk_allowance` to a temp variable and handle potential exceptions if any.

schenedx (changes requested):
See my inline feedback. Really good first attempt. Be aware of PEP8 and pycodestyles. I am sure you can find online resource on how to style newlines and brackets according to those guidelines.

edx_proctoring/api.py
mohtamba: For logging statements I'm not too sure exactly what I should be logging as the get allowance function would log when there is an error with a specific allowance is accepted or rejected. Should I log if the allowance value is acceptable?

schenedx: According to the existing [implementation ](https://github.com/edx/edx-proctoring/blob/84df549ba644d7a15556aa1c1281b2ab91a8a524/edx_proctoring/api.py#L430)of `add_allowance_for_user`, it looks like the normal condition is already logged. But I think that's not good enough. You should also log.info the beginning of this function with parameters passed into this function to help future engineers understand this function is invoked. Then you should catch exceptions thrown by `add_allowance_for_user` in this function, log and handle properly by keep going.

edx_proctoring/api.py
```diff
@@ -494,6 +496,93 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+            user_ids=user_ids
+        )
+    )
+    log.info(log_message)
+    if allowance_type == constants.TIME_MULTIPLIER:
+        err_msg = (
+            u'allowance_value "{value}" should be a float value greater than 1.'
+        ).format(value=value)
+        try:
+            multiplier = float(value) - 1
```
schenedx, line 521: Could you please handle the validation of multiplier value first? Check it's type as a number. See https://stackabuse.com/python-check-if-variable-is-a-number

edx_proctoring/api.py
```diff
@@ -494,6 +496,93 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+                        log_message = (
+                            'Failed to add allowance for user_id={user_id} '
+                            'for exam_id={exam_id}'.format(
+                                user_id=user_id,
+                                exam_id=exam_id
+                            )
+                        )
+                        log.error(log_message)
+        else:
+            raise AllowanceValueNotAllowedException(err_msg)
```
schenedx, line 553: Please move this raise function to the beginning of the logic, instead of the `else` statement here.

edx_proctoring/api.py
```diff
@@ -494,6 +496,93 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+    else:
+        err_msg = (
+            u'allowance_value "{value}" should be non-negative integer value.'
+        ).format(value=value)
+        if value.isdigit():
```
schenedx, line 558: This should be checked already by the same validation logic checked in the beginning of the function.

edx_proctoring/views.py
schenedx: Could you please address my comment here?

schenedx (changes requested):
See my newly added comments.

edx_proctoring/api.py
mohtamba: I was checking in both blocks of code for value validation because for the multiplier the value can be a float, however for the additional minutes the value must be an int. If I were to just check if the value was either a float or an int, then when I move into the additional time function I would have to check again to make sure that the value is an int. I was thinking I could leave it to the add_allowance_function, but the trade-off would then be that if the value is invalid for add_allowance we would still have to go through the for loop for every student and exam and receive the value error for each one rather than once in the start.

schenedx: If you read the [add_allowance_for_user](https://github.com/edx/edx-proctoring/blob/ebd87ee7759f3e8c842ef5e4706b9221efc2c577/edx_proctoring/models.py#L659) definition closely, you can see the validation of the value is already preformed there. This model function is the function called from the api function [here](https://github.com/edx/edx-proctoring/blob/84df549ba644d7a15556aa1c1281b2ab91a8a524/edx_proctoring/api.py#L444). AllowanceValueNotAllowedException is the exception raised and you did handle it. 
Notice the model field for `value` is actually a [character field](https://github.com/edx/edx-proctoring/blob/ebd87ee7759f3e8c842ef5e4706b9221efc2c577/edx_proctoring/models.py#L614), not a numbers field. 
All of this tells me your unit test cases are not covering for enough edge cases. I'll leave some comment there to prompt more test cases.

edx_proctoring/tests/test_api.py
```diff
@@ -499,6 +500,233 @@ def test_remove_allowance_for_user(self):
...
+    @ddt.data(
+        (
+            ADDITIONAL_TIME,
+            '30',
+            '30',
+            '30'
+        ),
+        (
+            TIME_MULTIPLIER,
+            '1.5',
```
schenedx, line 512: Can you please add non-digit test cases, like `abc` or `24bb`?

edx_proctoring/tests/test_api.py
```diff
@@ -499,6 +500,233 @@ def test_remove_allowance_for_user(self):
         remove_allowance_for_user(student_allowance.proctored_exam.id, self.user_id, self.key)
         self.assertEqual(len(ProctoredExamStudentAllowance.objects.filter()), 0)
 
+    @ddt.data(
+        (
+            ADDITIONAL_TIME,
+            '30',
+            '30',
```
schenedx, line 507: Please add test cases with non-int to see the behavior of the function when a non-int is passed into it?
I know in production, passing in int for now is improbable. But we should also ensure future development of this repo do not use the function with non-int `ADDITIONAL_TIME`.

edx_proctoring/tests/test_api.py
mohtamba: I have unit tests for invalid values under the test_views.py file, should I also add such tests to test_api.py. As a follow-up, how should I determine which tests should be in either file or both?
`
    @ddt.data(        (            ADDITIONAL_TIME,            '-30'        ),        (            TIME_MULTIPLIER,            '-1.5'        ),        (            TIME_MULTIPLIER,            'invalid_value'        )    )    @ddt.unpack    def test_add_bulk_allowance_invalid_allowance_value(self, allowance_type, value):  # pylint: disable=invalid-name        """        Test to add bulk allowance with invalid allowance value        """        # Create exams.        user_list = self.create_batch_users(3)        user_id_list = [user.email for user in user_list]        exam1 = ProctoredExam.objects.create(            course_id='a/b/c',            content_id='test_content',            exam_name='Test Exam',            time_limit_mins=90,            is_active=True            )        exam2 = ProctoredExam.objects.create(            course_id='a/b/c',            content_id='test_content2',            exam_name='Test Exam2',            time_limit_mins=90,            is_active=True            )        exam_list = [exam1.id, exam2.id]        allowance_data = {            'exam_ids': exam_list,            'user_ids': user_id_list,            'allowance_type': allowance_type,            'value': value        }        response = self.client.put(            reverse('edx_proctoring:proctored_exam.bulk_allowance'),            json.dumps(allowance_data),            content_type='application/json'        )        self.assertEqual(response.status_code, 400)`

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+        'for exams exam_ids={exam_ids} and users user_ids={user_ids}'.format(
+            allowance_type=allowance_type,
+            value=value,
+            exam_ids=exam_ids,
+            user_ids=user_ids
+        )
+    )
+    log.info(log_message)
+
+    multiplier = 0
```
schenedx, line 517: Please at least add some comments before this to say the follow section is validation section

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+            raise AllowanceValueNotAllowedException(err_msg)
+
+    if allowance_type == constants.ADDITIONAL_TIME:
+        err_msg = (
+            u'allowance_value "{value}" should be a non-negative integer value'
+        ).format(value=value)
+        if not value.isdigit():
+            raise AllowanceValueNotAllowedException(err_msg)
+
+    successes = 0
```
schenedx, line 536: Please add comments to say the processing logic begins here

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+                    'Attempted to get exam_id={exam_id}, but this exam does not exist.'.format(exam_id=exam_id)
+                )
+                log.error(log_message)
+                for user_id in user_ids:
+                    failures += 1
+                    data.append({
+
+                            'exam_id': exam_id,
+                            'user_id': user_id,
+                            'status': 400
```
schenedx, line 554: Use python HTTP status constants. https://docs.python.org/3/library/http.html#http.HTTPStatus

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+                try:
+                    add_allowance_for_user(exam_id, user_id,
+                                           ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED,
+                                           time_allowed)
+                    successes += 1
+                    data.append({
+
+                            'exam_id': exam_id,
+                            'user_id': user_id,
+                            'status': 200
```
schenedx, line 571: Use the enum at https://docs.python.org/3/library/http.html#http.HTTPStatus

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+                            exam_id=exam_id
+                        )
+                    )
+                    log.error(log_message)
+                    failures += 1
+                    data.append({
+
+                            'exam_id': exam_id,
+                            'user_id': user_id,
+                            'status': 400
```
schenedx, line 588: Use the enum at https://docs.python.org/3/library/http.html#http.HTTPStatus

edx_proctoring/api.py
```diff
@@ -494,6 +496,147 @@ def remove_allowance_for_user(exam_id, user_id, key):
...
+        if not value.isdigit():
+            raise AllowanceValueNotAllowedException(err_msg)
+
+    successes = 0
+    failures = 0
+    data = []
+    if allowance_type == constants.TIME_MULTIPLIER:
+        for exam_id in exam_ids:
+            try:
+                target_exam = get_exam_by_id(exam_id)
```
schenedx, line 542: This block that gets the target_exam can be part of the data validation logic block above. This code is repeated in either allowance_type

edx_proctoring/api.py
schenedx: It doesn't looks like this data is used on the view. Do you imaging we should use this on the frontend?
I rather we don't use HTTP status code for frontend. To Simplify, you can omit this, and only when there is clear need on the frontend, we can come back to here and provide the needed data when you are developing the frontend. Sounds good?

schenedx (commented):
Added more inline code suggestions. Looks much more comprehensive now.

edx_proctoring/api.py
mohtamba: Yeah that sounds fair, However I was thinking it might be helpful to know which users were added successfully and which weren't, to notify after submitting the allowances in the case that there might have been a typo on one of the users and the instructor has to re add the allowance.

schenedx (approved):
👍

edx_proctoring/views.py
```diff
@@ -1389,6 +1391,65 @@ def delete(self, request):
...
+        * returns Nothing. Add or update the allowances for the users and exams.
+    """
+
+    @method_decorator(require_course_or_global_staff)
+    def put(self, request):
+        """
+        HTTP PUT handler. Adds or updates Allowances for many exams and students
+        """
+        try:
+            data, succesess, failures = add_bulk_allowances(
```
bseverino, line 1427: NIT: minor typo here and on line 1433 for "successes"

bseverino (approved):
One minor nit but LGTM, good work!

PR #894, opened 2021-06-24
[BD-26] Timer bar on non-sequence pages

UvgenGen:
Change URL for ProctoredExamAttempt API

**Jira**: [EDUCATOR-5857](https://openedx.atlassian.net/browse/EDUCATOR-5857)

### DIFF
--- a/edx_proctoring/tests/test_mfe_views.py
+++ b/edx_proctoring/tests/test_mfe_views.py
@@ -3,6 +3,7 @@ Tests for the MFE proctored exam views.
 """
 import json
 from itertools import product
+from urllib.parse import urlencode
 
 import ddt
 from mock import patch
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @UvgenGen! I've created [BLENDED-899](https://openedx.atlassian.net/browse/BLENDED-899) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


When this pull request is ready, tag your edX technical lead.

idegtiarov (approved):
LGTM so far

Zacharis278:
It looks like we are actually hitting this endpoint outside of an exam sequence, meaning this endpoint is technically in use in production. Can we make this backward compatible with the old URL format (temporarily at least, it doesn't have to be pretty)

UvgenGen:
We can leave the old URL that will use the same view, and update the view to use the content id from the query or from the URL parameter. Is this solution okay?

Zacharis278:
yeah that seems like the best solution to me. We can remove that extra URL once the frontend is deployed.

idegtiarov (approved):
LGTM

Zacharis278 (approved):
looks good thanks!

PR #899, opened 2021-07-07
Mohtamba/adjust allowance modal

mohtamba:
**Description:**

Fully functional bulk allowance modal, which allows instructors to add multiple exam and student allowances at a time.
Screenshots of the new UI:
Screen Shot 2021-07-13 at 12 12 33 PM
Screen Shot 2021-07-13 at 12 13 49 PM

**JIRA:**

[XXX-XXXX](https://openedx.atlassian.net/browse/XXX-XXXX)

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/api.py
```diff
@@ -535,7 +537,7 @@ def add_bulk_allowances(exam_ids, user_ids, allowance_type, value):
         if multiplier < 0:
             raise AllowanceValueNotAllowedException(err_msg)
 
-    if allowance_type == constants.ADDITIONAL_TIME:
+    if allowance_type == ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED[0]:
```
schenedx, line 540: Why is the `ProctoredExamStudentAllowance.ADDITIONAL_TIME_GRANTED` an array?

edx_proctoring/static/proctoring/js/views/proctored_exam_add_allowance_view.js
```diff
@@ -160,38 +205,37 @@ edx = edx || {};
...
-
-                // In case of Timed Exams, we show the "Additional Time" label and hide the Allowance Types Select
-                $('#additional_time_label').show();
-                // Even though we have the Select element hidden, the backend will still be using
-                // the Select's value for the allowance type (key).
-                $('select#allowance_type').val('additional_time_granted').hide();
-            }
-            this.updateAllowanceLabels('additional_time_granted');
+        selectExamAtIndex: function(examID, examName) {
+            console.log(examID);
```
schenedx, line 209: Remember to update these console logs to either a reasonable console message, or complete remove them.

edx_proctoring/static/proctoring/js/views/proctored_exam_add_allowance_view.js
```diff
@@ -160,38 +205,37 @@ edx = edx || {};
...
+                $(this).trigger('click');
+            });
+            if($('#proctored_exam').is(":visible")){
+                $('#proctored_exam').hide();
+                $('#timed_exam').show();
+                $('#allowance_type option[value="review_policy_exception"]').remove();
+            } else {
+                $('#proctored_exam').show();
+                $('#timed_exam').hide();
+                $('#allowance_type').append(new Option(gettext('Review Policy Exception'), 'review_policy_exception'));
```
schenedx, line 235: What kind of allowance_type is `Review Policy Exception`? Is that a multiplier or a time in minutes value?

edx_proctoring/static/proctoring/js/views/proctored_exam_add_allowance_view.js
```diff
@@ -201,20 +245,53 @@ edx = edx || {};
                 $('#allowance_value_label').text(gettext('Value'));
             }
         },
+        sortExams: function() {
```
schenedx, line 248: Better name would be `SortExamsByExamType`

schenedx (changes requested):
Please see my inline feedback.

edx_proctoring/api.py
mohtamba: In the model.py file ADDITIONAL_TIME_GRANTED is actually an array so I needed the key text for this and not the helper text for the variable.

edx_proctoring/static/proctoring/js/views/proctored_exam_add_allowance_view.js
mohtamba: For Proctored exams instructors are able to input a review policy exception such as 'notes are allowed', so any text can be used for this input.

edx_proctoring/api.py
schenedx: I see. Look at this function: https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/models.py#L712. That's what you want to use instead of directly comparing with the model constant.

alangsto:
Should the new UI be rolled out using a waffle flag? I don't think the ticket specifies this (sorry!), but I'm not sure if we've discussed how this is being released.

schenedx:
> Should the new UI be rolled out using a waffle flag? I don't think the ticket specifies this (sorry!), but I'm not sure if we've discussed how this is being released.

Good catch. I think we should put these behind a waffle. But this PR can merge without the waffle piece?

alangsto:
> > Should the new UI be rolled out using a waffle flag? I don't think the ticket specifies this (sorry!), but I'm not sure if we've discussed how this is being released.
> 
> Good catch. I think we should put these behind a waffle. But this PR can merge without the waffle piece?

As long as we don't forget that this will be included in any tagged releases, I think it can be merged without the waffle flag piece.

edx_proctoring/static/proctoring/js/models/proctored_exam_allowance_model.js
```diff
@@ -7,7 +7,7 @@ edx = edx || {};
     edx.instructor_dashboard.proctoring = edx.instructor_dashboard.proctoring || {};
 
     edx.instructor_dashboard.proctoring.ProctoredExamAllowanceModel = Backbone.Model.extend({
-        url: '/api/edx_proctoring/v1/proctored_exam/allowance'
+        url: '/api/edx_proctoring/v1/proctored_exam/bulk_allowance'
```
schenedx, line 10: Instead of making the change directly here, you would need a new Model like this:
`edx.instructor_dashboard.proctoring.ProctoredExamBulkAllowanceModel = Backbone.Model.extend({
        url: '/api/edx_proctoring/v1/proctored_exam/bulk_allowance'`.
Leave the original model intact

edx_proctoring/static/proctoring/js/views/proctored_exam_add_allowance_view.js
```diff
@@ -11,7 +11,9 @@ edx = edx || {};
         template: null,
         template_url: '/static/proctoring/templates/add-new-allowance.underscore',
         initialize: function(options) {
-            this.proctored_exams = options.proctored_exams;
+            this.all_exams = options.proctored_exams;
```
schenedx, line 14: Instead of changing on the existing view, you would need to copy and past the code here into a new view file.
Keep the old view file unchanged.
At the place where the view is instantiated, check the waffle flag passed from the server, and render the different view based on waffle's flag's value

edx_proctoring/static/proctoring/js/views/proctored_exam_allowance_view.js
```diff
@@ -10,7 +10,8 @@ edx = edx || {};
         initialize: function() {
             this.allowance_types = [
                 ['additional_time_granted', gettext('Additional Time (minutes)')],
-                ['review_policy_exception', gettext('Review Policy Exception')]
+                ['review_policy_exception', gettext('Review Policy Exception')],
```
schenedx, line 13: You want this view to have the waffle flag injected into it.
Then you can use the waffle in this view to determine the `add_allowance_view` to use for the pop up.

edx_proctoring/static/proctoring/spec/proctored_exam_add_allowance_spec.js
```diff
@@ -79,15 +79,16 @@ describe('ProctoredExamAddAllowanceView', function() {
...
...      </select>\n            </td>\n        </tr>\n        <tr>\n            <td colspan="3">\n                <div id="selected_exams" ></div>\n            </td>\n        </tr>\n        <tr>\n            <td>\n                <label><%- gettext("Allowance Type") %></label>\n            </td>\n        </tr>\n        <tr>\n            <td>\n                <select id="allowance_type">\n                    <% _.each(allowance_types, function(allowance_type){ %>\n                    <option value="<%= allowance_type[0] %>">\n                        <%= allowance_type[1] %>\n                    </option>\n                    <% }); %>\n                </select>\n            </td>\n        </tr>\n        <tr>\n            <td>\n                <label id=\'allowance_value_label\'><%- gettext("Input Additional Minutes as a Positive Number") %></label>\n            </td>\n        </tr>\n        <tr>\n            <td>\n                <input type="text" id="allowance_value" />\n            </td>\n        </tr>\n        <tr>\n            <td>\n                <input id=\'addNewAllowance\' type=\'submit\' value=\'Create Allowance\' />\n            </td>\n        </tr>\n    </table>\n</form>';
```
schenedx, line 91: Your unit test needs to be updated. Instead of changing it directly, you can copy and past what you have, and then rename the new pasted test suite.

Revert your changes on the existing test suite to the old code to test against the old views.

edx_proctoring/static/proctoring/templates/add-new-allowance.underscore
```diff
@@ -2,39 +2,80 @@
     select#proctored_exam {
         width: 100%;
     }
+    select#timed_exam {
```
schenedx, line 5: You would need a new template underscore file. Copy everything here into the new one. Revert the changes you have on this underscore file to the original.

schenedx (changes requested):
Instead of changing the UI on the existing model pop view and template itself. I believe you should create new UI files and logic. Copy and paste is perfectly acceptable here.
Then use waffle flag value to control whether you want the allowance view to pop the old modal pop or the new modal pop.

edx_proctoring/static/proctoring/js/views/proctored_exam_allowance_view.js
```diff
@@ -143,15 +144,29 @@ edx = edx || {};
...
             self.proctoredExamCollection.fetch({
                 success: function() {
-                    // eslint-disable-next-line no-new
-                    new edx.instructor_dashboard.proctoring.AddAllowanceView({
-                        course_id: self.course_id,
-                        proctored_exams: self.proctoredExamCollection.toJSON(),
-                        proctored_exam_allowance_view: self,
-                        allowance_types: self.allowance_types
-                    });
+                    if (!enableBulkAllowance) {
```
schenedx, line 153: Why check the falsy value but not the truthy here?

schenedx (approved):
👍

PR #902, opened 2021-07-13
[MST-806] Use verified name in exam attempt creation

bseverino:
**Description:**

If the verified name feature is active through 1) runtime service and 2) course waffle flag (implemented in https://github.com/edx/edx-platform/pull/28187), the following features are added:

- During proctored exam attempt creation, check to see if the user has a valid verified name.
  - If they do, use that name for the backend provider's ID check.
  - If not, create a verified name for the user, linking the exam attempt ID.
- When a proctored exam attempt is verified, check to see if there is a linked verified name. If there is, flip the `is_verified` field to True.

**JIRA:**

[MST-806](https://openedx.atlassian.net/browse/MST-806)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -891,7 +891,9 @@ def _create_and_decline_attempt(exam_id, user_id):
     )
 
 
-def _register_proctored_exam_attempt(user_id, exam_id, exam, attempt_code, review_policy):
+def _register_proctored_exam_attempt(
+    user_id, exam_id, exam, attempt_code, review_policy, verified_name=None,
+):
     """
     Call the proctoring backend to register the exam attempt. If there are exceptions
     the external_id returned might be None. If the backend have onboarding status errors,
@@ -905,16 +907,19 @@ def _register_proctored_exam_attempt(user_id, exam_id, exam, attempt_code, revie
     review_policy_exception = ProctoredExamStudentAllowance.get_review_policy_exception(exam_id, user_id)
 
     # get the name of the user, if the service is available
-    full_name = ''
     email = None
     external_id = None
     force_status = None
+    full_name = verified_name or ''
+    profile_name = ''
 
     credit_service = get_runtime_service('credit')
     if credit_service:
         credit_state = credit_service.get_credit_state(user_id, exam['course_id'])
         if credit_state:
-            full_name = credit_state['profile_fullname']
+            profile_name = credit_state['profile_fullname']
+            if not verified_name:
+                full_name = profile_name
             email = credit_state['student_email']
 
     context = {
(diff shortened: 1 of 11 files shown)


edx_proctoring/api.py
```diff
@@ -1054,9 +1063,22 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
+        if is_verified_name_enabled:
+            # If the user has a verified name, use it instead of their profile name for
+            # the backend's ID verification. If they don't have one, create one with the
+            # exam attempt.
+            user = USER_MODEL.objects.get(id=user_id)
+            verified_name_obj = get_verified_name(user)
+            if not verified_name_obj or not verified_name_obj.is_verified:
+                should_create_verified_name = True
+            if verified_name_obj:
+                verified_name = verified_name_obj.verified_name
```
ashultz0, line 1079: in the case where verified_name_obj exists but it is_verified = false, we do should_create_verified name which seems right to make the exam attempt, but then we use verified_name_obj.verified_name which is not verified

that doesn't seem right

edx_proctoring/api.py
```diff
@@ -1072,6 +1094,27 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
+    # Only create a verified name after the attempt is created, as we need the attempt ID
+    if should_create_verified_name:
+        try:
+            create_verified_name(
+                user,
+                verified_name=full_name,
+                profile_name=profile_name,
+                proctored_exam_attempt_id=attempt.id,
+            )
+        except VerifiedNameEmptyString:
```
ashultz0, line 1106: why try/except rather than checking full_name here?

edx_proctoring/api.py
```diff
@@ -1586,6 +1629,14 @@ def update_attempt_status(attempt_id, to_status,
...
     emit_event(exam, attempt['status'], attempt=attempt)
 
+    if (to_status == ProctoredExamStudentAttemptStatus.verified and exam['is_proctored']):
+        # If the user has a verified name entry linked to this exam attempt, approve it
+        user = USER_MODEL.objects.get(id=user_id)
+        try:
+            update_is_verified_status(user, True, proctored_exam_attempt_id=attempt_id)
+        except VerifiedNameDoesNotExist:
+            pass
+
```
ashultz0, line 1639: this is not really something you can solve in this PR but I was reading back to see if I could understand this and update_attempt_status could use some chopping up, it's huge

edx_proctoring/api.py
bseverino: This is so that, if the user has a "pending" verified name (IE, they're in the middle of ID verification), they can still use it for a proctored exam attempt. (thank you for bringing this up, I should leave a comment so that the use case is clearer)

ashultz0: there is a test for this case but it's not checking how _register_protctored is called as far as I can tell

edx_proctoring/api.py
bseverino: Yeah that sounds like a better way to go

edx_proctoring/api.py
```diff
@@ -1052,9 +1059,26 @@ def create_exam_attempt(exam_id, user_id, taking_as_proctored=False):
...
         raise StudentExamAttemptOnPastDueProctoredExam(err_msg)
 
+    name_affirmation_service = get_runtime_service('name_affirmation')
+    should_create_verified_name = False
+
     if taking_as_proctored:
-        external_id, force_status = _register_proctored_exam_attempt(
-            user_id, exam_id, exam, attempt_code, review_policy
+        verified_name = None
+        if name_affirmation_service and is_verified_name_enabled:
```
schenedx, line 1067: Can we move this block of logic into a separate function? Private or not. the logic looks cleaner and becomes more re-usable when it can have a consistent input and output

edx_proctoring/templates/proctored_exam/footer.html
```diff
@@ -10,13 +10,18 @@
 </div>
 
 <script type="text/javascript">
+  // Course waffle flag for verified name
+  var main = document.getElementById('main');
+  var isVerifiedNameEnabled = main.dataset.isVerifiedNameEnabled.toLowerCase() === 'true';
```
schenedx, line 15: Are these value set by waffle flag pipelined in from edx-platform?

schenedx (approved):
I'd love to see a demo on this. 
The code is much clean now to read.
LGTM

edx_proctoring/templates/proctored_exam/footer.html
bseverino: Yep, the course waffle flag is implemented here: https://github.com/edx/edx-platform/pull/28187

PR #908, opened 2021-07-20
fix: remove deprecated ip fields (step 1 of 2)

ashultz0:
These IP fields are PII which blocks us from using a standard history
table which would copy them. They're not currently used and there is
no data in the DB more recent than 2016. This removes them from code,
the next release will drop them from the DB.

MST-872

**Pre-Merge Checklist:**

- [x] Confirmed Github reports all automated tests/checks are passing. (**ignoring tiny code coverage reduction** due to deleting some covered lines without adding new covered lines to keep the percentage up)

**Post-Merge:**


ashultz0:
ignoring tiny code coverage reduction due to deleting some covered lines without adding new covered lines to keep the percentage up

schenedx (approved):
👍

PR #918, opened 2021-07-27
docs: replace out of date config info with pointer to better info

ashultz0:
The JSON file instructions were out of date, and it was better to point at developing.rst than duplicate the better instructions. The email and debugging info also seemed more suited to the developer docs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,51 +52,12 @@ For a full list of Make targets:
 
     make help
 
-Configuration
+Developing
 -------------
 
-In order to use edx-proctoring, you must obtain an account (and secret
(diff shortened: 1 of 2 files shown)


Zacharis278:
humm I think we'd want to keep the instructions for having this configred with software secure somewhere. It definitely is not the normal development workflow anymore.

ashultz0:
@Zacharis278 there is some sandbox yml stuff on it still at https://github.com/edx/edx-proctoring/blob/ashultz0/rm-old-proctor-config-doc/docs/developing.rst#rpnow

Zacharis278 (approved):

PR #920, opened 2021-07-30
replace proctored exam attempt history

ashultz0:
The existing attempt history object is pretty weird and only updates when status changes. Simple history just records changes and is a lot easier to think about. For a while we'll write both, with the idea of removing the old one once some time has passed and no one could possibly want old history.

That also gives a little time to be confident that simple history covers the weird usages of the original history. That seems to be limited to using it for a little data when events come in after history delete and simple history should do fine.

One possible motivation for the weird original was the presence of PII in the history, that's gone now.

https://openedx.atlassian.net/browse/MST-872

Table name does not match specified name in ticket because that one is already full of the existing weird history.

**Pre-Merge Checklist:**

**Post-Merge:**


schenedx (approved):
LGTM
I like the return the new record first. If not found, default back to the old record filter logic.
That is the backward compatibility I agree with.

PR #929, opened 2021-08-06
fix: Fix bug with user IDs returned by proctoring vendor for learners not enrolled in eligible mode

MichaelRoytman:
**Description:**

This pull request fixes a bug that occurs when a proctoring vendor returns onboarding information that includes user IDs that represent learners that are not returned by the edX API as being enrolled in the course in a proctoring eligible mode. It also adds a logging statement to enable further investigation.

**JIRA:**

None.

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -760,9 +760,15 @@ class StudentOnboardingStatusByCourseView(ProctoredAPIView):
 
             obscured_user_ids_to_users = {obscured_user_id(user.id, onboarding_exam.backend): user for user in users}
 
+            missing_user_ids = []
             for onboarding_profile in onboarding_profile_info:
                 obscured_id = onboarding_profile['user_id']
                 user = obscured_user_ids_to_users.get(obscured_id)
+
+                if not user:
+                    missing_user_ids.append(onboarding_profile['user_id'])
+                    continue
+
                 onboarding_status = onboarding_profile['status']
                 data = {
                     'username': user.username,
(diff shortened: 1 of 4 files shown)


edx_proctoring/views.py
```diff
@@ -763,6 +763,18 @@ def get(self, request, course_id):
             for onboarding_profile in onboarding_profile_info:
                 obscured_id = onboarding_profile['user_id']
                 user = obscured_user_ids_to_users.get(obscured_id)
+
+                if not user:
+                    log_message = (
+                        'User with obscured user ID {obs_id} not in list of learners enrolled in course '
+                        '{course_id} with proctoring eligible enrollments.'
```
ashultz0, line 770: how many of these does the proctoring system return at a time? Will this be hundreds of duplicate log lines in a big mass? It might make sense to log a count or log a count plus a few examples.

ashultz0 (changes requested):
codeconv at first seems uninteresting but on reflection it means there's no test that this actually prevents the error, and it would be easy for a future editor to mess up the continue. Adding one missing user to a test somewhere would make it happy. 

also one question about amount of logging

MichaelRoytman:
@ashultz0 I agree with you about the test. I wasn't necessarily thinking this would be a permanent fix. I don't understand why this error is occurring, because Verificient shouldn't be sending us user IDs for users that we don't have a record of being enrolled in the course. I was planning to add a test once I have a permanent fix (the logging is meant to help elucidate what's going on). Do you feel that a test is still valuable here? I'm happy to add it.

ashultz0:
nah if this is meant to have another pass that's fine 👍

ashultz0 (approved):

edx_proctoring/views.py
MichaelRoytman: That's a good point. I don't really know. Theoretically, we didn't expect for this case to happen at all, so it's tough to guess how many logs this might generate. I can see what I can do about just logging a few examples.

ashultz0 (approved):

PR #938, opened 2021-08-17
feat: add signal emitters for proctored exam attempts

alangsto:
**Description:**

Describe in a couple of sentences how this pull request modifies the repository.

**JIRA:**

[XXX-XXXX](https://openedx.atlassian.net/browse/XXX-XXXX)

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/api.py
```diff
@@ -1132,9 +1101,15 @@ def create_exam_attempt(
         time_remaining_seconds=time_remaining_seconds,
     )
 
-    # Only create a verified name after the attempt is created, as we need the attempt ID
-    if should_create_verified_name:
-        _create_verified_name(user_id, full_name, profile_name, attempt.id, name_affirmation_service)
+    if taking_as_proctored:
+        # Emit signal for attempt creation
+        attempt_status_signal.send(
+            sender='edx_proctoring.create_exam_attempt',
```
ashultz0, line 1107: same comment as in the IDV one, we might not need sender since the signal is proctoring specific

alangsto: will update!

edx_proctoring/utils.py
```diff
@@ -393,3 +394,7 @@ def categorize_inaccessible_exams_by_date(onboarding_exams, details):
                 non_date_inaccessible_exams.append(onboarding_exam)
 
     return non_date_inaccessible_exams, future_exams, past_due_exams
+
+
+# Signal that is emitted when an attempt status is updated. Added to utils to avoid cyclic import in signals.py file
```
ashultz0, line 399: this seems weird, what's the cycle?

Zacharis278:
I missed the discussion, can you briefly explain how the receiver (name affirmation) is keeping track of the various senders and status translations?

edx_proctoring/utils.py
alangsto: signals.py imports edx_proctoring.api, so I can't import edx_proctoring.signals into api.py

edx_proctoring/api.py
```diff
@@ -1132,9 +1101,15 @@ def create_exam_attempt(
...
+            attempt_id=attempt.id,
+            user_id=user_id,
+            status=attempt.status,
+            full_name=full_name,
+        )
```
ashultz0, line 1112: we lost the concept of "should create verified name" - is that logic expected to be in the name affirmation service now?

alangsto:
> I missed the discussion, can you briefly explain how the receiver (name affirmation) is keeping track of the various senders and status translations?

As part of https://openedx.atlassian.net/browse/MST-972, we will add receivers in name affirmation for these signals. Name affirmation will then decide 1) how to translate these statuses and then 2) what to do with them (i.e. should we create, update, or do nothing for a verified name object). 

There were previous discussions for where we should translate the statuses, so if you feel strongly about where those translations should happen, let me know! Having name affirmation handle both translation and updating/creating felt best to me, as opposed to splitting those responsibilities between different apps.

edx_proctoring/api.py
alangsto: Yes, that was my intention - have both the translation of statuses and updating/creating a verified name be controlled by name affirmation

edx_proctoring/utils.py
ashultz0: feels like a cleanup ticket for proctoring/signals and this signal is in order. Though maybe after we remove 

signals uses api.ProctoredExamStudentAttemptSerializer which does not need to be in api
and it uses update_attempt_status which will be thornier to untangle

alangsto: Yeah, I had tried to get rid of the cyclic import but it seemed a little out of scope for this ticket. I can definitely file a follow up ticket for clean up though. 

I found it odd that we were importing ProctoredExamStudentAttemptSerializer from `api.py` as opposed to `serializers.py`, but wasn't sure if there was a reason for that.

Zacharis278:
> > I missed the discussion, can you briefly explain how the receiver (name affirmation) is keeping track of the various senders and status translations?
> 
> As part of https://openedx.atlassian.net/browse/MST-972, we will add receivers in name affirmation for these signals. Name affirmation will then decide 1) how to translate these statuses and then 2) what to do with them (i.e. should we create, update, or do nothing for a verified name object).
> 
> There were previous discussions for where we should translate the statuses, so if you feel strongly about where those translations should happen, let me know! Having name affirmation handle both translation and updating/creating felt best to me, as opposed to splitting those responsibilities between different services.

Slightly concerned the way we handle these translations on the name affirmation side might be more complicated since we'll want to have some sort of integration layer / configuration to manage it. Anything to keep specific proctoring/IDV fields/statuses from working their way into the core service for name affirmation. Sounds like we just need to figure this out in the other ticket.

schenedx (approved):
👍

edx_proctoring/api.py
```diff
@@ -1009,46 +1010,14 @@ def _get_verified_name(user_id, name_affirmation_service):
     new verified name should be created along with the proctored exam attempt.
```
ashultz0, line 1010: comment is now out of date

PR #942, opened 2021-08-20
Fix course staff allowance issue

mohtamba:
**Description:**

Fix issue where course staff can not add allowances for a course.

**JIRA:**

[MST-992](https://openedx.atlassian.net/browse/MST-992)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -4873,6 +4873,7 @@ class ExamBulkAllowanceView(LoggedInTestCase):
         self.user.is_staff = True
         self.user.save()
         self.client.login_user(self.user)
+        self.course_id = 'a/b/c'
         self.student_taking_exam = User()
         self.student_taking_exam.save()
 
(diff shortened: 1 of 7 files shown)


ashultz0:
this looks as expected but we should probably add a unit test for this case and clean up the existing unit test for test_add_bulk_allowance_non_staff_user which passes course id in the kwargs which isn't what the front end actually does

schenedx (approved):
👍

edx_proctoring/tests/test_views.py
```diff
@@ -4961,6 +4967,7 @@ def test_add_bulk_allowance_non_staff_user(self):  # pylint: disable=invalid-nam
         exam_list = [exam1.id, exam2.id]
 
         allowance_data = {
+            'course_id': self.course_id,
```
ashultz0, line 4970: I think we should also remove course_id from the kwargs since that's not how it's expected to come in

PR #948, opened 2021-08-27
fix: we no longer use the verfied enabled flag, strip it out

ashultz0:
Investigating how we ever had a bug I found that there was some remaining is_verified_enabled stuff left that can be stripped because nothing ever sends us true.

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/api.py
```diff
@@ -892,7 +892,7 @@ def _create_and_decline_attempt(exam_id, user_id):
     )
 
 
-def _register_proctored_exam_attempt(user_id, exam_id, exam, attempt_code, review_policy, verified_name=None):
+def _register_proctored_exam_attempt(user_id, exam_id, exam, attempt_code, review_policy):
```
alangsto, line 895: Why do we not want to use a verified name for registering a proctored exam attempt anymore?

ashultz0: I think there are no callers now which set verified enabled which means this name was never populated.

Possibly this PR points to a completely different fix which does not go this direction.

ashultz0:
from discussion with @alangsto we don't want to get rid of all this code, we want to restore is_verified_enabled to actually work by reading a waffle from the backend. That will look totally different so closing this PR.

PR #950, opened 2021-08-30
fix: restore verified name waffle flag to proctoring

ashultz0:
This flag was read and passed from the front end but we've removed that. If verified names are in use we still want to pass them to the proctoring when available. So read the waffle flag during the POST.

This is the thing we should actually do to fix the apparently unused code instead of ripping it out like https://github.com/edx/edx-proctoring/pull/948

Edit: "just reading the waffle flag" is more complicated. Name affirmation 0.9.0 exposes the waffle flag for us to prevent cross-library waffle sharing which is not allowed at this brunch.

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[3.24.3] - 2021-09-02
+~~~~~~~~~~~~~~~~~~~~~
+* Get verified name enabled from name affirmation service.
+
 [3.24.2] - 2021-09-01
 ~~~~~~~~~~~~~~~~~~~~~
 * Add exception handler and logging to proctored exam attempt emails. This prevents user errors
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -3,6 +3,6 @@ The exam proctoring subsystem for the Open edX platform.
 """
 
 # Be sure to update the version number in edx_proctoring/package.json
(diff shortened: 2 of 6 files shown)


schenedx (approved):
👍

edx_proctoring/views.py
```diff
@@ -1364,7 +1365,8 @@ def post(self, request):
         exam_attempt_id = create_exam_attempt(
             exam_id=exam_id,
             user_id=user_id,
-            taking_as_proctored=attempt_proctored
+            taking_as_proctored=attempt_proctored,
+            is_verified_name_enabled=waffle.switch_is_active(VERIFIED_NAME_FLAG)
```
alangsto, line 1369: I think this should be `is_enabled`, not `switch_is_active` because this is defined as a WaffleFlag in name affirmation. I'm also not positive about how flags can be accessed across services, have you tested that this works locally?

alangsto: Could be very off base with the function name, I just assumed it would be closer to `VERIFIED_NAME_FLAG.is_enabled()` which is in the name affirmation repo.

ashultz0: I was just patterning off other waffle calls in proctoring, I wasn't aware there was a whole set of possible alternate calls. I'll look into it more

ashultz0:
@alangsto @schenedx this got completely rewritten after the waffle flag quest

alangsto (approved):
LGTM

PR #951, opened 2021-08-30
fix: add email back for exam registration

alangsto:
This line was accidentally removed in https://github.com/edx/edx-proctoring/pull/938, need to add email back in.


mattcarter (approved):
Approved

PR #956, opened 2021-09-02
feat: management command for triggering a status update to exam attempt

alangsto:
**Description:**

Add management command for updating an attempt status based on its associated review. 

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/management/commands/tests/test_update_attempt_status_from_review.py
+++ b/edx_proctoring/management/commands/tests/test_update_attempt_status_from_review.py
@@ -1,0 +1,107 @@
+"""
+Tests for the update_attempt_status_from_review management command
+"""
+from datetime import datetime, timedelta
+
+import pytz
+
+from django.core.management import call_command
+
+from edx_proctoring.api import create_exam, create_exam_attempt, get_exam_attempt_by_id
+from edx_proctoring.models import ProctoredExamSoftwareSecureReview
+from edx_proctoring.runtime import set_runtime_service
+from edx_proctoring.statuses import ProctoredExamStudentAttemptStatus, SoftwareSecureReviewStatus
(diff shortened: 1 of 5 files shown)


edx_proctoring/management/commands/tests/test_update_attempt_status_from_review.py
```diff
@@ -0,0 +1,107 @@
...
+            taking_as_proctored=True
+        )
+        self.second_attempt = get_exam_attempt_by_id(self.second_attempt_id)
+        self.second_attempt_review_object = ProctoredExamSoftwareSecureReview.objects.create(
+            attempt_code=self.second_attempt['attempt_code'],
+            exam_id=self.second_exam_id,
+            student_id=self.user.id,
+        )
+
+        # update both reviews, and use update instead of save to avoid triggering a post_save signal
```
ashultz0, line 70: clever

MichaelRoytman:
Can we document somewhere the need for this management command, please? Most of the context is in an ephemeral Slack thread. Putting that context in the commit message or the MST ticket that you would link to from the commit message or the PR body would be helpful in the future for understanding why this management command was added.

alangsto:
> Can we document somewhere the need for this management command, please? Most of the context is in an ephemeral Slack thread. Putting that context in the commit message or the MST ticket that you would link to from the commit message or the PR body would be helpful in the future for understanding why this management command was added.

Just added a comment to the MST ticket and updated the commit message as well.

matthugs (approved):
I was worried about exam attempts that had been reset during the outage, but it seems like we'll use the newer history table for those so we shouldn't have trouble passing over them without trying to update a no-longer-existing attempt.

PR #957, opened 2021-09-03
fix: remove direct handler import in handler test

ashultz0:
without this import the signal/handler test depends on the
load bearing unused import in apps.py and if that is broken fails

with this import, it works even with the apps.py connection is broken

**Pre-Merge Checklist:**

**Post-Merge:**


alangsto (approved):
NIT: Could you add a comment to the unreleased section of the changelog for this?

PR #963, opened 2021-09-21
fix: if multiple attempts for a user exist, return attempt with non-reset status if possible

alangsto:
**Description:**
If a user has two or more active attempts, and one of them is has an `onboarding_reset` status, we should never return the `onboarding_reset` attempt if there is a more recently created attempt with a greater `id`. 

If we return an attempt with `onboarding_reset`, even though there are newer attempts, we risk showing incorrect information to the instructors. In this specific bug, if an attempt with an `onboarding_reset` status is considered the most recent (i.e. has a greater modified date than the attempt created after it, which is an edge case), an incorrect status will be shown on the instructor dashboard panel. 

To fix this, we should not return an attempt with an `onboarding_reset` status if there is a more recently created attempt, and instead return the more recently created attempt.

**JIRA:**

[MST-995](https://openedx.atlassian.net/browse/MST-995)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -870,7 +870,7 @@ class StudentOnboardingStatusByCourseView(ProctoredAPIView):
         onboarding_attempts = ProctoredExamStudentAttempt.objects.get_proctored_practice_attempts_by_course_id(
             course_id,
             users
-        ).values('user_id', 'status', 'modified')
+        ).values('user_id', 'status', 'modified', 'id')
 
         # get all of the last verified onboarding attempts of these users
         last_verified_attempt_dict = get_last_verified_onboarding_attempts_per_user(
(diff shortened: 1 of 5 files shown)


edx_proctoring/views.py
```diff
@@ -979,6 +980,12 @@ def _get_relevant_attempt_per_user(self, attempts):
                 # Always return a verified attempt if it exists.
                 if attempt['status'] == ProctoredExamStudentAttemptStatus.verified:
                     onboarding_attempts_per_user[attempt['user_id']] = attempt
+                # Always return a non reset attempt if its ID is greater
+                if (
+                    existing_attempt['status'] == ProctoredExamStudentAttemptStatus.onboarding_reset
+                    and existing_attempt['id'] < attempt['id']
```
alangsto, line 986: Not sure if `id` is the best field to compare. I'm using `id` on the assumption that a higher ID would have been created more recently.

schenedx (approved):
👍

edx_proctoring/views.py
ashultz0: assuming ID is the standard autoincrement field that's a pretty safe bet

there's probably some sequence of interleaved DB ops which would make it not true but it'd be pretty fiddly to set up for the same exam and user

PR #966, opened 2021-09-28
test: add GH action for migrations check

bseverino:
**Description:**

This check will result in an error like this if there is an invalid field in Django admin:

```
./manage.py check --settings=test_settings
SystemCheckError: System check identified some issues:

ERRORS:
<class 'edx_proctoring.admin.ProctoredExamStudentAttemptAdmin'>: (admin.E035) The value of 'readonly_fields[12]' is not a callable, an attribute of 'ProctoredExamStudentAttemptAdmin', or an attribute of 'edx_proctoring.ProctoredExamStudentAttempt'.
```

**JIRA:**

[MST-915](https://openedx.atlassian.net/browse/MST-915)

**Pre-Merge Checklist:**

### DIFF
--- a/.github/workflows/migrations-check.yml
+++ b/.github/workflows/migrations-check.yml
@@ -1,0 +1,37 @@
+name: Migrations Check
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  run_tests:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04 ]
+        python-version: [ 3.8 ]
+
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+
+    - name: Setup Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/migrations-check.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Install Python dependencies
+      run: |
+        pip install -r requirements/pip.txt
+        pip install -r requirements/dev.txt
+        pip install -r requirements/test.txt
+
+    - name: Run Tests
+      run: |
+        echo "Running migrations."
+        ./manage.py migrate --settings=test_settings
```
ashultz0, line 37: the example given was ./manage.py check --settings=test_settings and this one uses migrate instead. I think migrate runs check as a start anyway so it has that effect but also the migrate effect.

is this just what we normally do in edX repos?

bseverino: yeah, I'm not sure if this is overkill, I just pulled directly from the test platform is using. `check` is probably enough for our purposes, I have no problem updating it

ashultz0 (approved):

PR #969, opened 2021-09-30
feat: store encrypted video url from PSI

alangsto:
**Description:**

We want to capture a video review link that is sent to us by PSI in an exam review. Because the PSI endpoint contains sensitive information, we should encrypt the link before saving it to the review model.

**JIRA:**

[MST-951](https://openedx.atlassian.net/browse/MST-951)

**Pre-Merge Checklist:**

**Post-Merge:**


edx_proctoring/tests/test_reviews.py
```diff
@@ -175,6 +178,34 @@ def test_psi_review_callback(self, psi_review_status, review_status, credit_requ
             else:
                 self.assertEqual(len(notifications), 0)
 
+
+    def test_psi_video_url(self):
+        """
+        Test that review callback from PSI produces encrypted video link
+        """
+
+        with patch.object(SoftwareSecureBackendProvider, 'get_video_review_aes_key') as key_mock:
```
alangsto, line 187: Not an easy way to mock out the backend settings in a way that actually affects the provider class, as the provider class is initialized upon app load. In order to get the desired aes key value, I had to mock the getter.

edx_proctoring/utils.py
```diff
@@ -395,3 +403,58 @@ def categorize_inaccessible_exams_by_date(onboarding_exams, details):
                 non_date_inaccessible_exams.append(onboarding_exam)
 
     return non_date_inaccessible_exams, future_exams, past_due_exams
+
+
+def encrypt_and_encode(data, key):
```
alangsto, line 408: The functions added to utils.py are taken from https://github.com/edx/edx-platform/blob/master/lms/djangoapps/verify_student/ssencrypt.py. Other places in edx-platform use similar steps for AES key encryption.

schenedx (approved):
👍

edx_proctoring/utils.py
```diff
@@ -2,12 +2,18 @@
...
 import hmac
 import logging
 from datetime import datetime, timedelta
 
 import pytz
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import Cipher
+from cryptography.hazmat.primitives.ciphers.algorithms import AES
+from cryptography.hazmat.primitives.ciphers.modes import CBC
+from cryptography.hazmat.primitives.padding import PKCS7
```
matthugs, line 16: Based on https://cryptography.io/en/latest/#layout it sounds like the choice to import from hazmat was made in the past by folks who really deeply understand the cryptography at play, something I'm not sure we understand at this point. I wanted to flag this because the same section refers to a closer-to-the-surface and "more reasonable defaults" possible imports offered from the same library, which I'm wondering if we explored using.

If this specific way of encrypting/decrypting is required by the vendor, then we don't really have a choice here and should probably just resolve this comment without taking action. But I'd like to confirm that's the case, if possible.

edx_proctoring/utils.py
```diff
@@ -395,3 +403,58 @@ def categorize_inaccessible_exams_by_date(onboarding_exams, details):
...
+
+
+def aes_encrypt(data, key):
+    """
+    Return a version of the data that has been encrypted
+    """
+    cipher = Cipher(AES(key), CBC(generate_aes_iv(key)), backend=default_backend())
+    padded_data = pad(data)
+    encryptor = cipher.encryptor()
+    return encryptor.update(padded_data) + encryptor.finalize()
```
matthugs, line 425: It's absolutely fascinating to me that the api being used here uses addition in this way. I haven't dug far enough into the documentation to understand why it would be, but I trust that this is what we were doing before. Just... so weird.

edx_proctoring/views.py
```diff
@@ -1805,6 +1807,19 @@ def make_review(self, attempt, data, backend=None):
...
             )
 
+        video_review_link = backend_review.get('payload', {}).get('videoReviewLink')
+        if video_review_link:
+            aes_key_str = backend.get_video_review_aes_key()
+            if aes_key_str:
+                try:
+                    aes_key = codecs.decode(aes_key_str, "hex")
+                    review.encrypted_video_url = encrypt_and_encode(video_review_link.encode("utf-8"), aes_key)
+                except Exception as ex:  # pylint: disable=broad-except
```
matthugs, line 1817: Do we have any idea what a less broad except might be for this? Does the cryptography library have a base exception class we could use instead? (If we have a similar broad except in idv code, could we check logs to see what more-specific exceptions we've seen from that code?)

matthugs (approved):
was just checking this out to try to glean whether only the key is necessary for me to encrypt for [MST-1102](https://openedx.atlassian.net/browse/MST-1102), looks like it is. Have a few things I wanted to understand slightly better.

edx_proctoring/utils.py
alangsto: Hmm, I had not explored other import options, but saw that `hazmat` is used for both IDV and in `UsernameCipher`: https://github.com/edx/edx-platform/blob/3152c14192dcccac79b7b7a76eba91b380eba78a/lms/djangoapps/discussion/notification_prefs/views.py#L34. 

For this PR, we're not tied to a specific way of encrypting/decrypting because we (edX) will be the only user/consumer of this key.

edx_proctoring/views.py
alangsto: This is something I added because the encryption of the video url isn't behind a waffle flag, and I didn't want it to break incoming PSI reviews if the AES key that was added was mistakingly invalid (I might have been overly cautious with the addition of this try/catch though).

Once this is tested on stage or prod this try/catch should probably be removed.

alangsto: Actually, if we could only add the AES key on stage first, we could test the encryption of the video url there without worrying about breaking prod. Then there's no longer a need for this error handling (which doesn't exist in IDV anyway)

edx_proctoring/utils.py
matthugs: Could we write a follow-up ticket to explore refactoring this to use more tried-and-true encryption strategies? I imagine we're copying and pasting code here to avoid expanding scope into the whole rabbit hole of cryptographic algorithms, but maybe it makes sense to spend some time seeing whether we can use this library in a safer way?

edx_proctoring/views.py
matthugs: cool, glad I added the keys for prod & stage in separate PRs. That works out well

PR #1006, opened 2021-12-01
[MST-1162] Catch all worker errors

bseverino:
**Description:**

Catch errors from `onStartExamAttempt` and `onEndExamAttempt`.

**JIRA:**

[MST-1162](https://openedx.atlassian.net/browse/MST-1162)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[4.7.3] - 2021-12-06
+~~~~~~~~~~~~~~~~~~~~
+* Catch errors from `onStartExamAttempt` and `onEndExamAttempt`.
+
 [4.7.2] - 2021-11-17
 ~~~~~~~~~~~~~~~~~~~~
 * Add SimpleHistory to proctoring_proctoredexam table
--- a/edx_proctoring/__init__.py
+++ b/edx_proctoring/__init__.py
@@ -3,6 +3,6 @@ The exam proctoring subsystem for the Open edX platform.
 """
 
 # Be sure to update the version number in edx_proctoring/package.json
-__version__ = '4.7.2'
+__version__ = '4.7.3'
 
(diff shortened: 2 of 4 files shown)


schenedx (changes requested):
@bseverino The logging you made in this PR is when user submits exam. I think this is good. But we also want the logging to happen when user submits from `ready_to_start` to `started` state. That is done from https://github.com/edx/edx-proctoring/blob/master/edx_proctoring/static/index.js#L15
Could you look into that?

edx_proctoring/static/proctoring/js/exam_action_handler.js
```diff
@@ -48,6 +48,20 @@ edx = edx || {};
             });
     }
 
+    function logError(error, url) {
```
Zacharis278, line 51: A thought on making this call here. It might make more sense to have the special-exams library handle it. For example, when ping is called right now we return the worker promise to the react app and call that actionURL. In that case we'd end up calling the endpoint twice now, which isn't a huge deal but overall think we should do things on the react side if at all possible.

bseverino: Ohhh interesting - okay, let me look into that now.

Zacharis278: I guess my suggestion is to simply return the promise for each of the functions exposed on this.

Zacharis278: I think I may have misunderstood this a bit. My general suggestion still stands, I was thinking this code was called within the worker for some reason. All of this is already implemented separately on the react side, we mean that need to change this file at all so long as we don't break anything.

see: https://github.com/edx/frontend-lib-special-exams/blob/addb8f8fcba1dedbe582ee9d4f3c2a2a52d6950a/src/data/messages/handlers.js

Zacharis278 (approved):
would like to see how the other parts of this hook up, but this change lgtm 👍

schenedx (approved):
👍

PR #1008, opened 2021-12-07
remove old attempt history step 1

ashultz0:
First remove all uses of the model and the model itself. Next release will have the DB migration.

part of https://openedx.atlassian.net/browse/MST-963


ashultz0:
❌ is codecov complaining that some tested code was deleted

Zacharis278 (approved):
👋

PR #1018, opened 2022-01-12
fix: Proper handling to review callback on Non-proctored exam attempts

schenedx:
**Description:**

Sometimes, Verificient would send attempt reviewed status back to edX, but the exam associated with the attempts would be updated to be no longer a proctored exam. In this case, the review callback endpoint on edX would return 500 errors. This is not desired. Instead, we should log and return 200 to acknowledge we have received the event.

**JIRA:**

[MST-1301](https://openedx.atlassian.net/browse/MST-1301)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/tests/test_views.py
+++ b/edx_proctoring/tests/test_views.py
@@ -28,7 +28,8 @@ from edx_proctoring.api import (
     get_exam_attempt_by_id,
     mark_exam_attempt_as_ready_to_resume,
     reset_practice_exam,
-    update_attempt_status
+    update_attempt_status,
+    update_exam
 )
 from edx_proctoring.backends.tests.test_backend import TestBackendProvider
 from edx_proctoring.backends.tests.test_review_payload import create_test_review_payload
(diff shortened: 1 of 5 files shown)


Zacharis278:
part of me thinks this should be a 4xx error sense it's a write call to a resource that no longer exists. I suppose we could still alert on this, if it happens a lot something's probably wrong

edx_proctoring/tests/test_views.py
```diff
@@ -28,7 +28,7 @@
     get_exam_attempt_by_id,
     mark_exam_attempt_as_ready_to_resume,
     reset_practice_exam,
-    update_attempt_status
```
alangsto, line 31: I don't think you want to delete this

alangsto (commented):
Aside from the accidental delete, I think this looks good. I agree with @Zacharis278 that it may be useful to differentiate this type of response from other 200 responses if we want to alert on this behavior.

alangsto (approved):

Zacharis278 (approved):

PR #1037, opened 2022-03-31
fix: Resolved syntax error to fix python CI

Jawayria:
Python CI isn't running since the PR for Django 32 cleanup was merged due to syntax error. Fixed the error in this PR


Jawayria:
Python CI fixed so no need of this PR.

PR #1051, opened 2022-05-02
remove sure constraint (and remove sure)

ashultz0:
cleaning constraints out, proctoring is mostly benign except for one "sure" library version pin

we don't actually seem to use sure, so out it goes


schenedx (approved):
👍

alangsto (approved):
🔥

PR #1068, opened 2022-08-15
feat: endpoint for creating/updating exams

m-pitera:
**Description:**
Adds a new endpoint to create and update exams. This endpoint should take a list of all course exams and match the behavior as if calling [register_proctored_exams](https://github.com/openedx/edx-platform/blob/08eb1551b2d12fce876f0d6e7e6fdb2f83877b96/cms/djangoapps/contentstore/proctoring.py#L28) in edx-platform.

**JIRA:**
[MST-1521](https://2u-internal.atlassian.net/browse/MST-1521)

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -477,6 +477,90 @@ class ProctoredExamView(ProctoredAPIView):
                 )
         return Response(data)
 
+    def patch(self, request, course_id):
+        """
+        Create or update a list of exams.
+        """
+
+        # What proctoring.py Does:
+            # some things that I don't have to think about
+                # check whether 'enable_special_exams' is enabled
+                #   if not then exit
+                # check whether the course exists
+                #   if not then exit
+                # if timed or proctored exams are enabled
+                #   if not then exit
+                # ^ this is something that's still going to happen in the edx-platform
+                # and doesn't need to be done here
+
+            # get a list of requested exams
+            # for each exam
+                # grab the metadata
+
+                # check if exam exists
+                #   you can do this by checking by exam_content_id
+                # if it exists then:
+                #   define the exam_id
+                #   update_exam
+                # else:
+                #   define the metadata course_id and content_id
+                #   create_exam
+                
+            # mark any non-requested exams as inactive
+            # return a request
+
+        # grab the request
+        # this is the list of metadata that is passed from platform
+        request_exams = request.data
+        serializer = ProctoredExamSerializer(data=request.data, many=True)
+
+        # validate the serializer
+        if serializer.is_valid():
+
+            # for each exam
+            for exam in request_exams:
+                exam_metadata = {
+                        'exam_name': exam['exam_name'],
+                        'time_limit_mins': exam['time_limit_mins'],
+                        'due_date': exam['due_date'],
+                        'is_proctored': exam['is_proctored'],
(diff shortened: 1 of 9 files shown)


Zacharis278:
Opted to move this to its own view since there's some subtle differences in the object we are serializing here.

edx_proctoring/views.py
```diff
@@ -478,6 +479,86 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+                        is_proctored=request_exam.get('is_proctored', None),
+                        is_practice_exam=request_exam.get('is_practice_exam', None),
+                        external_id=request_exam.get('external_id', None),
+                        is_active=request_exam.get('is_active', None),
+                        hide_after_due=request_exam.get('hide_after_due', None),
+                        backend=request_exam.get('backend', None),
+                    )
+
+                    write_result.append({'exam_id': exam.id})
+                    msg = f'Updated exam {exam.id}'
```
ashultz0, line 518: do we want exam ID or the content_id which is the lookup key? Or both?

edx_proctoring/views.py
```diff
@@ -478,6 +479,86 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+                        is_proctored=request_exam.get('is_proctored'),
+                        is_practice_exam=request_exam.get('is_practice_exam'),
+                        external_id=request_exam.get('external_id'),
+                        is_active=request_exam.get('is_active'),
+                        hide_after_due=request_exam.get('hide_after_due'),
+                        backend=request_exam.get('backend'),
+                    )
+
+                    write_result.append({'exam_id': exam_id})
+                    msg = f'Created new exam {exam_id}'
```
ashultz0, line 536: same logging question

edx_proctoring/views.py
```diff
@@ -478,6 +479,86 @@ def get(self, request, exam_id=None, course_id=None, content_id=None):  # pylint
...
+            # The remaining exams exist but are not included in the registration payload.
+            # That means this exam should be disabled.
+            # We need to mark these exams as inactive (we don't delete!)
+            for exam in course_exams.values():
+                if exam.is_active:
+                    msg = f'Disabling exam {exam.id}'
+                    LOG.info(msg)
+                    update_exam(
+                        exam.id,
+                        is_proctored=False,
```
ashultz0, line 548: do we need to set is_proctored to false? It was a proctored exam, and now it is inactive but it still was a proctored exam when it was alive.

ashultz0 (approved):
looks pretty good, a few logging thoughts you can take or leave

I think realizing that registration is not really the same as updating freed you up to make some good twists to the code

edx_proctoring/views.py
Zacharis278: personally, if I'm debugging anything I'm just going to look this up by the database id which is a whole lot cleaner than messing with a big string (content_id)

edx_proctoring/views.py
Zacharis278: actually I don't even know why this is necessary, studio does this today so I just kept the behavior. https://github.com/openedx/edx-platform/blob/6c47a6d0aa0ea490cb84650f624c085565014050/cms/djangoapps/contentstore/proctoring.py#L147

PR #1084, opened 2022-10-19
feat: pass external attempt id to provider js

Zacharis278:
**Description:**

- API now returns the external attempt id to the UI
- The UI will pass the external id into the worker (https://github.com/edx/frontend-lib-special-exams/pull/72)
- JS worker code here will hand off that id to the vendor's interface

**JIRA:**

[MST-1659](https://openedx.atlassian.net/browse/MST-1659)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -818,6 +818,7 @@ def get_exam_attempt_data(exam_id, attempt_id, is_learning_mfe=False):
         'critically_low_threshold_sec': critically_low_threshold,
         'course_id': exam['course_id'],
         'attempt_id': attempt['id'],
+        'external_id': attempt['external_id'],
         'accessibility_time_string': _('you have {remaining_time} remaining').format(
             remaining_time=humanized_time(int(round(time_remaining_seconds / 60.0, 0)))
         ),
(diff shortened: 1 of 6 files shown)


edx_proctoring/api.py
```diff
@@ -818,6 +818,7 @@ def get_exam_attempt_data(exam_id, attempt_id, is_learning_mfe=False):
         'critically_low_threshold_sec': critically_low_threshold,
         'course_id': exam['course_id'],
         'attempt_id': attempt['id'],
+        'external_id': attempt['external_id'],
```
ashultz0, line 821: test_get_exam_attempt_data doesn't test all fields but since this new field would cause a hard to understand error if it breaks I think it's worth adding a check

ashultz0 (approved):
code looks fine but I think could use a test case update on the way out the door

edx_proctoring/api.py
Zacharis278: yeah that's fair. I didn't want to get into the business of testing all of the fields for this small change but a one-off couldn't hurt.

PR #1109, opened 2023-03-16
feat: API updates to support edx-exams

Zacharis278:
**Description:**

- Adds a new URL to return just the currently active exam attempt. This is the same behavior as calling the existing endpoint  `edx_proctoring/v1/proctored_exam/attempt/course_id/{settings.COURSE_ID_PATTERN}$` except it does not require a course id.
- Adds a flag to exams/attempts returned to UI to indicate this data originated from edx-proctoring. This is so the UI can know what service to request attempt create / update on.

These changes are needed to support https://github.com/edx/edx-exams/pull/88

**JIRA:**

[MST-1654](https://2u-internal.atlassian.net/browse/MST-1654)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/views.py
+++ b/edx_proctoring/views.py
@@ -112,6 +112,8 @@ from edx_proctoring.utils import (
     obscured_user_id
 )
 
+User = get_user_model()
+
 ATTEMPTS_PER_PAGE = 25
 
 LOG = logging.getLogger("edx_proctoring_views")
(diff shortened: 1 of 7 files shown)


edx_proctoring/views.py
```diff
@@ -284,7 +338,8 @@ def get(self, request, course_id, content_id=None):
         if exam:
             provider = get_backend_provider(exam)
             exam['type'] = get_exam_type(exam, provider)['type']
-            exam['passed_due_date'] = is_exam_passed_due(exam, user=request.user.id)
+            exam['passed_due_date'] = is_exam_passed_due(exam, user=User.objects.get(id=user_id))
```
Zacharis278, line 341: this was definitely broken before...

alangsto: nice catch

edx_proctoring/views.py
```diff
@@ -228,7 +271,18 @@ def get(self, request, course_id, content_id=None):
         is_learning_mfe = request.GET.get('is_learning_mfe') in ['1', 'true', 'True']
         content_id = request.GET.get('content_id', content_id)
 
-        active_exams = get_active_exams_for_user(request.user.id)
+        user_id = request.user.id
+        requested_user_id = request.GET.get('user_id', None)
+        if requested_user_id:
+            if request.user.is_staff:
```
alangsto, line 277: Just checking my own understanding, but a user can never supply the `requested_user_id` value via some edx-exams flow. And the edx-exams service worker is staff, so it will still get the correct learner data.

alangsto (approved):
LGTM 👍

edx_proctoring/views.py
Zacharis278: yeah the user_id passed in here is determined by edx exams and not any user input.

PR #1119, opened 2023-04-19
fix: user id should be an integer, not a string

alangsto:
**Description:**

Because the user_id can be read from a query parameter, if no type casting is done, the user_id will be a string. Other functions that take in a user_id expect an int value. 

**JIRA:**

[MST-1877](https://openedx.atlassian.net/browse/MST-1877)

**Pre-Merge Checklist:**

**Post-Merge:**


Zacharis278 (approved):
ah ha! that's what's going on

PR #1165, opened 2023-11-09
chore: remove bok-choy

salman2013:
**Description:**

As the bok-choy has been deprecated https://github.com/openedx/public-engineering/issues/13, we are removing its usage from the code in this PR.

Ticket : https://github.com/orgs/openedx/projects/55/views/1?pane=issue&itemId=43871891

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -9,18 +9,22 @@ appdirs==1.4.4
     # via fs
 asgiref==3.7.2
     # via django
+attrs==23.1.0
+    # via
+    #   outcome
+    #   trio
 backports-zoneinfo[tzdata]==0.2.1
     # via
     #   celery
     #   kombu
     # via celery
-bok-choy==2.0.2
-    # via -r requirements/test.in
     # via
     #   -c requirements/constraints.txt
     #   event-tracking
 certifi==2023.7.22
-    # via requests
+    # via
+    #   requests
+    #   selenium
 cffi==1.16.0
     # via
     #   cryptography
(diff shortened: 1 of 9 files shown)


farhan (approved):

requirements/test.txt
```diff
@@ -9,22 +9,27 @@ appdirs==1.4.4
...
-bok-choy==2.0.2
-    # via -r requirements/test.in
     # via
     #   -c requirements/constraints.txt
     #   event-tracking
 certifi==2023.7.22
-    # via requests
+    # via
+    #   requests
+    #   selenium
```
feanil, line 27: It doesn't look like anything else is using selenium either, so it's probably safe to remove that from `test.in` as well.

feanil (approved):
We should also drop the dependency on selenium but then I think this is good to merge.

requirements/test.txt
salman2013: @feanil should i remove the selenium under the scope of this PR? or create separate one?

feanil: I think we should remove selenium along with bok-choy if it's not being used independently of bok-choy.

UsamaSadiq:
Starting with `tox 4.0`  we need to specifically allow all scripts/tools to be used in the tox commands section.
In this specific case, we are directly mentioning the script which has not been specified in the `allowlist_externals` section for the `version_check` env.
To make it work with `tox 4` (Tested on my end locally), we can simply replace the following line
```
{toxinidir}/edx_proctoring/scripts/version_check.py
```
with
```
python {toxinidir}/edx_proctoring/scripts/version_check.py
```

requirements/test.txt
salman2013: @feanil done.

PR #1237, opened 2024-10-02
fix: allow timed exams for providers that are no longer valid

alangsto:
**Description:**

If a proctoring backend is removed, timed exams should render as expected. This PR also fixes a bug where the instructor dashboard was not available for courses that had a proctoring provider that was no longer valid.

**JIRA:**

[COSMO-515](https://2u-internal.atlassian.net/browse/COSMO-515)
[COSMO-513](https://2u-internal.atlassian.net/browse/COSMO-513)

**Pre-Merge Checklist:**

**Post-Merge:**

### DIFF
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -2943,7 +2943,7 @@ def get_student_view(user_id, course_id, content_id,
     is_proctored_exam = exam['is_proctored'] and not exam['is_practice_exam']
     is_timed_exam = not exam['is_proctored'] and not exam['is_practice_exam']
 
-    exam_backend = get_backend_provider(name=exam['backend'])
+    exam_backend = get_backend_provider(exam=exam)
 
     sub_view_func = None
     if is_timed_exam:
@@ -3019,8 +3019,17 @@ def is_backend_dashboard_available(course_id):
         active_only=True
     )
     for exam in exams:
-        if get_backend_provider(name=exam.backend).has_dashboard:
-            return True
+        try:
+            if get_backend_provider(name=exam.backend).has_dashboard:
+                return True
+        except NotImplementedError:
+            log.exception(
+                'No proctoring backend configured for backend=%(backend)s',
+                {
+                    'backend': exam.backend,
+                }
+            )
+
     return False
 
 
--- a/edx_proctoring/api.py
+++ b/edx_proctoring/api.py
@@ -1669,7 +1669,7 @@ def update_attempt_status(attempt_id, to_status,
         if email:
             try:
                 email.send()
-            except Exception as err:  # pylint: disable=broad-except
+            except Exception as err:
                 log.exception(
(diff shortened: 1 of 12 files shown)


edx_proctoring/api.py
```diff
@@ -2943,7 +2943,7 @@ def get_student_view(user_id, course_id, content_id,
     is_proctored_exam = exam['is_proctored'] and not exam['is_practice_exam']
     is_timed_exam = not exam['is_proctored'] and not exam['is_practice_exam']
 
-    exam_backend = get_backend_provider(name=exam['backend'])
+    exam_backend = get_backend_provider(exam=exam)
```
alangsto, line 2946: If an exam is passed in, as opposed to the provider name, a timed exam will always return "None".

edx_proctoring/api.py
```diff
@@ -3019,8 +3019,17 @@ def is_backend_dashboard_available(course_id):
         active_only=True
     )
     for exam in exams:
-        if get_backend_provider(name=exam.backend).has_dashboard:
-            return True
+        try:
+            if get_backend_provider(name=exam.backend).has_dashboard:
```
alangsto, line 3023: We don't want this case to throw an error because we still want the instructor dashboard to load for courses that still have software secure chosen.

edx_proctoring/handlers.py
```diff
@@ -27,8 +27,16 @@ def check_for_category_switch(sender, instance, **kwargs):
             exam = ProctoredExamJSONSafeSerializer(instance).data
             # from the perspective of the backend, the exam is now inactive.
             exam['is_active'] = False
-            backend = get_backend_provider(name=exam['backend'])
-            backend.on_exam_saved(exam)
+            try:
+                backend = get_backend_provider(name=exam['backend'])
```
alangsto, line 31: courses with software secure should be allowed to switch from a proctored to a timed exam. We don't want this code to throw an error in that case

edx_proctoring/handlers.py
```diff
@@ -127,15 +135,16 @@ def on_attempt_changed(sender, instance, signal, **kwargs):
...
-        backend = get_backend_provider(name=instance.proctored_exam.backend)
-        if backend:
-            result = backend.remove_exam_attempt(instance.proctored_exam.external_id, instance.external_id)
-            if not result:
-                log.error(
-                    'Failed to remove attempt_id=%s from backend=%s',
-                    instance.id,
-                    instance.proctored_exam.backend,
-                )
+        if instance.proctored_exam.is_proctored:
```
alangsto, line 138: Only remove from the backend if it is a proctored exam, so this does fix a bug where instructors wouldn't be able to delete a timed exam.

[**Question**] For the case of PSI, do we care if removing attempts blows up?

schenedx (approved):
👍

PR #1247, opened 2024-11-21
build: Assume that SoftwareSecure PII is retained (to fix annotations)

kdmccormick:
This helps us fix the PII checker in edx-platform. See: https://github.com/openedx/edx-proctoring/issues/1241


kdmccormick:
@nsprenkle @bmtcril , could I get one or both of your reviews on this?

bmtcril (approved):
Looks good to me, thanks!

PR #1268, opened 2025-02-13
test: disable  OpenSSL for PhantomJS

brian-smith-tcril:
**Description:**

The errors reported in https://github.com/openedx/edx-proctoring/issues/1264 seem to be stemming from an OpenSSL version conflict of some sort. I don't believe we need OpenSSL for the PhantomJS tests in this repo, so this PR disables it.

See https://stackoverflow.com/a/72679175

**Post-Merge:**


feanil (approved):
Makes sense. I don't think we need SSL for this testing.

PR #1280, opened 2025-04-22
Feat: Add django5.2 Support

abdullahQureshee:
Resolves [Add Django 5.2 support to edx-proctoring #1277](https://github.com/openedx/edx-proctoring/issues/1277)

## Add Django 5.2 Support
This PR updates the codebase to be compatible with Django 5.2 while maintaining compatibility with Django 4.2.

### Changes Made:
- Update ci and tox files
- Ran django-upgrade to apply necessary syntax updates for Django 5.2.
- Run tests using tox command and resolved the warning
- Ensured all modifications remain backward-compatible with Django 4.2.

### django-upgrade usage:
`git ls-files -z -- '*.py' | xargs -0r django-upgrade --target-version 5.2`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         toxenv:
           [
             "django42-drflatest",
+            "django52-drflatest",
             "quality",
             "pii_check",
             "version_check",
@@ -28,6 +29,7 @@ jobs:
             "js_lint",
             "rst_validation",
             "translations-django42",
+            "translations-django52",
           ]
     steps:
       - uses: actions/checkout@v1
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,11 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+[5.2.0] - 2025-04-22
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -56,7 +58,7 @@ jobs:
         run: tox
 
       - name: Run Coverage
-        if: matrix.python-version == '3.12' && matrix.toxenv == 'django42-drflatest'
+        if: matrix.python-version == '3.12' && (matrix.toxenv == 'django42-drflatest' || matrix.toxenv == 'django52-drflatest')
```
awais786, line 61: Not really necessary, but no big deal.

awais786 (approved):

.github/workflows/ci.yml
feanil: I think we should remove it, we don't want to run coverage multiple times, it might become confusing as the numbers could change between different versions of the tests and  might lead to confusion depending on which test run finishes first.  In general, I think we should run coverage on only the latest supported version.

awais786:
@feanil please review.

feanil (approved):

PR #1337, opened 2026-07-24
fix: reset proctored exam attempts reliably when the provider is unavailable

Anas12091101:
**Description**

When an instructor resets a student's proctored exam, edX also tells the
proctoring provider to delete its copy of the attempt. If the provider was down,
this used to fail with a confusing server error, and a reset could get stuck.

This PR makes the reset ask the provider to delete **first**, then delete locally,
and handles the provider's reply safely:
- If the provider is down or errors out, the instructor sees a clear "provider is
  temporarily unavailable, please try again" message and the attempt stays put
  (nothing is half-deleted).
- If the provider says the attempt is already gone, that counts as success, so
  trying again works.
- If the provider replies but doesn't confirm the delete, the attempt is kept so it
  can be retried instead of quietly disappearing.
- Every call to the provider now has a timeout (default 30s, configurable per
  backend), so a slow provider can't hang the request.

The `reset_attempts` admin command was also updated to tell the provider before
deleting, and to stop calling a provider that's down so it doesn't stall.

**Testing**

Automated: `pytest edx_proctoring/tests/test_api.py -k remove_exam_attempt`

Manual:
1. Set up a proctored exam and have a learner start an attempt (so it registers
   with the provider).
2. Make the provider unavailable — point its URL at an unreachable host, or make
   its delete call fail.
3. As an instructor, reset that learner's attempt → you see the "try again"
   message and the attempt stays in the list.
4. Bring the provider back and reset again → the reset works and the attempt
   disappears.

**Pre-Merge Checklist:**
- [x] Updated the version number in `edx_proctoring/__init__.py` and `package.json`
- [x] Confirmed Github reports all automated tests/checks are passing
- [ ] Approved by at least one additional reviewer

**Post-Merge:**
- [ ] Create a tag matching the new version number

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,10 +14,12 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 [6.1.0] - 2026-07-24
-* Add an explicit request timeout to the REST proctoring backend, and log and raise a
-  typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) when the provider fails to
-  remove an attempt, so callers can return a descriptive error to the instructor instead
-  of an unhandled 500.
+
+* Add an explicit request timeout to the REST proctoring backend (default 30s,
+  overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``), and log and
+  raise a typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) when the provider fails
+  to remove an attempt, so callers can return a descriptive error to the instructor
+  instead of an unhandled 500.
 
 [5.2.1] - 2025-12-05
 * Remove all references to Proctortrack
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,13 +13,20 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
-[6.1.0] - 2026-07-24
+[6.1.0] - 2026-07-28
 
-* Add an explicit request timeout to the REST proctoring backend (default 30s,
-  overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``), and log and
-  raise a typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) when the provider fails
-  to remove an attempt, so callers can return a descriptive error to the instructor
-  instead of an unhandled 500.
+* Make proctored exam attempt removal fail loudly when the proctoring provider errors,
+  instead of a silent local-only reset or an unhandled 500:
+
+  * Add an explicit request timeout to the REST proctoring backend (default 30s,
+    overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``).
+  * Raise a typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) both when the
+    provider is unreachable/times out and when it responds with an HTTP error status,
+    returning a clear message (including the provider's HTTP status) and logging the raw
+    provider response for debugging. Callers can then return a descriptive error to the
+    instructor instead of a silent local-only reset or an unhandled 500. Raising on a
+    provider HTTP error is on by default and overridable per backend via a
+    ``raise_on_remove_error`` key in ``PROCTORING_BACKENDS``.
 
 [5.2.1] - 2025-12-05
 * Remove all references to Proctortrack
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,12 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+[6.1.0] - 2026-07-24
+* Add an explicit request timeout to the REST proctoring backend, and log and raise a
+  typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) when the provider fails to
+  remove an attempt, so callers can return a descriptive error to the instructor instead
+  of an unhandled 500.
+
 [5.2.1] - 2025-12-05
 * Remove all references to Proctortrack
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,8 +19,9 @@ Unreleased
   is unavailable, instead of an unhandled 500 or a stuck reset:
 
   * Add an explicit request timeout to the REST proctoring backend (default 30s,
-    overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``) so a slow
-    provider cannot hang the request indefinitely.
+    overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``), applied to
+    every outbound provider request, so a slow provider cannot hang the request
+    indefinitely.
   * Call the provider from ``remove_exam_attempt`` **before** the local delete (rather
     than from the ``pre_delete`` signal). If the provider is unreachable/errors, a typed
     ``BackendProviderCannotRemoveAttempt`` (HTTP 502) is raised before anything is
@@ -28,9 +29,11 @@ Unreleased
     attempt is preserved. Because the failure no longer originates inside ``delete()``'s
     ``pre_delete`` signal, it does not leave the DB connection in a needs-rollback state,
     so a partially-failed multi-attempt reset can be retried in the same request.
-  * A provider that does not confirm removal (attempt never registered, or already gone
-    upstream) is logged and treated as already-removed, so the local delete still
-    proceeds and a retry converges instead of failing forever.
+  * On removal, a provider ``404`` (the attempt is already gone upstream) is treated as
+    success so a retried, partially-failed reset converges; any other HTTP error is raised
+    so the local attempt is kept for a later retry instead of being silently dropped.
+  * Route the ``reset_attempts`` management command through the same provider-removal
+    path, since provider cleanup no longer happens in the ``pre_delete`` signal.
 
 [5.2.1] - 2025-12-05
 * Remove all references to Proctortrack
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,20 +13,24 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
-[6.1.0] - 2026-07-28
+[6.1.0] - 2026-09-01
 
-* Make proctored exam attempt removal fail loudly when the proctoring provider errors,
-  instead of a silent local-only reset or an unhandled 500:
+* Make resetting a proctored exam attempt fail gracefully when the proctoring provider
+  is unavailable, instead of an unhandled 500 or a stuck reset:
 
   * Add an explicit request timeout to the REST proctoring backend (default 30s,
-    overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``).
-  * Raise a typed ``BackendProviderCannotRemoveAttempt`` (HTTP 502) both when the
-    provider is unreachable/times out and when it responds with an HTTP error status,
-    returning a clear message (including the provider's HTTP status) and logging the raw
-    provider response for debugging. Callers can then return a descriptive error to the
-    instructor instead of a silent local-only reset or an unhandled 500. Raising on a
-    provider HTTP error is on by default and overridable per backend via a
-    ``raise_on_remove_error`` key in ``PROCTORING_BACKENDS``.
+    overridable per backend via a ``timeout`` key in ``PROCTORING_BACKENDS``) so a slow
+    provider cannot hang the request indefinitely.
+  * Call the provider from ``remove_exam_attempt`` **before** the local delete (rather
+    than from the ``pre_delete`` signal). If the provider is unreachable/errors, a typed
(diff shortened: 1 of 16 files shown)


openedx-webhooks:
Thanks for the pull request, @Anas12091101!


This repository is currently maintained by `@openedx/openedx-unmaintained`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

edx_proctoring/api.py
```diff
@@ -1890,6 +1891,91 @@ def reset_practice_exam(exam_id, user_id, requesting_user):
...
+        )
+        raise BackendProviderCannotRemoveAttempt(unavailable_message) from exc
+    if not result:
+        # The provider responded but did not confirm removal, so keep the local attempt: a
+        # retry can converge once the provider is healthy instead of silently dropping it.
+        log.warning('Backend %s did not confirm removal of attempt_id=%s.', exam.backend, attempt.id)
+        raise BackendProviderCannotRemoveAttempt(unavailable_message)
+
+
+def _remove_exam_attempts_from_backend(attempts, failed_backends=None):
```
feanil, line 1951: This pattern where the failed_backend is a collection that is passed but then has side-effects added to it is not a great pattern.  It's difficult to read and beacause it can still be none, in some cases it will change the passed in set and in others it will not.  Why not just have a function that returns the set of backends that failed as the return value? Or better yet, a set of `(attempt, [failed_backends])`

edx_proctoring/api.py
```diff
@@ -1916,6 +2002,12 @@ def remove_exam_attempt(attempt_id, requesting_user):
     content_id = existing_attempt.proctored_exam.content_id
     to_status = existing_attempt.status
 
+    # Remove the attempt on the proctoring provider BEFORE deleting it locally, so a
+    # provider outage surfaces a descriptive error instead of a rolled-back, half-applied
+    # delete. See _remove_exam_attempt_from_backend for why this must not happen in the
+    # pre_delete signal.
+    _remove_exam_attempt_from_backend(existing_attempt)
+
     existing_attempt.delete_exam_attempt()
```
feanil, line 2011: Isn't the point to not delete this locally if there are failed attempts? If a removal fails, shouldn't we skip trying to delete it locally?

AhtishamShahid (approved):

edx_proctoring/api.py
Anas12091101: It already does. The call just above raises on any provider failure, so `delete_exam_attempt()` is never reached and the attempt is kept. The instructor gets a 502 with a "try again" message. An earlier revision had a `raise_on_error` flag, which may be what you saw.

One thing worth flagging, though: the other two delete paths, `clear_onboarding_errors` and the `reset_attempts` command, still delete the local attempt even when the provider fails. This is intentional and matches the behavior of the old `pre_delete` signal, which only logged the failure.

The trade-off is that leaving them as-is allows cleanup to continue during a provider outage, but can leave an orphaned record on the provider side. Changing this would mean nothing gets cleaned up until the provider recovers.

I'd leave them as they are, but I'm happy to change both if you prefer.

edx_proctoring/api.py
```diff
@@ -1890,6 +1891,91 @@ def reset_practice_exam(exam_id, user_id, requesting_user):
...
+            {
+                'attempt_id': attempt.id,
+                'external_id': attempt.external_id,
+                'exam_id': exam.id,
+                'user_id': attempt.user_id,
+                'backend': exam.backend,
+            }
+        )
+        raise BackendProviderCannotRemoveAttempt(unavailable_message) from exc
+    if not result:
```
feanil, line 1944: This strictness reaches further than the PR description suggests.

`reset_practice_exam` (`api.py:1884`) calls `remove_exam_attempt` when the attempt is `submitted`, and that path is reached from `views.py:1371` via the learner's own `reset_attempt` action - guarded at `views.py:1349` so it can only ever be their own attempt. That's the **learner's onboarding-exam retry**, not an instructor reset.

I confirmed it with a throwaway test on this branch (practice exam, `is_proctored=True`, backend `null`, `submitted` attempt, provider raising):

```
RAISED http_status=502 msg='The proctoring provider is temporarily unavailable,
      so this attempt could not be fully reset. Please try again in a few minutes.'
```
The connection-error case was already broken on master (it raised out of `pre_delete` as a 500), so that's a clear improvement. But *this* branch - provider responded, didn't confirm - goes from master's log-and-proceed to blocking with a 502. A learner who could previously retry onboarding during a partial provider problem now can't, and they see instructor-worded copy ("this attempt could not be fully reset").

Is blocking the learner's onboarding retry intended? If not, `reset_practice_exam` probably wants the best-effort helper. If it is intended, it's worth calling out in the CHANGELOG and giving the learner path a message written for them.

edx_proctoring/api.py
```diff
@@ -1890,6 +1891,91 @@ def reset_practice_exam(exam_id, user_id, requesting_user):
...
+        if backend in failed_backends:
+            continue
+        try:
+            _remove_exam_attempt_from_backend(attempt)
+        except BackendProviderCannotRemoveAttempt:
+            log.warning(
+                'Provider removal failed for backend %r; skipping further provider calls '
+                'for it during this cleanup and removing locally only.', backend
+            )
+            failed_backends.add(backend)
```
feanil, line 1976: Following up on my earlier comment on the signature - this is the line that makes the pattern hard to follow. When a set is passed in it's mutated in place; when it isn't, the mutation lands on a throwaway. Same function, two different contracts depending on the call site.

Returning the accumulator instead reads straight through and drops the `None` special-case:

```python
def _remove_exam_attempts_from_backend(attempts, failed_backends=frozenset()):
    # fresh mutable set per call; the shared default is only ever copied from, never mutated
    failed = set(failed_backends)
    ...
            failed.add(backend)
    return failed
```

with `reset_attempts.py` doing `failed_backends = _remove_exam_attempts_from_backend(qs, failed_backends)` per batch, so the threading across batches is visible at the call site instead of happening through the argument.

(An immutable default sidesteps the `None` check rather than relying on it - a plain `failed_backends=None` with `set(failed_backends or ())` works just as well if you prefer the more familiar idiom. The part that matters is returning rather than mutating.)

edx_proctoring/handlers.py
```diff
@@ -99,12 +99,16 @@ def on_allowance_changed(sender, instance, signal, **kwargs):
 
 
 @receiver(pre_save, sender=models.ProctoredExamStudentAttempt)
-@receiver(pre_delete, sender=models.ProctoredExamStudentAttempt)
```
feanil, line 102: Dropping this receiver also drops provider cleanup from the cascade paths.

`ProctoredExamStudentAttempt.user` and `.proctored_exam` are both `on_delete=CASCADE` (`models.py:375,377`), and Django's cascade collector fires `pre_delete` per collected row - so before this change, deleting a `ProctoredExam` notified the provider for every attempt under it. Now only the three explicit call sites do.

The user path is probably fine, since retirement has its own `retire_user` backend call. The `ProctoredExam` deletion path is the one I'd want confirmed: is that acceptable (we accept orphaned attempts upstream when an exam is deleted)?

edx_proctoring/handlers.py
```diff
@@ -99,12 +99,16 @@ def on_allowance_changed(sender, instance, signal, **kwargs):
...
     modified. Make a new entry with the previous value of the status in the
     ProctoredExamStudentAttemptHistory table.
+
+    Note: the proctoring provider is no longer notified from this signal on delete.
+    ``api.remove_exam_attempt`` calls the backend *before* the local delete so that a
+    provider outage does not roll back a half-applied delete (raising from a pre_delete
+    signal marks the DB connection needs-rollback).
     """
 
     if signal is pre_save:
```
feanil, line 114: Minor cleanup while you're in here: now that this receiver is only registered for `pre_save`, this condition is always true and the `else: return` closing the block is unreachable. The `signal` parameter is likewise only ever `pre_save` now. Could all be flattened.

edx_proctoring/backends/rest.py
```diff
@@ -41,6 +41,9 @@ class BaseRestProctoringProvider(ProctoringBackendProvider):
     has_dashboard = True
     supports_onboarding = True
     passing_statuses = (SoftwareSecureReviewStatus.clean,)
+    # Timeout (in seconds) applied to every outbound request to the provider so
+    # that a slow or unavailable provider cannot hang the request indefinitely.
+    timeout = 30
```
feanil, line 46: Worth knowing: `requests` applies a scalar `timeout` to the connect and read phases *separately*, so this is an upper bound of roughly 60s rather than 30s. If a hard 30s ceiling is the intent, a tuple like `(5, 25)` gets you there.

Also, this is user-facing config now (via the `timeout` key you added to `BACKEND_CONFIGURATION_ALLOW_LIST`), so it should probably be documented alongside the `PROCTORING_BACKENDS` example in `docs/developing.rst`.

edx_proctoring/backends/rest.py
```diff
@@ -227,13 +230,29 @@ def stop_exam_attempt(self, exam, attempt):
...
+        response = self.session.request('DELETE', url, timeout=self.timeout)
+        if response.status_code == 404:
+            return True
+        response.raise_for_status()
+        try:
+            data = response.json()
+        except ValueError:
+            log.exception('Decoding attempt removal %r -> %r', attempt, response.content)
+            return False
+        return data.get('status', None) == 'deleted'
```
feanil, line 255: `docs/backends.rst:151-157` documents this endpoint's contract as just "on success, return `{"status": "deleted"}`". This PR layers meaningful behaviour on top of that which provider implementers need to know: a 404 is now treated as success, any other non-2xx raises, and a 2xx that doesn't carry `status: deleted` is treated as *not* confirmed. Worth writing into that doc.

CHANGELOG.rst
```diff
@@ -13,6 +13,28 @@ Change Log
 
 Unreleased
 ~~~~~~~~~~
+[6.1.0] - 2026-09-01
```
feanil, line 16: Two things here:

- The behaviour change is worth spelling out for API consumers: the reset endpoint now returns **502** where it previously returned a 500 or silently succeeded. Anything driving the instructor dashboard reset will want to handle that.
- Minor formatting: this entry sits under `Unreleased` without its own `~~~~` underline. Separately (and not caused by this PR) there's no `[6.0.0]` entry at all - master is on 6.0.0 but the changelog tops out at 5.2.1, so this might be a chance to backfill it.

edx_proctoring/api.py
feanil: This still needs to be addressed.

